PR #133, opened 2014-04-10
Add ability to scan all running processes using -e (Windows only)

edeca:
Simple additions to enable scanning across all running processes, currently implemented on Windows only.

Processes that begin or end whilst yara is running will not be scanned as the process list is built once at startup.


edeca:
I'm closing this because I never received a reply to the query above and the effort required to reconcile code changes is now excessive.

PR #204, opened 2014-11-28
Override the functions that Yara uses for memory management.

dtilloy:
User can use yr_set_mem_functions() to override malloc,calloc,realloc,free,strdup and strndup function.
If you want to use the wrapper, you need to define function following the exact original prototype, and to define all of them.

The function yr_set_mem_functions() will return:
- ERROR_SUCCESS if handler set without error
- ERROR_WRONG_ARGUMENTS if an handler is missing

Note: this is not supported on Windows platform, and cannot be used with DMALLOC

### DIFF
--- a/libyara/include/yara/libyara.h
+++ b/libyara/include/yara/libyara.h
@@ -33,4 +33,11 @@ YR_API int yr_get_tidx(void);
 
 YR_API void yr_set_tidx(int);
 
+
+YR_API int yr_set_mem_functions(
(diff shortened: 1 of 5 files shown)


plusvic:
I was just waiting to have time to look into it. After a quick review I've noticed that there's no implementation for WIN32, do you plan to implement it? I would prefer maintaining the API identical in all platforms.

plusvic:
The WIN32 implementation never came and nobody else asked for this feature.

Issue #226, opened 2014-12-31
Rename entropy module to math and add more functions


wxsBSD:
Putting this here as a placeholder so you know what I'm working on. I'd hate to see the entropy module go out in a release as is, because I really think it needs to be renamed and expanded.

Here's what I intend to do:

* Rename entropy to math.
* Add other functions useful for calculating properties, especially related to encryption. I've got a few of these I'm adding currently.
* Any other functions that are useful when doing the calculation, specifically things like abs() and fabs().

PR #227, opened 2015-01-01
Rename the entropy module and add more to it.

wxsBSD:
Rename the entropy module to math, and add more functions to it.

The added functions are:

* arithmetic mean
* serial correlation
* monte carlo from pi

This is in addition to the original entropy function.

Each one takes either a string or an offset and length pair as arguments. Each function returns the absolute value of how far the calculation is from the expected value.

More details on these tests are at http://fourmilab.ch/random/. The only one I have not implemented from there yet is the Chi-square test.

This addresses #226.


wxsBSD:
The monte_carlo_pi functions return the percentage away from pi. Anything less than or equal to 0.06 is random.

The serial correlation test returns a value that approaches 0 for random input, and 1 for non-random input. In the case of the result being 0 the function will return -100000, to indicate that all values were equal.

The arithmetic mean test returns the percentage away from 127.5.

The entropy test returns the entropy value.

Here are some test rules:

```
// Anything below 0.06 is likely random.
// This rule fires because the string is not highly random.
rule monte_carlo_pi_test_string {
  condition:
    math.monte_carlo_pi("foobarbazfooooooooo") > 0.06
}

// Feed this the output of /dev/random and it should fire. Feed it any binary and it likely won't fire.
rule monte_carlo_pi_test {
  condition:
    math.monte_carlo_pi(0, filesize) <= 0.06
}

// The closer to 0 the result is, the more random it is. The closer to 1 the less random.
rule serial_correlation_test {
  condition:
    math.serial_correlation(0, filesize) < 0.2
}

// This function returns the distance from 127.5. The closer to 127.5 the more random the input.
rule arithmetic_mean_test {
  condition:
    math.arithmetic_mean(0, filesize) < 0.07
}
```

Using these functions allows you to find potentially encrypted (not poorly obfuscated) files or chunks of files. For example, here's a rule that will look for a resource has a high entropy value and a low monte carlo pi error percentage (and thus likely encrypted):

```
// Totally untested...
rule encrypted_resource {
  condition:
    for any i in (0..pe.number_of_resources - 1):
      (math.entropy(pe.resources[i].offset, pe.resources[i].length) > 7 and
       math.monte_carlo_pi(pe.resources[i].offset, pe.resources[i].length) <= 0.06)
}
```

libyara/modules/math.c
```diff
@@ -0,0 +1,496 @@
...
+
+    if (block->base + block->size > offset + length)
+      break;
+  }
+
+  if (!past_first_block)
+    return_double(UNDEFINED);
+
+  sum = sum / (double) total_len;
+  return_double(fabs((sum - 127.5) / 127.5));
```
plusvic: I find this function confusing, because it's called arithmetic_mean but it actually returns a deviation percentage from the mean. What I propose is creating a function named "deviation" receiving the string (or offset, lenght pair) and a number between 0 and 255 which will be the central point. The function will return the deviation defined by:

image

Where m(X) is our central point. Use the mean as m(X) and you'll have the standard deviation.

This will give you what you want in a more standard way that people familiar with statistical language will recognise. I would also rename "arithmetic_mean" to just "mean", to keep names as shorter as possible.

wxsBSD:
I thought going with just "deviation" and "mean" is a bit confusing. To make it clear I ended up on ```mean_err``` and ```mean_deviation```.

```mean_err``` takes the mean as an argument (must be a double) and returns the percentage difference between the argument and the calculated mean.

```mean_deviation``` takes the same argument (again, must be a double) and returns the standard deviation of the input against that value.

Here's some examples of it in action. I fed the output of /dev/random to a file and then fed that file into YARA with these rules.

```
rule mean_err {
  condition:
    math.mean_err(0, filesize, 127.5) < 0.07
}

// TODO: Find out a way to say (64 +/- 0.1)
rule mean_deviation {
  condition:
    math.mean_deviation(0, filesize, 127.5) >= 63.9 and
    math.mean_deviation(0, filesize, 127.5) <= 64.1
}
```

You'll notice that deviation is a bit cumbersome to use right now because it really only works by saying "the result falls within this range", which means the deviation is calculated twice. What do you think about  adding another argument which is the +/- value and making the function return 1 or 0. For example:

```
rule mean_deviation {
  condition:
    math.mean_deviation(0, filesize, 127.5, 63.9, 64.1) == 1
}
```

This would calculate the deviation using 127.5 as ```m(X)``` and then return true if ```63.9 <= result <= 64.1```.

**Comment by plusvic** [comment]
But why keeping both functions mean_err and mean_deviation? Both functions give a notion about the same thing, how far are the numbers in the data from the given mean. I propose the deviation because is the standard way to do that. Besides, the "mean_err" name suggests something related to the standard error of the mean, which is a very different statistical concept. So, my proposal is:

math.mean(data): 
  Returns the mean for data

math.deviation(data, mean):
  Returns the mean deviation of data from the provided mean.

math.deviation(data, math.mean(data)):
 This would return the standard deviation, which is the deviation from the actual mean.

To determine if the deviation is in a range, instead of passing two additional arguments to the function, we can create another function, like:

math.in(value, lower_bound, upper_bound)

So you can say:

math.in(math.deviation(data, mean), 63.9, 64.1)

We can even consider changing the grammar to accept expressions, like:

63.9 < math.deviation(data, mean) < 64.1

But I haven't thought about the implications yet, perhaps is not that easy to do. The idea is creating a set of simple functions that can be mixed together in different ways for different purposes, instead of having too specialized  functions. Just like the UNIX philosophy for command-line tools. What do you think?

**Comment by wxsBSD** [comment]
I agree with everything, and will make these changes. My desire to keep a mean_err is because when I started down this path my discussions with people were all centered around that instead of a standard deviation. I just thought it might be easier for people who already have an understanding of these tests to determine how far they are from the mean of a pseudo-random byte sequence as a percentage rather than have to switch over to using a standard deviation.

You are right though that standard deviation is a more widely used and understood mechanism for describing this, so I will make the changes.

I thought about changing the grammar but decided against it in the interest of time. I'll try and get all these changes done tonight.

**Comment by wxsBSD** [comment]
I think this should address all your concerns. I had to go with ```in_range``` as a function name since ```in``` is a reserved keyword. Also, ```in_range``` is an inclusive check.

I decided to leave the MEAN_BYTES constant in there (not the best name though) so people can use it as the expected mean for a random distribution of bytes. This way you can do things like:

```
rule in_range {
  condition:
    math.in_range(math.deviation(0, filesize, math.MEAN_BYTES), 63.9, 64.1)
}
```

To see if your sample is within +/- 0.1 of one standard deviation of a random distribution, and is thus likely random. I think that's how it works?

plusvic:
Perfect, if you don't have any pending change I'll merge it.

wxsBSD:
I don't have anything pending. Thanks for the feedback!

PR #232, opened 2015-01-04
Add all known Machine types.

wxsBSD:
Expand the check for valid machine types to include all the known ones.


wxsBSD:
I need to read the specs again but I think there may have a problem elsewhere when this check is expanded. In particular you seem to use it in a few places which I think may be incorrect now that accepted values are expanded:

```
  if (pe->header->FileHeader.Machine == IMAGE_FILE_MACHINE_AMD64)
    result = &((PIMAGE_NT_HEADERS64) pe->header)->
        OptionalHeader.DataDirectory[entry];
  else
    result = &pe->header->OptionalHeader.DataDirectory[entry];
```

I'm going to check through things a bit more but I think we should be using ```IMAGE_NT_OPTIONAL_HDR32_MAGIC``` and ```IMAGE_NT_OPTIONAL_HDR64_MAGIC``` magic. I should be getting my hands on some ARM PE binaries early this week, and I hope at least one of them is PE32+, so I can properly test this.

For now, I wouldn't merge this until I can verify that the problem actually exists.

wxsBSD:
I've fixed the incorrect usage of IMAGE_FILE_MACHINE_AMD64 everywhere I could find it. I've tested this by walking through the various checks in a debugger while using an ARM64 file as input. If you would like the file please let me know, I'll be happy to mail it to you.

PR #233, opened 2015-01-09
Section characteristics and section_index_addr

wxsBSD:
Expose flags for characteristics. The list of flags is from section 3.1 of the specification but in the
interest of keeping the list short I stripped out all the reserved or object file only values.

Provide a function which behaves like section_index but takes an integer as an argument and returns the section index where that offset lives. When scanning a process it will assume the argument is an RVA relative to base address and perform the addition for you.


wxsBSD:
These changes allow you to do things like this:

```
// Entry point in .data section, that section is RW and contains initialized data.
rule section_test {
  condition:
    pe.sections[pe.section_index(pe.entry_point)].name == ".data" and
    pe.sections[pe.section_index(pe.entry_point)].characteristics & pe.SECTION_CNT_INITIALIZED_DATA | pe.SECTION_MEM_READ | pe.SECTION_MEM_WRITE) == (pe.SECTION_CNT_INITIALIZED_DATA | pe.SECTION_MEM_READ | pe.SECTION_MEM_WRITE)
}
```

When combined with ```hash``` or ```math``` modules it makes for very interesting possibilities.

wxsBSD:
I am not able to test this scanning a process on a windows machine. I don't have a way to build it on windows.

wxsBSD:
I've got a question on section_index functions. Should they return -1 if the index can not be found? This allows you to write rules like this:

```
// First match of $b should not be in any section.
rule test_not_in_2 {
  strings:
    $b = "This program cannot be run in DOS mode"
  condition:
    pe.section_index(@b[1]) < 0
}
```

Right now they return UNDEFINED, which seems like the wrong thing to do if you want to be able to say "there is no section that contains that address" or "there is no section which has that name".

wxsBSD:
Any thoughts on making section_index return something other than UNDEFINED?

plusvic:
At first I though it would be a good idea, then I started to think about implementing a function is_undefined(X) which will return True if X is undefined, it could be useful in other situations too. I didn't take any decisions yet. What do you think about it?

wxsBSD:
I think that would be fine and provide increased flexibility.

PR #254, opened 2015-02-10
Fix unmapped memory access in PE parser

a0rtega:
This commit fixes a crash that takes place when trying to access an invalid memory location in PE parser.

A poc (proof of crash) file can be provided if necessary.

In addition, I noticed that in this source file, "fits_in_pe" check is always done over the struct and not the char pointer contained in it, even when the string is used later:

```
if (!fits_in_pe(pe, version_info, sizeof("VS_VERSION_INFO")))
    return;

  if (strcmp_w(version_info->Key, "VS_VERSION_INFO") != 0)
    return;
```

Also as strcmp_w is a wide string, sizeof should be multiplied by two, right?

These two points could open the door to more crashes.

Cheers!

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -633,10 +633,8 @@ void pe_parse_version_info(
   if (!struct_fits_in_pe(pe, string_file_info, VERSION_INFO))
     return;
 
-  if (!fits_in_pe(pe, string_file_info, sizeof("StringFileInfo")))
-    return;
-
-  while(strcmp_w(string_file_info->Key, "StringFileInfo") == 0)
+  while(fits_in_pe(pe, string_file_info, sizeof("StringFileInfo")) &&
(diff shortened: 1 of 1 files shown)


wxsBSD:
I'd like to see the test file for this. Feel free to drop it to me in an email if you don't want to post a link to it.

a0rtega:
Forget about "sizeof * 2" sentence I said, just had a look at strcmp_w.

moritzkroll:
@a0rtega: You were right about the "sizeof * 2", I added it to my version of the fix :) You weren't by chance looking at the same Tinba sample like me?
But also the pointer has to be corrected. We have to ensure, that the strcmp_w call on version_info->Key will be valid. Thus the check has to be

    if (!fits_in_pe(pe, version_info->Key, sizeof("VS_VERSION_INFO") * 2))

Same for the other strcmp_w call, of course.

plusvic:
I've merged mortizkroll's patch, so I'm closing this one.

PR #256, opened 2015-02-11
PE module: Fix unchecked access to version info buffers

moritzkroll:
Fixes crash (at least on Windows) for
https://www.virustotal.com/en/file/a906103c7b972e7adf8ea9c736ca44373949480576d7fc6042c495c47e9bbff5/analysis/1423694573/
and adds additional safety for other cases like unterminated keys

### DIFF
--- a/libyara/include/yara/strutils.h
+++ b/libyara/include/yara/strutils.h
@@ -59,8 +59,9 @@ void* memmem(
 #endif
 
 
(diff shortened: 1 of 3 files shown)


moritzkroll:
This is a more complete fix for https://github.com/plusvic/yara/pull/254

PR #262, opened 2015-02-21
Move the stack used by exec.c into YARA_CONTEXT

ckane:
Modify the behavior so that it allocates space from heap memory, providing options for larger stack size allocation. Also moved the stack size #define into yara/limits.h


plusvic:
Already implemented in 66229acfb7c1baa583b205193ffde7500b7edeeb

PR #263, opened 2015-02-21
Add workaround for link failure (undefined reference to '_time32') when ...

hillu:
...cross-compiling with MinGW

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,6 +20,12 @@ AC_PROG_YACC
 LT_INIT
 AC_PROG_LIBTOOL
 
+AC_CANONICAL_HOST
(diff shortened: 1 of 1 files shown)


hillu:
When trynig to link libyara.a produced using

    ./configure --host=i686-w64-mingw32

on a Debian/unstable system, I got this error `undefined reference to '_time32'`. The mingw-w64 documentation is not very useful, but several mailing list discussions suggested defining `__MINGW_USE_VC2005_COMPAT` for other linker failures.

Works for me and since this is only enabled when building 32-bit MinGW targets, it should not affect anybody else.

PR #268, opened 2015-02-26
Fix for running 'configure' in another folder

axqd:
To run 'configure' in another folder instead of the source tree, we need to specify path to 'include' folder and 'libyara.sym' symbol file.


hillu:
Looks good to me.

PR #275, opened 2015-03-09
fixed build on freebsd 10 with a conditional #define so that warnings ar...

kai5263499:
...e not introduced in other builds


wxsBSD:
I've been using a similar-ish patch for the FreeBSD port.

https://svnweb.freebsd.org/ports/head/security/yara/files/patch-libyara__proc.c?revision=379102&view=markup

PR #276, opened 2015-03-11
Do not recompute match chain lengths

cblichmann-google:
Hi Victor,

Please merge this patch that terminates the recursion early in _yr_scan_update_match_chain_length() if the match is already updated. This avoids an exponential blow-up in that function when scanning for long hex signatures with wildcards.

Cheers,

Christian


plusvic:
Great! I'll just add an initialization to make sure that chain_length doesn't have a random value the first time it gets updated.

PR #277, opened 2015-03-11
PE module: Add lower boundary to access bounds check.

hillu:
On a 32-bit Linux system, I witnessed a crash around
pe_parse_import_descriptor() (pe.c:867) because strndup() wanted to
read from an invalid address that was the result from an integer
overflow and turned out to be less than pe->data.

Since that overflow does not occur on 64-bit systems, they are not
affected.

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -92,7 +92,8 @@ limitations under the License.
 
 
 #define fits_in_pe(pe, pointer, size) \
-    ((uint8_t*)(pointer) + size <= pe->data + pe->data_size)
+    (pe->data <= (uint8_t*)(pointer) && \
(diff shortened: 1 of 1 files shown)


plusvic:
This change doesn't fix the problem in all cases. "pointer" could be greater then pe->data and still overflow if "size" is too large. A more comprehensive solution is:

#define fits_in_pe(pe, pointer, size) \
    (size <= pe->data_size && \
     (uint8_t*)(pointer) >= pe->data && \
     (uint8_t*)(pointer) + size <= pe->data + pe->data_size)

plusvic:
Fixed in 223daaf9c4d335b87805654466fba32f3988ad5a

PR #279, opened 2015-03-12
Increase limits and dynamically allocate stack.

shanehuntley:
Limits changes for long yara files and complex regexps.
Dynamically allocate stack in exec.c to reduce stack usage.


plusvic:
Hey Shane, take a look at commit 21e796dfb6e217457de92c9c9b1e526a0f5df11f, the stack wasn't released when function returned early due to a timeout or error in function call. You may want to fix it in your branch too.

Issue #260, opened 2015-02-15
Don't hide min and max with macros


JusticeRage:
The following is taken from `utils.h`:
```C
#ifndef min
#define min(x, y) ((x < y) ? (x) : (y))
#endif

#ifndef max
#define max(x, y) ((x > y) ? (x) : (y))
#endif
```

When C++ programs using the STL are compiled with Yara, std::min and std::max are hidden which causes compilation errors.

Windows headers have the exact same problem and have resorted to the following solution. I'm not sure if it's adequate here:
```C
#ifndef NOMINMAX
#ifndef max
#define max(a,b) (((a) > (b)) ? (a) : (b))
#endif
#ifndef min
#define min(a,b) (((a) < (b)) ? (a) : (b))
#endif
#endif /* NOMINMAX */
```

axqd:
+1. Same here. I have to change the order of '#include' to fix it.

wxsBSD:
Same here, oddly enough the problem is only a warning on OS X with clang but Ubuntu 14.04 with gcc 4.8.2 is an error.

shanehuntley:
+1. I rename to yr_min and yr_max which might be the cleanest solution and will stop the next person trying to link to C++ hitting the same errors.

plusvic:
I've renamed min and max to yr_min and yr_max in commit fc4696c8b725be1ac099d340359c8d550d116041. The issue with get_object is still pending.

PR #280, opened 2015-03-13
Implement NOMINMAX.

wxsBSD:
As discussed in #260 implement a NOMINMAX check around the min and max
macros.

This allows c++ code which includes yara.h to #define NOMINMAX before
the include and #undef it after.

### DIFF
--- a/libyara/include/yara/utils.h
+++ b/libyara/include/yara/utils.h
@@ -36,6 +36,7 @@ limitations under the License.
 #define YR_API
 #endif
 
+#ifndef NOMINMAX
 #ifndef min
 #define min(x, y) ((x < y) ? (x) : (y))
 #endif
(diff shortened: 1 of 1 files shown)


wxsBSD:
I'm not sure I like this solution much, but it does allow me to build c++ code which includes yara.h by putting a ```#define NOMINMAX``` before the include and ```#undef NOMINMAX``` after. I wanted to change the macros to be ```yr_min``` and ```yr_max``` but I'm not sure how accepted that woudl be. If that is the preferred solution just let me know and I will make the change and sweep the tree to clean up any references.

plusvic:
I've renamed min and max to yr_min and yr_max in commit fc4696c8b725be1ac099d340359c8d550d116041. That's probably the better solution.

PR #284, opened 2015-03-17
Fix regression in pe_collect_resources().

wxsBSD:
It appears that after 223daaf9c4d335b87805654466fba32f3988ad5a no
resources were properly collected. Fix it by passing in the correct
offset, which is pe->data + offset.


wxsBSD:
This is easily reproducible. A rule as simple as ```pe.number_of_resources == 5``` when run against ```0861e9bc073d37f0eeb7a7323553f2ca``` does not match. I did some light testing with other code that relies upon ```fits_in_pe()``` and didn't spot any obvious problems.

PR #294, opened 2015-03-31
Fix PRINT_MODULE_DATA.

wxsBSD:
This fixes things so that PRINT_MODULE_DATA works correctly.

### DIFF
--- a/libyara/object.c
+++ b/libyara/object.c
@@ -1095,11 +1095,22 @@ void yr_object_print_data(
 
     case OBJECT_TYPE_STRING:
       if (((YR_OBJECT_STRING*) object)->value != NULL)
+      {
         printf(
-            "%s%s = \"%s\"\n",
+            "%s%s = \"",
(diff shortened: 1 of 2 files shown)


wxsBSD:
No clue why I didn't see it earlier but I was looking to see if this existed before I wrote my own implementation, glad I found it. :smile: This is a great feature!

wxsBSD:
I also changed string printing so that non-printable characters are printed in hex form.

wxsBSD:
Would you be open to changing it from a compile-time knob to a run-time option? Having this ability is a great way to debug signatures. I can keep a binary with it turned on around but I suspect making it a run-time option is going to increase usage and by extension the new module features!

plusvic:
Sure, I think that a run-time option is a better way to control this feature. Do you plan to implement it?

wxsBSD:
Yes. I will try and implement it in the next couple of weeks.

PR #305, opened 2015-04-30
Ensure thread ids are globally unique even if multiple rules instances.

shanehuntley:
Fixes multi-threaded issue where if you instantiate multiple instances of YARA_RULE objects you will get thread id collisions causing race conditions and occasional segfaults with module unload.


plusvic:
A simpler solution in commit ef05a686c6962dba1b721535aa06b0a5cb25068e

Issue #271, opened 2015-03-06
No matching with MIPS e_machine in module elf


nyx0:
I have no matching with a rule that should match to MIPS binary using elf module.
It works perfectly with x86 but not with MIPS.

``` yara
$ cat rules.yar
import "elf"

rule elf_x86
{
    condition:
        elf.machine == elf.EM_386
}
rule elf_mips
{
    condition:
        elf.machine == elf.EM_MIPS
}

$ readelf.py -h ls_x86 | grep Machine
    Machine:                           Intel 80386

$ readelf.py -h ls_mips | grep Machine
    Machine:                           MIPS R3000

$ yara -s rules.yar .
elf_x86 ./ls_x86
```

nyx0:
There is the same problem on ARM.

wxsBSD:
Can you share the file? I'll be happy to see if I can fix this.

nyx0:
Sure -> http://we.tl/8nf2cbyZOD (d188ecabd3219b2df1318f3b0385c20a1569cc77) you will find the yara rules and 3 binaries for MIPS, x86 and ARM (ls binary)

wxsBSD:
There's at least two things going on, neither of which I can fix right now - though I may be able to in a week or so.

1. The MIPS binary is big endian so the machine value (and others) is swapped. Should be as simple as forcing them to be little endian.

2. The ARM binary is little endian but I believe the definition in elf.h is incorrect. It currently uses:

```
#define ELF_EM_ARM      0x0032  // ARM
```

According to wikipedia (http://en.wikipedia.org/wiki/Executable_and_Linkable_Format) this is for IA-64, and the correct value should be 0x28.

Also, might as well expand the machine definitions to include AARCH64 (0xb7).

Hopefully someone else gets to this before I do, otherwise I will try to circle back around the end of this week. Sadly, I'm not as familiar with the ELF specification as other file formats, otherwise I would feel more comfortable about changing these quickly.

PR #309, opened 2015-05-09
Fix the ARM architecture value

nyx0:
This commit fix the first part of #271

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,7 @@ stamp-h1
 # Project specific files
 /yara
 /yarac
+
+# Linux and Mac files
+*.swp
+.DS_Store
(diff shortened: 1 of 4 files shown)


nyx0:
I also add "new" architectures to support according to @wxsBSD in #271 (MIPS RS3 LE, PPC, PPC64, AARCH64).
For the big endian issue, I wrote the following function but I don't know if it's the best way to do that:

```C
int force_litle(uint8_t elf_machine){
  elf_ident_t* elf_ident;
  
  if (elf_ident->data == ELF_DATA_2LSB)
    return elf_machine;
  else if (elf_ident->data == ELF_DATA_2MSB)
    return elf_machine >> 8;
  else
    return elf_machine;
}
```

PR #311, opened 2015-05-10
PE module: Fix bounds check for iterating over certificates

hillu:
Found using afl-fuzz


plusvic:
Fixed in 5f02d37a8e50a7f53b09216b08dfc530b7940be7. Didn't included the fprintf statement.

PR #314, opened 2015-05-12
PE module: Do not use strlcpy()-type functions

hillu:
strlcpy() reads from the source buffer until it hits a null character
to determine the string length. This may lead to crashes due to
out-of-bounds reads.

Since we don't use the string length, using strncpy() and making sure
that the target buffer is null-terminated is safer.

### DIFF
--- a/libyara/include/yara/strutils.h
+++ b/libyara/include/yara/strutils.h
@@ -68,7 +68,7 @@ int strcmp_w(
     const char* str);
 
 
-size_t strlcpy_w(
+char* strncpy_w(
     char* dst,
     const char* w_src,
     size_t n);
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -709,8 +709,10 @@ void pe_parse_version_info(
             char key[64];
             char value[256];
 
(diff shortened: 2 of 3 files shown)


hillu:
A bit of demonstration of `strlcpy` badness: I can't find the sample that caused an actual crash, but...

Running
<pre>valgrind yara pe.yar 34209bcf50eaecee0af359c8f397450f</pre>
on yara 3.4 (on an x86_64 Debian/unstable machine) gives me an error: "Invalid read of size 1" at `strlcpy`.

The rules file only contains `import "pe"`, the sample with md5sum=34209bcf50eaecee0af359c8f397450f is available on VirusTotal.

hillu:
md5sum=1d161d9a10f76acdddf3fa3811e78156 is another sample valgrind gets upset about.

plusvic:
In commit 14dbf62b132c1d7294fa3b4a535d5fec3f6246b9 I'm using strncpy as you propose to replace the specific line causing this bug. But I'm maintaining strlcpy_w in the other cases, which are protected against buffer overruns by calls to wide_string_fits_in_pe.

Issue #321, opened 2015-06-26
memory-mapped file access can lead to crash due to uncaught exception/signal


hillu:
Yara crashes on Windows and Linux systems while scanning files that cannot be read due to filesystem corruption: The memory mapping operation (CreateFileMapping()+MapViewOfFile() or mmap()) was successful, but an exception (Windows: EXCEPTION_IN_PAGE_ERROR, Unix: SIGBUS) is thrown as soon as data that cannot be read from disk is accessed. This is documented behavior -- see [Reading and Writing From a File View](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366801%28v=vs.85%29.aspx) and the mmap(2) manpage. In my opinion, libyara should catch those errors because how files are accessed is an implementation detail that the user does not need to be concerned with.

On Unix systems, it is possible to trigger SIGBUS by truncating the file while yara is scanning. I have also been able to trigger EXCEPTION_ACCESS_VIOLATION (instead of EXCEPTION_IN_PAGE_ERROR) in a yara process running under Wine by doing the same. I have not yet found an easy way to trigger this problem on a "real" Windows system.

hillu:
In the meantime, I have found a way to reproduce the problem on Windows: If a removable volume (such as a CD-ROM drive) is removed while yara is scanning a large file on that volume, it will crash.

plusvic:
Fixed in db22ffda588b9b3abc619f8a2f616d0be57e9335

PR #322, opened 2015-06-26
Add exception handling when scanning memory blocks

hillu:
Unix: signal handler for SIGBUS + sigsetjmp/siglongjmp
Win32: vectored exception handler + setjmp/longjmp

This should eliminate crashes when accessing the contents of files
that turn out to be unreadable after they have successfully beea
mapped into memory.

This can occur if the mapped file is truncated (Unix) or if the
filesystem is corrupted.

### DIFF
--- a/libyara/exception.h
+++ b/libyara/exception.h
@@ -1,0 +1,96 @@
+#ifndef _EXCEPTION_H_
+#define _EXCEPTION_H_
+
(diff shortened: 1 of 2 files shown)


hillu:
This is a fix for #321.

plusvic:
Great job!

Issue #71, opened 2013-11-24
Use tags in rule conditions


plusvic:
_From [golgotr...@gmail.com](https://code.google.com/u/114188526611906850564/) on January 29, 2013 11:44:46_

I think that It will be very useful if we can reference rule tags in the condition of the rule like in this example:

rule banker_1 : malware
{
  strings:
   $a = "bankofamerica.com" 
  condition:
   $a and 2 of :banker: and 2 of :suspicious: and :packers:
}

To hit banker_1 first need to hit 2 rules with banker tag and one rule with suspicious tag and one from packers.

rule Trojan_Agent : malware
{
  condition:
   4 of :suspicious:
}

I think it will be handy to discover unknow malware

_Original issue: http://code.google.com/p/yara-project/issues/detail?id=71_

dkindlund:
Developing a full-fledged rules engine in YARA may cause significant performance/resource issues, long term.  I'd recommend keeping YARA fast/agile.  There are plenty of ways to achieve this type of logic as a post-processing operation (after YARA results are provided).

PR #333, opened 2015-07-28
Allow tags to be used in conditions.

wxsBSD:
This PR comes from discussion and code done in collaboration with Mike Goffin (@mgoffin). The idea is to be able to use tags in conditions. For example:

```
rule foo: mytag {
  condition:
    true
}

rule bar: mytag {
  condition:
    true
}

rule baz {
  condition:
   2 of mytag
}
```

In this example the baz rule will match because at least 2 other rules that match have the tag of mytag.

We believe this is a useful feature to have because it allows you to tag rules with specific things (eg: backdoor name, actor name, etc.) and then be able to operate on them as a grouping.

This is just an initial attempt and can likely be optimized, but we wanted to open this PR to see what you think about it. If you think this is a good idea we are more than willing to address any concerns you have.

Because the use of arenas is still uncertain to us both we were hesitant to try to optimize this. One area we were thinking of doing is finding a way to only store pointers to the rules which have a given tag, potentially by using the tag as a key in a hash table and the value being a linked list of pointers to rules with that tag. This would allow us to optimize the runtime behavior in OP_PUSH_TAG so that it doesn't have to iterate over every rule. We aren't entirely sure how to store such a hash table in an arena, so we decided to at least get your input on it and maybe some pointers as to how you wish to proceed.

Here is the rest of the details on how it works:

When a compiler is created a hash table (tags_table) is created which stores a tag as the key. When a rule is reduced via yr_parser_reduce_rule_declaration_phase_2() each tag for the rule is
stored in the tags_table. The value is not important as we only care if the tag exists or not.

There is a new function called yr_parser_store_tag() which will write the tag identifier to the sz_arena and set the "tag" argument to point to the newly written data. If the tag identifier is not known - ie: it
wasn't added to tags_table in yr_parser_reduce_rule_declaration_phase_2() - then an UNDEFINED_IDENTIFIER is returned.

When an identifier is handled in grammar.y and it is not an object or a rule then it must be a tag. The tag is stored in sz_arena using yr_parser_store_tag() and two instructions are emitted:

OP_PUSH 1
OP_PUSH_TAG(tag)

When using the "N of tag" syntax the following instructions are emitted:

OP_PUSH N # Handled by for_expression
OP_PUSH_TAG(tag).

Add the handler for OP_PUSH_TAG instruction, which will get the tag as an argument, pop the number of rules to match and walk each tag of matching rules until at least N rules have matched.

### DIFF
--- a/libyara/compiler.c
+++ b/libyara/compiler.c
@@ -59,6 +59,9 @@ YR_API int yr_compiler_create(
   result = yr_hash_table_create(10007, &new_compiler->rules_table);
 
   if (result == ERROR_SUCCESS)
+    result = yr_hash_table_create(10007, &new_compiler->tags_table);
+
+  if (result == ERROR_SUCCESS)
     result = yr_hash_table_create(10007, &new_compiler->objects_table);
 
   if (result == ERROR_SUCCESS)
@@ -123,6 +126,8 @@ YR_API void yr_compiler_destroy(
   yr_hash_table_destroy(
       compiler->rules_table,
       NULL);
+
+  yr_hash_table_destroy(compiler->tags_table, NULL);
 
   yr_hash_table_destroy(
       compiler->objects_table,
--- a/libyara/exec.c
+++ b/libyara/exec.c
@@ -176,6 +176,7 @@ int yr_execute_code(
 
   char* identifier;
   char* args_fmt;
+  char* tag;
 
   int i;
   int found;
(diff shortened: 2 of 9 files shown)


wxsBSD:
As I said above, we don't believe this is quite ready to be merged but are trying to circulate the idea and get feedback so we can improve upon it if the feature is deemed useful.

plusvic:
This feature is very interesting, and someone else already asked for it (see issue #71). But I'm thinking about introducing the concept of iterators in YARA, for example to be able to do this: 

for any section in pe.sections : ( section.name == ".text")

This could change the way in which this feature is implemented, because the "x of <whatever>" could be a general purpose construct where <whatever> is an iterator returning boolean values on each iteration. So, I prefer delaying this feature until iterators are implemented.

wxsBSD:
Iterators will certainly change everything so I agree on holding off until the iterator work is in. I'll be keeping an eye out for that and submit a new PR once @mgoffin and I have reworked it.

mgoffin:
As a side-thing for when iterators are in, is the path we are on a good path to continue? @wxsBSD and I have both discussed potential ways to optimize this, one which revolves around creating a `tags` arena and putting a hash table in it. Not sure if that's possible or even the right solution. Would you be able to point us in the right direction so we can plan ahead? :)

mgoffin:
Curious if iterators have made their way into the code base so that we can revisit this feature?

PR #334, opened 2015-07-29
Fix Error_on_warning in yara-python

frbapolkosnik:
Without this patch it was impossible to supress error on warning in python because it was set before the check, and nothing was done to unset it.

### DIFF
--- a/yara-python/yara-python.c
+++ b/yara-python/yara-python.c
@@ -1481,8 +1481,6 @@ static PyObject* yara_compile(
     if (error != ERROR_SUCCESS)
       return handle_error(error, NULL);
 
-    yr_compiler_set_callback(compiler, raise_exception_on_error, NULL);
-
     if (error_on_warning != NULL)
     {
       if (PyBool_Check(error_on_warning))


frbapolkosnik:
Thanks for pointing that out, I misread the code, a bit. Nevertheless
[code]
 try:
                rules = yara.compile(source=data)
            except yara.SyntaxError:
                message = "Not a valid yara rules file: %s" % sigfile
[/code]
throws an exception on a ruleset that produces only warnings when compiled with yarac.

When I changed the code to:
[code]
rules = yara.compile(source=data,error_on_warning=True )
[/code]
Throws:
WarningError: Using deprecated "entrypoint" keyword. Use the "entry_point" function from PE module instead.

plusvic:
I'm trying to reproduce the issue you mentioned, but everything works fine for me:

```
data = 'rule test { strings: $a = "dummy" condition: $a at entrypoint }'
rules = yara.compile(source=data)
```

In this case I got no exceptions, even if the rule raises a warning.

```
data = 'rule test { strings: $a = "dummy" condition: $a at entrypoint }'
rules = yara.compile(source=data, error_on_warning=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
yara.WarningError: Using deprecated "entrypoint" keyword. Use the "entry_point" function from PE module instead.
```

In this other case I got the exception as expected. Can you share the rules causing the issue? Maybe it's caused by another kind of warning.

frbapolkosnik:
I mistakenly thought that it was an yara-python issue. Thank you for all your great work!

PR #336, opened 2015-08-06
Java bindings

papostolescu:
Java wrapper, works in two modes:
- embedded: 
   - yara shared library is loaded and executes in process
   - the shared library is loaded from the jar resources
   - binding code uses efficient JNI calls
- external: 
    - instead of going against the shared library the wrapper uses the yara executable out of process
    - yara executable is extracted from the jar file resources
    - requires a small patch (part of the pull request) to escape yara output quotes (otherwise some data cannot be parsed out reliably)

The bindings expose a fully working end-to-end flow from rule loading to compiler creation and execution with callbacks.

Some functionality is missing but I'm planning to contribute it in time.


rferreira:
Hey folks any chance this PR can get a review?

plusvic:
Hi Rafael,

I'm very busy with other projects, but as soon I can I'll take a look your
pull request. I think it would be great to be able to use Yara from Java,
and many people could benefit from this contribution.

Cheers,
Víctor

hillu:
Given that this patch does not seem change any YARA core functionality, wouldn't it be a good idea to publish the Java bindings in a separate repository and document the requisites to build from source?

papostolescu:
There is only one change required in yara.c to escape rules metadata when using external (out of process) yara execution.  Source is built using maven, has only one compile-time dependency and the bindings can be made available on one of the public maven repos (like maven central) so there will be no need for direct compilation.

It's up to the maintainer(s) to decide how to handle this, I'll make the necessary changes.

plusvic:
Hilko is probably right, this could be maintained in a separate repository
instead of the main YARA repo. I can put a link to that repository for
people to notice that they can use YARA from Java too. Including the Java
bindings in the YARA repository has the drawback that we are forced to keep
them up-to-date whenever we launch a new YARA version, so I think it's
better to keep it as a separate project.

rferreira:
hey folks, for full disclosure that is not my patch, I'm just an interested party in its status :)

papostolescu:
I've moved the bindings only into a separate repo here: https://github.com/papostolescu/yara-java

There is a separate pull request for the patch required to parse metadata output: https://github.com/plusvic/yara/pull/345

PR #343, opened 2015-09-10
Override pe.imports().

wxsBSD:
Sometimes you only care if a DLL is used, not a specific function from
it. In this case you can use the new version of pe.imports() which takes
a single string which is the DLL name.


wxsBSD:
I also added an imports_ordinal() which will check ordinal imports. This commit message has more details:

https://github.com/wxsBSD/yara/commit/e301fbcef4a50a962b247c6281e5bc67b76c8f46

PR #345, opened 2015-09-14
escape metadata values

papostolescu:
escape matched rules metadata values when printing, for example when the following rule matches 

rule EscapeMeta {
  meta:
      name = "this is a \"test\""
  strings:
      $a = "test"
   condition:
      $a
}

the new output will be

     EscapeMeta [name="this is a \"test\""] test.yara

instead of

     EscapeMeta [name="this is a "test""] test.yara

### DIFF
--- a/yara.c
+++ b/yara.c
@@ -400,6 +400,39 @@ void print_string(
   printf("\n");
 }
 
+void print_escaped(
+        char* str,
+        int length)
+{
+  for (int i = 0; i < length; i++) {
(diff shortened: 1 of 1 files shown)


hillu:
I have provided an alternative implementation in <https://github.com/hillu/yara/tree/escape-meta>.

plusvic:
I've used Hillu's implementation in c3e7487738489939213ef82d488c628430bf828e

PR #347, opened 2015-09-22
Crypto crash fix

wxsBSD:
Fix a crash in OpenSSL due to not properly initializing with respect to threading.

### DIFF
--- a/libyara/libyara.c
+++ b/libyara/libyara.c
@@ -23,6 +23,10 @@ limitations under the License.
 #include <yara/modules.h>
 #include <yara/mem.h>
 
+#ifdef HAVE_LIBCRYPTO
+#include <openssl/crypto.h>
+#endif
+
 #ifdef _WIN32
 #define snprintf _snprintf
 #endif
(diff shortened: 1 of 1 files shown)


wxsBSD:
I'll shoot you a mail with exact hashes and signatures that reproduce this. My laptop is currently burning through another round of testing but I am fairly confident this fixes the problem.

libyara/libyara.c
```diff
@@ -112,6 +143,13 @@ YR_API int yr_finalize(void)
 {
   yr_re_finalize_thread();
 
+  #ifdef HAVE_LIBCRYPTO
```
plusvic: This should appear after the init_count > 0 check

wxsBSD: Fixed! It was a bad merge on my part. Thanks for catching it! :)

libyara/libyara.c
```diff
@@ -115,6 +145,13 @@ YR_API int yr_finalize(void)
   if (--init_count > 0)
     return ERROR_SUCCESS;
 
+  #ifdef HAVE_LIBCRYPTO
+  int i;
```
plusvic: Just a small note: I recently started to comply with the C89 standard. Declaring "i" in the middle of the function is not accepted by C89, the declaration should be moved to the beginning of the function.

wxsBSD:
I've got it fixed locally. Will push whenever I'm back on stable wifi, which may be as late as Monday. I'll try to keep this in mind while I work on my next PR. 😛

Issue #337, opened 2015-08-07
Imphash not hitting


jcannell:
Hi,

I was trying to write a rule that used an imphash today. I have the right imphash according to VT and pefile. Here is my rule

import "pe"

rule cryptowall_1 : ransomware
{
   meta:
      author      = "Joshua Cannell"
      description = "Cyptowall Ransomware 3.0, Visual C++"

   condition:
      pe.imphash() == "251177b9dbdc928d5ac0e23bfe73ad1d"

}

When i run the rule against my file (sha256: 403a80f653631d2e07c181c0e76d575ed3b875305b5bf3ccc617d70e54a79d1f) I don't get anything. The file is on VT.

Never tried imphash w/yara before today, seems like I'm doing it right, though. Am I missing something?

wxsBSD:
I can take a look at this early next week.

wxsBSD:
This was likely broken in https://github.com/plusvic/yara/commit/1cc98f75fa922614aa440ccf9c397d65db7c825c. The file in question imports more than 256 functions from RPCRT4.dll. I checked the PE specification and don't see anywhere that mentions this limit, so it seems both arbitrary and incorrect to have in YARA. I understand the need for it but it does break things in this case (and also anyone using the ```imports()``` function will potentially see incorrect results.

Adding @shanehuntley since he is the one who made the commit in question.

plusvic:
The best idea is probably increasing MAX_PE_IMPORTS to some value unlikely seen in actual PE files. Maybe 16384?

wxsBSD:
That's reasonable but still arbitrary. When parsing imports there isn't that much data used per import, so what's wrong with having no limit?

One other option is to move the import parsing to be an "on demand" thing that is done only during a function that requires it (imphash and imports). The trade off would be a longer run time but a lower memory footprint as neither function would allocate anything while parsing the import table. Just an idea...

Ultimately I can see bumping the limit as you suggest but it still seems arbitrary to me. I'd love to find an example that causes "blow up".

plusvic:
Maybe @shanehuntley can provide some problematic PE. I suspect he introduced this limit because of some malformed PEs causing thousands or even millions of iterations on that loop.

jcannell:
I like the "on demand" idea. No reason to parse the imports unless it's needed, yes?

And I think it's a good idea to have a limit, even if it seems arbitrary. I'm not real familiar with yara at the source level but I know things tend to get abused when there aren't limits :\

[...]

PR #348, opened 2015-09-22
Bump MAX_PE_IMPORTS to 16384.

wxsBSD:
Because the imports are parsed and stored for later analysis anytime we
run into the limit, imphash(), imports() and related functions would
return incorrect results when compared to other tools.

This is a compromise to bump it up from 256 to 16384 but it ultimately
doesn't fix the problem that was introduced in 1cc98f7. It seems to be a
conflict of interest between parsing accuracy and memory usage.

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -73,7 +73,7 @@ limitations under the License.
 
 
 #define MAX_PE_SECTIONS              96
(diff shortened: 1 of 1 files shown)


wxsBSD:
This fixes one of the issues raised in #337.

PR #352, opened 2015-10-01
Fix build on OS X.

### LINKED ISSUE #351: Build broken on OS X 10.11 and xcode 7.0.1
This is more an FYI for others, and I am working on it. I updated my laptop to 10.11 last night, and also updated to xcode 7.0.1. According to this link http://lists.apple.com/archives/macnetworkprog/2015/Jun/msg00025.html, Apple has finally killed off the OpenSSL headers in their latest xcode. They still ship the libraries but no longer ship headers. This breaks the build for obvious reasons.

It sounds like the right fix is to use brew provided openssl, which will be required now. I'll be working on an update to that.

wxsBSD:
With the latest Xcode on OS X, Apple removed the OpenSSL headers but
left the libraries. This meant the configure script was finding
libcrypto in /usr/lib but building things including OpenSSL headers
would fail.

Since building anything that uses OpenSSL is now broken, the only
reasonable way forward is to use OpenSSL from brew (or something else).
This commit fixes the build to look in the standard brew location for
OpenSSL for OS X.

For more information see this link:
http://lists.apple.com/archives/macnetworkprog/2015/Jun/msg00025.html

Fixes #351.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,6 +28,10 @@ case $host_alias in
      i?86-*-mingw*) CFLAGS="$CFLAGS -D__MINGW_USE_VC2005_COMPAT" ;;
 esac
 
+case $host_os in
+    darwin*) CFLAGS="$CFLAGS -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib" ;;
+esac
+
 
 AC_ARG_ENABLE([profiling],
   [AS_HELP_STRING([--enable-profiling], [enable profiling support])],


wxsBSD:
I'm not super pleased with this because I hate to hardcode a path like that, even if it is specific to Darwin. I'm not aware of any way to get brew to install an openssl-config binary that we can use to get the proper location for headers and libraries. I'm open to suggestions but this does fix the build with the latest Xcode.

plusvic:
Not a critical issue, but if you don't have those directories you'll get two warnings:

clang: warning: argument unused during compilation: '-L/usr/local/opt/openssl/lib'
ld: warning: directory not found for option '-L/usr/local/opt/openssl/lib'

PR #355, opened 2015-10-10
Make stack size run-time configurable

ckane:
Follow up to #262 that moves the logic for setting the stack size into a new -k/--stack-size commandline argument, that enables this property to be a run-time configurable. API change necessary due to it needing to be passed into the YR_SCAN_CONTEXT, similar to how timeout is handled.

This way, larger yara rules (such as 200000 strings) can be given to binary yara without a local source-code change and recompile. This means you can better minimize memory footprint where desired.


plusvic:
Great contribution. The only drawback I see is that we are modifying the API by adding a new argument to yr_rules_scan_xxx functions. This would break existing C programs relying on the API. The stack size is something that users won't be changing for individual files, so it probably makes sense to add a new function to set the stack size during YARA's initialization instead of passing the size to each call to yr_rules_scan_xxx.

This function could be flexible enough to control more settings in the future, like:

yr_set_configuration(YR_CONFIG_STACK_SIZE, <value>);

With this function we could control any other integer setting, like in:

yr_set_configuration(YR_CONFIG_MAX_STRING_MATCHES, <value>);

ckane:
I was concerned about that too, and was thinking maybe it would make sense to put this, as well as the timeout argument into a args/config struct that then has its pointer passed into init and runtime functions that use it to modify behavior.

ckane:
Then this would be a good place to stick future modifications of this nature without breaking API.

ckane:
Ok @plusvic I've rewritten the stack size logic to implement a global config array in libyara.c that could potentially be extended for other purposes in the future, and I implemented the get/set API as described above. Could you take a look and tell me if this new thing is what you had in mind? Hoping to get it considered for inclusion in 3.5.0.

ckane:
This seems ready for review now

libyara/exec.c
```diff
@@ -184,12 +183,15 @@ int yr_execute_code(
   int stop = FALSE;
   int cycle = 0;
   int tidx = context->tidx;
+  unsigned int stack_sz;
```
plusvic: Make stack size a signed integer. "sp" is signed too.

ckane:
Sounds good, done!

PR #356, opened 2015-10-14
Add support for CLI parsing.

wxsBSD:
This commit adds support for parsing the COM Runtime descriptor
directory entry, which is only the start of the rabbit hole. ;)

Ultimately this leads us to parsing the CLI MetaData streams, including
GUIDs and a convoluted table structure containing various information
relevant to the CLI.

In particular, here are the things we parse out:

```
+  declare_string("dotnet_version");
+  declare_string("dotnet_module_name");
+  begin_struct_array("dotnet_streams");
+    declare_string("name");
+    declare_integer("offset");
+    declare_integer("size");
+  end_struct_array("dotnet_streams");
+  declare_integer("number_of_dotnet_streams");
+  declare_string_array("dotnet_guids");
+  begin_struct_array("dotnet_resources");
+    declare_integer("offset");
+    declare_integer("length");
+    declare_string("name");
+  end_struct_array("dotnet_resources");
+  declare_integer("number_of_dotnet_resources");
+  begin_struct("dotnet_assembly");
+    begin_struct("version");
+      declare_integer("major");
+      declare_integer("minor");
+      declare_integer("build_number");
+      declare_integer("revision_number");
+    end_struct("version");
+    declare_string("name");
+    declare_string("culture");
+  end_struct("dotnet_assembly");
+  declare_string_array("dotnet_modulerefs");
+  declare_integer("number_of_dotnet_modulerefs");
```

Unfortunately there are a lot of manual table walking because the tables
contain variable size columns, which depend upon the number of rows in
other tables. This makes the table walking code more messy than I would
like.

And since I'm here and noticed it was unused, remove the unused
RICH_DATA structure.


wxsBSD:
This is an early version of parsing useful information out of .NET binaries. It's a bit convoluted because of the structure of the tables and their interdependence on each other. I tried to provide pointers to the various documentation used, but if you have any questions please let me know.

I've run it against a thousand or so .NET binaries I have laying around and it parsed almost all of them. Those that it could not parse correctly were corrupt in one way or another (but YARA did not crash on them). I've got a whole lot more .NET binaries laying around which I will try and make room to extract and test against.

I will also be updating this PR with the documentation necessary for this, but I wanted to push this now so others could see it.

plusvic:
Nice feature! I'm thinking about moving this to a separate "dotnet" module, given its complexity and because it only applies to a subset of PE files. We would need to move functions and macros like pe_get_directory_entry, pe_rva_to_offset, struct_fits_in_pe to pe_utils.c to use them in both "pe" and "dotnet" modules. Does it makes sense for you?

I would also try to simplify the pe_parse_com function by moving some code out of it, specially the big switches.

wxsBSD:
I did the split to a dotnet module. I'm still not 100% sure I got everything correct. I need to review it closer and then go through and split ```pe_parse_com()``` into smaller functions where it makes sense.

wxsBSD:
I refactored things so they should be more manageable now. The relevant details are in https://github.com/wxsBSD/yara/commit/c59323e3ee17bd04b933cb770e50dc549b89d260. I've put it through the same light testing as before (only a thousand or so files, including some known problematic ones) and it is looking good.

One thing to note is that now that everything is in the ```dotnet``` module I removed the ```dotnet``` from all the exposed variables.

wxsBSD:
I'm getting word my usage of ```strnstr(3)``` is not portable to most Linux (specifically I've received a report about Ubuntu 14.04). It looks like it may be available via some package, but I will have to investigate more. I really don't want to require another package so it may make sense to just include a ```strnstr(3)``` implementation from FreeBSD (https://github.com/freebsd/freebsd/blob/master/lib/libc/string/strnstr.c) for portability. Sadly, I am without a laptop for the next week or so, so I'm not sure when I will be able to address this. I should be able to get back to this sometime after November 16th.

wxsBSD:
I've only been able to put commit d508967 (switching away from strnstr(3)) through very light testing due to not having access to my normal test cases for a while more. I believe this is ready for a merge now, pending any comments from you. :)

wxsBSD:
This had been sitting for a while and was no longer merging cleanly. I've updated it to use the latest stuff from master but I'd appreciate a second look to make sure I didn't screw anything up. In particular I moved pe_rva_to_offset() over to pe_utils.c since it is used by both pe and dotnet module. There have been a number of changes in there fixing corner cases, so I want to make sure I have all those reflected in my latest update.

79617261:
Any idea when this is coming into the main branch?

plusvic:
I want to release this in the next version. I didn't include it in version 2.5.0 because its a lot of code and didn't want to compromise stability. Now is a good moment to merge it in the master branch.

plusvic:
btw @wxsBSD, no problem with changing the license to BSD, right?

wxsBSD:
No problem. As you might have guessed from my username, it is my preferred license. :)

wxsBSD:
Also, thanks for the merge!

Issue #354, opened 2015-10-07
-D option not thread safe and dump JSON?


wxsBSD:
The -D option causes multiple prints to the screen to interleave when running multithreaded. This isn't a huge deal because it isn't a commonly done operation but I'm creating this issue so it isn't lost. I have plans to tackle this after my next pull request.

One thing that would be nice is if the -D option dumped in JSON format. It would make it much easier to ingest and parse with other tools. Any thoughts on that?

moofusecurity:
+1 for JSON

79617261:
+1 to JSON too

wxsBSD:
For those playing along at home an easy work around was committed to yara-python which exposes module callback data as a dictionary which you can then serialize to json if you want.

PR #360, opened 2015-10-29
Expose module data to library user.

msuvajac:
This has elements that solve most of issue #354.
When using libyara one should have access to serialized module data if a user set the `SCAN_FLAGS_SHOW_MODULE_DATA`. The way we can expose this is to add a callback message that will forward the library user this data. This also decouples library and user code. User might want to have multiple output formats for this data and in this way we enable any user to write their own deserializer -> output.
Another issue with current approach this solves is using `printf()` inside library code which messes up user application output using libyara. We solve this by making `yr_object_print_data()` a utility function not directly called by the library code, but making it available to user which can serve also as a demonstration to users how to write deserializers.
Next issue we solve here is output thread-safety which is now handled by the library user as demonstrated with the change to yara.c.


wxsBSD:
I haven't had a chance to look at this, but thanks for submitting it. If I'm understanding your changes correctly this would allow yara-python to do the same thing? There would have to be some massaging of data to get it into an appropriate format for python, and then serialized out to JSON or something else. If so, this is probably something I can look into as I have some experience with python modules.

msuvajac:
That's right. This is just incremental change that exposes serialized module data in a convenient way. Once exposed, it's up to library user to do something with it. That includes yara-python or any other "user" of libyara.

plusvic:
I agree with the intention of this PR but not with the way it was implemented. I think that removing the printing code from the library and allow the client of the library to call yr_object_print_data himself is a good idea, it allows solving multi-threading issues and let the client decide whether he want to actually print module structures and values or do something else with them.

But I would completely remove the ``SCAN_FLAGS_SHOW_MODULE_DATA`` flag and the yr_modules_print_data function and simply introduce a new CALLBACK_MSG_MODULE_IMPORTED message which would be sent after the module was successfully loaded, and it would have the module structure object as the argument. At this moment the client can call yr_print_object_data or do whatever it wants with the module structure object.

This simplify the design even more. Any opinions?

msuvajac:
I agree that would be even better solution. I've made these changes just now.

wxsBSD:
FYI, I'm working on making these available via Python.

wxsBSD:
FYI, my intention is to have a PR up in a week or so that exposes this via Python.

plusvic:
Ok, perfect!

PR #365, opened 2015-11-01
Fix certificate parsing.

wxsBSD:
Looks like 5d6d8b1e9f3b4fc2b51857366048a7e9f8bd3708 is breaking parsing
of certificates. win_cert->Length is the length of the entire win_cert
structure. By using win_cert->Certificate as the pointer and
win_cert->Length as the size, the check will always fail because the
length needs to be decremented by the number of bytes you skipped over
when starting from win_cert->Certificate.

Fix this by decrementing length for the number of bytes we just skipped
over.

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -1134,7 +1134,7 @@ void pe_parse_certificates(
   //
 
   while (struct_fits_in_pe(pe, win_cert, WIN_CERTIFICATE) &&
-         fits_in_pe(pe, win_cert->Certificate, win_cert->Length) &&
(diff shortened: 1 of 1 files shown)


plusvic:
Take a look at this commit 2fac34b744813465ed39c19452289401f5430e69

It should solve this issue and improves the overall legibility of the condition.

wxsBSD:
I'll take a look after the kids go to bed tonight. :)

wxsBSD:
This fixes it, and I agree that it improves the legibility of the condition. Thanks for the quick fix!

PR #366, opened 2015-11-03
Properly handle negative serial numbers.

wxsBSD:
Negative serial numbers are stored in an ASN1_INTEGER using two's
complement. This meant that negative serials were not properly handled
prior to this commit.

To fix this always use i2c_ASN1_INTEGER() to convert a serial number to
it's "character" representation. You have to call i2c_ASN1_INTEGER()
twice. The first call needs to be with a NULL second argument, which
will return the number of bytes needed to store the serial, you can then
allocate a buffer and call i2c_ASN1_INTEGER() a second time to get the
properly formatted serial number.

In the interest of making the code easier to read I chose to do this
even for positive ASN1_INTEGER structures.

Negative serial numbers are against the RFC but they do exist. For
example 4bfe05f182aa273e113db6ed7dae4bb8.

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -1182,6 +1182,7 @@ void pe_parse_certificates(
     {
       const char* sig_alg;
       char buffer[256];
+      int bytes;
 
       ASN1_INTEGER* serial;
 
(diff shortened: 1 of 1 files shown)


wxsBSD:
A good explanation of this is available here: https://github.com/openssl/openssl/blob/984d6c6052169bcae8010de33f7796e455536d61/crypto/asn1/a_int.c#L93

PR #374, opened 2015-11-20
added rich_version function (+ documentation)

mrexodia:
Sample rules file with this function (this can be used for MSVC version detection as the tool versions used to compile the file are included in the rich header):

```
import "pe"

rule MSVC2013
{
meta:
    author = "mrexodia"

condition:
    pe.rich_version(21005) and pe.rich_version(40629)
}
```

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -1817,7 +1817,6 @@ define_function(language)
 define_function(rich_version)
 {
     YR_OBJECT* module = module();
-    PE* pe = (PE*)module->data;
     uint64_t version = integer_argument(1);
     size_t rich_len;
     PRICH_SIGNATURE clear_rich_signature;
@@ -1826,10 +1825,6 @@ define_function(rich_version)
 
     // Check if the required fields are set
     if (is_undefined(module, "rich_signature.length"))
-        return_integer(UNDEFINED);
-
-    // If not a PE file, return UNDEFINED
-    if (pe == NULL)
         return_integer(UNDEFINED);
 
     rich_len = get_integer(module, "rich_signature.length");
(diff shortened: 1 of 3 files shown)


libyara/modules/pe.c
```diff
@@ -1961,6 +1999,7 @@ begin_declarations;
   declare_function("imports", "s", "i", imports_dll);
   declare_function("locale", "i", "i", locale);
   declare_function("language", "i", "i", language);
+  declare_function("rich_version", "i", "i", rich_version);
```
plusvic: I would put this function inside the "rich_signature" structure and rename it to just "version". The rule would looks like: pe.rich_signature.version(21005).

And what about checking for tool IDs, or IDs/version pairs? Does it makes sense for you?

libyara/modules/pe.c
```diff
@@ -1814,6 +1814,44 @@ define_function(language)
   return_integer(0);
 }
 
+define_function(rich_version)
+{
+    YR_OBJECT* module = module();
+    PE* pe = (PE*)module->data;
```
plusvic: You don't use this pointer in the function. Can be removed.

libyara/modules/pe.c
```diff
@@ -1814,6 +1814,44 @@ define_function(language)
...
+    PRICH_SIGNATURE clear_rich_signature;
+    SIZED_STRING* rich_string;
+    int i;
+
+    // Check if the required fields are set
+    if (is_undefined(module, "rich_signature.length"))
+        return_integer(UNDEFINED);
+
+    // If not a PE file, return UNDEFINED
+    if (pe == NULL)
```
plusvic: Can be removed too.

libyara/modules/pe.c
mrexodia: I implemented your suggestions and I also added a `toolid` function. Checking ID/version pairs can also be done, but how would that function look? `rich_signature.toolversion` maybe? I thought about optional parameters too. Maybe something like `rich_signature.version(version, toolid [optional)` and `rich_signature.toolid(toolid, version [optional]`?

plusvic: Yes, a pair of functions like this:

```
rich_signature.version(version, toolid [optional])
rich_signature.toolid(toolid, version [optional])
```

This way you can check for any combination of toolid/version.

This would requiere actually four functions, two overloaded "version" functions and two overloaded "toolid" functions. I suggest writing a single internal function receiving both a toolid and a version and accepting a special value as a wildcard for either toolid or version. The other 4 functions could invoke the internal function.

mrexodia: Alright, I will try to do it later this evening.

mrexodia:
this should do it, a rules file like this works perfectly!

```
import "pe"

rule MSVC2013
{
meta:
    author = "mrexodia"

condition:
    pe.rich_signature.version(21005) and
    pe.rich_signature.version(30319, 158) and
    pe.rich_signature.toolid(225) and
    pe.rich_signature.toolid(225, 21005)
}
```

PS please also take a look at my pull request #376 I fixed building on Windows with VS2010.

wxsBSD:
Can we add constants here to make things nicer? I am guessing 21005 means something specific about a version of VS?

mrexodia:
@Wesley Shields Not really, it is just part of a version number in
devenv.exe or cl.exe (and various other tools). It is very undocumented so
I'd advise against constants (rule authors should do that).

PR #376, opened 2015-11-20
fixed building on windows + ignore windows build files

mrexodia:
alternative for pull request #375 that doesn't break multi-threading on windows, doesn't require upgrading to VS2015

[Review on Reviewable](https://reviewable.io/reviews/plusvic/yara/376)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,12 @@ stamp-h1
 # Linux and Mac files
 *.swp
 .DS_Store
+
+# Windows build files
+Debug/
+Release/
+*.vcxproj.user
+*.sdf
+*.suo
+*.opensdf
+*.obj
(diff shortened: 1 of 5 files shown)


mrexodia:
@keithjjones that's why I didn't update the project/solution files (I compile with VS2010 which are supported by the .lib files). This just fixes the code issues for compiling with Visual Studio in general.

Binaries on Github is always controversial but if you have the right visual studio version it is perfect :+1:

mrexodia:
Oh, I saw you asked my opinion on project files, I believe yara should migrate to CMake (mainly because it is less maintaining with all the different platforms involved) but that decision is up to @plusvic and comes with other problems

kallanreed:
The VS2010 tool chain is becoming difficult to find; I prefer @keithjjones #377 instead of this PR because of this.

PR #377, opened 2015-11-22
Visual Studio 2015 Update

keithjjones:
Added the threading issue in PR #376.  Added libraries required to compile with VS2015.  The project files and solution file are ready for anyone that wants to build Yara in VS2015.


kallanreed:
+1 on this PR. The VS2010 tool chain is becoming difficult to find not to mention that the 2010 tools are 9 years old.

mrexodia:
I disagree with this one mainly because of the bloat but I cheer for a vs13 (v120_xp) one! :+1:

kallanreed:
#407 included fixes for the merge conflicts.

@mrexodia i hear you on the bloat, but tool chain selection should be first based on the quality and safety of the output, second on availability of the tools and then other considerations like bloat.

The *ix people have it easy. On Windows the compiler improvements are unfortunately tied to VS and free tool availability is usually limited to Latest - 1. 2013 would work for me, but 2015 CE is really going to be the tool going forward for the next few years.

All that said, I mainly care that the project is easily buildable on Windows to encourage contribution. The pthread issue alone was almost enough to turn me away (thanks for the impl).

mrexodia:
Adding and removing binaries in commits is the kind of bloat that I would
never accept in open source projects I manage. I understand you need a
working tool chain but as a manager I wouldn't accept this pull request :)

When I'm back home I'll manually merge all the things from this (and other)
pull requests related to the Windows tool chain and provide a new pull
request that I would merge myself.

kallanreed:
@keithjjones why did the libs need to be updated? Seems strange that they wouldn't still link with the new tools.

kallanreed:
@mrexodia Mainly because I'm curious, how would you prevent the bloat? Submodule a second repo with only the binaries?

keithjjones:
@mrexodia are you suggesting my PR has unnecessary bloat in it?  Or that VS2015 has unnecessary bloat?  If you are referring to the bloat regarding the binary libraries in the Yara source code, that was something I pointed out on your pull request, and you agreed, and I suggested that they be removed.  You suggested CMake, and I think that is a good idea.  I only included these libraries to be consistent with the prior code base.  I wouldn't put binary libraries like this in the repositories I manage as well, but I do not manage this repository so I believe I only pushed consistent changes that mirrored the structure of this repository.

mrexodia:
https://github.com/keithjjones/yara/commit/26ebc379fe235f12daed7d5b114a5563af0901c5

And

https://github.com/keithjjones/yara/commit/e2ad6299e8c1e038ca3d293c387932d6977d4b4f

keithjjones:
@mrexodia Ok, you want to get rid of commits you asked me to do to get rid of files?  I don't understand.

If there is a way to merge them into a prior commit so the files never show up, please tell me how and I'd be happy to do it.  I do not know of a way to do it.

mrexodia:
You can delete the commits but it's not pretty :) that's why I suggest
manually merging these commits and do a new pull request.

keithjjones:
@mrexodia What do you mean by "manually"?   Is there a way to do that with Git, or do you mean really manually using a text editor?

kallanreed:
@keithjjones How would CMake help the problem of checking in the binaries? Wouldn't the libs still be needed to reference on Windows?

Also, I'm not entirely sure I understand the problem with the existing libs and VS2015. Is it because the RTL changed between VS versions so the exports are different/missing and linking fails? Where did the new libs come from?

keithjjones:
@kallanreed I don't know that CMake would take care of the problem, but it would be easier to reference the external libraries.  

The existing libraries were compiled with VS2010, and they don't link with VS2015.

kallanreed:
I see. Did you compile the new ones you included? I'm just wondering if it's possible to sub module the source and compile it with the build. Maybe not if those projects aren't themselves VS friendly.

kallanreed:
@keithjjones also, if you want I can merge master again and pull in the changes by hand and submit a new PR.

If you want to do it, take a look at #407 because there were some additional changes. Pick an .sln to use because you added a new one and I had modified the existing one.

keithjjones:
@kai5263499 If you want to do that, that would be great.  If not, I can look at it later this week.  Thank you sir!

kallanreed:
@keithjjones @mrexodia
So I'm clear, the bloat that's the problem with this PR is because of the deleted lib files, not because of the newly checked in lib files.

I'm going to pull master again and then manually repatch the code changes and include the new lib files.

keithjjones:
@kallanreed Sounds right to me.

mrexodia:
I don't know anymore lol... I think visual studio 2010 is a great tool
chain and it's easy enough to convert the solution and projects to a new
visual studio version without having to change anything in the current
repository (apart from the few gitignore and compile error changes I did).

As for the static libraries, just slightly modify the project files to
include those from an external directory no need to delete or replace any
of the current ones.

My suggestion would be if you want to compile this project that you include
this repo in a submodule and have your own solution and static libraries in
the parent repo.

keithjjones:
The same could be said for legacy VS2010 files.  I think we all agree that putting dependency binary lib files in a repository isn't the best solution, but we can to be consistent with the original repository.  Therefore, that is why I included them... so other VS2015 people didn't have to compile them or wonder why they wouldn't link.

I just simply want to load VS2015 and be able to compile it without a lot of fuss.  If we want it to be compatible and support older VS versions, that could be a different branch  or something.

kallanreed:
VS2010 is great. VS2015 is great. The problem is that getting the 2010 tools is becoming a pain.

Thoughts on pulling out all of the VS stuff into submodules? Would having a submodule for VS2010 and another for VS2015 be terrible? In this project I'd add a readme to fetch the appropriate submodule?

mrexodia:
I think manually changing paths to a few lib files isn't a great hassle.
You just need the compiled files for your favorite visual studio version. I
wasn't suggesting adding submodules in this repository. I use Yara inside
the x64dbg repository and to compile I would have my own vs projects.

kallanreed:
Off this topic but related to the change to pthreads, I'm getting an invalid handle exception in the debugger for this call
```
int pthread_mutex_destroy(pthread_mutex_t *mutex)
 {
 	return CloseHandle(mutex) == TRUE ? 0 : -1;
 }
```

Have you seen this @mrexodia ?

mrexodia:
To be honest I didn't really test much. You can use GetHandleInformation to
check if the handle is valid before closing it, that should make the
exceptions go away. Anyways just ignore those exceptions, they only happen
under a debugger anyway.

There might be an invalid call to mutex_destroy see the callstack. Another
problem might be that init doesn't return a valid mutex handle somehow.

kallanreed:
The handle looks like a valid handle (pointer with a small value). I'll dig in a little more.

mrexodia:
Possibly close is called multiple times?

keithjjones:
Any way is fine with me as long as I don't have to do a big migration to VS2015 every time I pull from master.  There were more changes than simply changing library paths.  VS2010 won't be compatible with new projects forever, and it will be used less and less, so to me it makes more sense to move the project along as technology changes.  If legacy support is desired, and I'm not knocking the need for it, you could make that support a different part of the project on a case by case basis as needed by the users of that component or project.  This is where CMake would make more sense, because you could specify the version of solution to be created by CMake.  The only problem lies in the required binaries, which to me, should be compiled and linked separately by the developer rather than depending on what is in the repository, but I don't make the decisions here.  You could also just make different solution files for each version in different subdirectories with the required binary libraries.  That is how I have seen other projects deal with the issue.

Submodules, Branches, Forks, Subdirectories, whatever.  It all could work.  We just need to pick one and implement it.

kallanreed:
I'm going to try this
1. Re-patch all of the windows changes so they are clean (minus VS changes)
2. Leave the existing projects in place for VS2010
3. Add a new repo to hold the VS2015 solution and dependencies
4. Submodule the VS2015 repo under windows and add a readme there explaining

Why add the submodule to yara instead of the other way around? Because I expect yara to update all the time and the actual project file to update infrequently. It's a giant pain to maintain a reference to an active submodule.

My goal is to have this compile on Windows without a bunch of messing around trying to find dependencies and build them. I really don't like working on projects that a bunch of hidden setup work. At the very least, a "building for windows" readme would go a long way toward making this dev friendly.

mrexodia:
Fixed the pmutex code:

```
#ifndef HAVE_PTHREAD
#ifdef _WIN32

typedef HANDLE pthread_mutex_t;

unsigned long pthread_self()
{
    return GetCurrentThreadId();
}

int pthread_mutex_init(pthread_mutex_t *mutex, void* attr)
{
    *mutex = CreateSemaphore(NULL, 1, 1, NULL);
    return *mutex == NULL;
}

int pthread_mutex_destroy(pthread_mutex_t *mutex)
{
    BOOL result;

    result = CloseHandle(*mutex);
    *mutex = NULL;
    return result == TRUE ? 0 : -1;
}

int pthread_mutex_lock(pthread_mutex_t mutex)
{
    return WaitForSingleObject(mutex, INFINITE) == WAIT_OBJECT_0 ? 0 : -1;
}

int pthread_mutex_unlock(pthread_mutex_t mutex)
{
    return ReleaseSemaphore(mutex, 1, NULL) == TRUE ? 0 : -1;
}
#endif
#endif
```

keithjjones:
@kallanreed Sounds like a great plan.  If I have time later this week to lend a hand, let me know if you need help.

mrexodia:
I created a new pull request that can be merged regardless of what will happen with the visual studio projects #408

keithjjones:
Thakns

kallanreed:
One downside of having a second project is that it will fall out of sync with the other one. This could be avoided with wildcard file includes so the projects don't need to be touched when new files are added. The makefiles do explicitly list the sources out which makes me wonder if that's a good idea or not.

PR #387, opened 2016-01-03
Document modules_callback for yara-python.

wxsBSD:


wxsBSD:
This should be merged only if https://github.com/plusvic/yara-python/pull/4 is merged. I put this in a second PR because it shouldn't hold up the other one I just submitted for documenting CALLBACK_MSG_MODULE_IMPORTED.

Issue #388, opened 2016-01-03
Nested arrays and dictionaries cause segfault


Northern-Lights:
In the `Writing your own modules` documentation, it is stated that arrays may be nested.
https://yara.readthedocs.org/en/v3.4.0/writingmodules.html

An example given in the documentation is as follows:
```
begin_declarations;

    declare_integer_array("foo");

    begin_struct_array("bar")
        declare_string("baz");
        declare_integer_array("qux");
    end_struct_array("bar");

end_declarations;
```

Examples for accessing these data structures were given as follows:
```
set_integer(<value>, module, "foo[0]");              //Works
set_integer(<value>, module, "foo[%i]", 2);        //Works
set_string(<value>, module, "bar[%i].baz", 5);   //segfaults
set_string(<value>, module, "bar[0].qux[0]");     //set_integer() should be used, but it also segfaults, regardless
set_string(<value>, module, "bar[0].qux[%i]", 0);
set_string(<value>, module, "bar[%i].qux[%i]", 100, 200);
```

It appears that any use of the `bar[]` array with `set_integer()` or `set_string()`, even with a hard-coded index, and even when not accessing the `qux[]` nested array, causes a segfault.

The same segfaulting issue occurs with nested dictionaries, as well.

Northern-Lights:
This should fix it: https://github.com/plusvic/yara/pull/392

Northern-Lights:
There was also an issue where if a `dict` had a nested `dict`, then looking up a value (for an `integer` or something within the struct) that had not been set would cause a segfault because `r2.o` would be UNDEFINED in exec.c.  For example, the following was causing a segfault:
```
import "demo"

rule ddddemo {
condition:
	demo.d1["abc"].i1 == 5 and
	demo.d2["def"].i2 == 5 and
	demo.d2["def"].d2_2["ghi"].i3 == 5 and
	demo.d2["xyz"].d2_2["abc"].i3 > 1
}
```
Where demo.c had these lines:
```
set_integer(5, module_object, "d1[\"abc\"].i1");
set_integer(5, module_object, "d2[\"def\"].i2");
set_integer(5, module_object, "d2[\"def\"].d2_2[\"ghi\"].i3");
```

The yara CLI util would segfault when checking the type of `r2.o` because was UNDEFINED.

My solution: also check for `ensure_defined(r2);` in exec.c in the `OP_LOOKUP_DICT` case.

plusvic:
[...]

PR #392, opened 2016-01-08
Fixes issue #388 (nested dictionaries causing segfault)

Northern-Lights:
`yr_object_copy` was missing a case in the switch which should have initialized the `prototype_item` of the dictionary object.  This was causing a segfault for nested dictionaries when it came time for them to be copied.

I also added an `assert(FALSE)` as the default case, because if we don't know what we're copying, we should fail.

### DIFF
--- a/libyara/object.c
+++ b/libyara/object.c
@@ -628,6 +628,21 @@ int yr_object_copy(
       }
 
       break;
+
+    case OBJECT_TYPE_DICTIONARY:
+      ;
+      YR_OBJECT_DICTIONARY *dict = (YR_OBJECT_DICTIONARY *) object;
+      YR_OBJECT *prototype_item;
+
+      yr_object_copy(dict->prototype_item, &prototype_item);
+
+      ((YR_OBJECT_DICTIONARY *)copy)->prototype_item = prototype_item;
+
+      break;
+
+    default:
+      assert(FALSE);
+
   }
 
   *object_copy = copy;
--- a/libyara/object.c
+++ b/libyara/object.c
@@ -545,7 +545,6 @@ int yr_object_copy(
   YR_OBJECT* copy;
   YR_OBJECT* o;
 
-  YR_ARRAY_ITEMS* array_items;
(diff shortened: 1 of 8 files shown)


libyara/object.c
```diff
@@ -609,6 +609,28 @@ int yr_object_copy(
       break;
 
     case OBJECT_TYPE_ARRAY:
+      ;
+      YR_OBJECT_ARRAY *array = (YR_OBJECT_ARRAY *) object;
```
plusvic: Declaring variable inside switch statements doesn't work in C89. You can use the "o" variable.

libyara/object.c
```diff
@@ -609,6 +609,28 @@ int yr_object_copy(
...
+      YR_OBJECT *array_prototype_item;
+
+      yr_object_copy(array->prototype_item, &array_prototype_item);
+
+      ((YR_OBJECT_ARRAY *)copy)->prototype_item = array_prototype_item;
+
+      if (array->items == NULL)
```
plusvic: You don't need to initialize array->items here. This should be enough:

```
      yr_object_copy(((YR_OBJECT_ARRAY*) object)->prototype_item, &o);
      ((YR_OBJECT_ARRAY*) copy)->prototype_item = o
```

libyara/object.c
```diff
@@ -628,6 +650,21 @@ int yr_object_copy(
       }
 
       break;
+
+    case OBJECT_TYPE_DICTIONARY:
+      ;
+      YR_OBJECT_DICTIONARY *dict = (YR_OBJECT_DICTIONARY *) object;
```
plusvic: Same thing here, invalid variable declaration in C89. This should work:

```
      yr_object_copy(((YR_OBJECT_DICTIONARY*) object)->prototype_item, &o);
      ((YR_OBJECT_DICTIONARY*) copy)->prototype_item = o;
```

plusvic:
Very good job!

Northern-Lights:
Changes implemented.  Thanks for the suggestions!

PR #396, opened 2016-01-22
Fix match issue.

wxsBSD:
When the length of remaining bytes is less than the length of the string we are
searching for, don't do the search as it can't possibly match. This was causing
false positives when using a rule that looks like this:

pe.exports("Driver")

Anything that starts with "D" will be incorrectly matched when run against
b71c531d50d2b634e6000fcdabc9bbb4, because the file is truncated in the export
table and the entry happens to start with D.


wxsBSD:
Good point. It is a nice solution!

PR #398, opened 2016-01-27
DllCharacteristics and subsystem PE flags.

campt:
This adds in some missing PE SUBSYSTEM types and adds the DllCharacteristics and flag values.  We are using YARA internally and would like to keep the upstream PE module in sync with our usage.

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -1386,6 +1386,10 @@ void pe_parse_header(
       OptionalHeader(Subsystem),
       pe->object, "subsystem");
 
+  set_integer(
+      OptionalHeader(DllCharacteristics),
+      pe->object, "dllcharacteristics");
+
   pe_iterate_resources(
       pe,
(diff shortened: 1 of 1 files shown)


libyara/modules/pe.c
```diff
@@ -1386,6 +1386,10 @@ void pe_parse_header(
       OptionalHeader(Subsystem),
       pe->object, "subsystem");
 
+  set_integer(
+      OptionalHeader(DllCharacteristics),
+      pe->object, "dllcharacteristics");
```
plusvic: To be coherent with the naming schema of the module I would add an underscore: dll_characteristics

plusvic:
Thank you for contribution @campt. Would you mind adding this new feature to the documentation before I merge it into the master?

campt:
Sure.  I’ll post another pull request sometime next week.

 

(Tracy) Camp

Bit9+CarbonBlack

PR #401, opened 2016-02-10
Improve detection of entry point offset

zavrac:
This commit improves detection of entry point offset in ELF files. If
detection of entry point offset based on section adresses fails,
algorithm iterates over segments and tries to compute offset of entry
point based on segment adresses.

This commit also enables detection of entry point offset in PE files for
other architectures than x86 and x86_64, because my project is focused
also on ARM PE files.

### DIFF
--- a/libyara/exefiles.c
+++ b/libyara/exefiles.c
@@ -61,8 +61,6 @@ PIMAGE_NT_HEADERS32 yr_get_pe_header(
   headers_size += pe_header->FileHeader.SizeOfOptionalHeader;
 
   if (pe_header->Signature == IMAGE_NT_SIGNATURE &&
-      (pe_header->FileHeader.Machine == IMAGE_FILE_MACHINE_I386 ||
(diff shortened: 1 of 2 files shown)


plusvic:
The use of "entrypoint" keyword in YARA rule is deprecated. This should be implemented in the ELF module (modules/elf.c)

zavrac:
I improve detection of entry point also in ELF module.

hillu:
Can you provide an ELF binary with which the fix can be demonstrated?

zavrac:
I have some ELF binaries that contains only segments, not sections. These binaries are available at https://www.dropbox.com/sh/dinavl46emdndko/AACGxMH5blSMDRgW1foiF2nUa?dl=0

plusvic:
@jacob-baines already did some work to solve this issue in #606. The entry point for the provided files are calculated correctly with the latest code.

PR #407, opened 2016-02-21
Build support for VS2015

kallanreed:
VS2015 changes pulled from PR #377 with some additional fixes for changes in master since the previous pull was requested.


keithjjones:
+1

VS2015 community is free, and includes a lot more functionality in the free edition than prior versions.  It makes sense to upgrade.

kallanreed:
This contains a bunch of deleted files so we're going to make a new PR that's clean.

PR #409, opened 2016-02-21
Split text and hex strings

hillu:
When writing rules that are based upon disassembler output, I would like to be able to sprinkle 
bits of text produced by the used tool (e.g. IDA Pro) as comments between the hex strings.

The proposed change is inspired by the way C understands `"abcd" "ef"` just like `"abcdef"`.


hillu:
So it turns out that the feature as implemented causes the existing test suite to fail. Good thing I did the porting. ;-)

PR #408, opened 2016-02-21
visual studio building

mrexodia:
related to #377 and #407 and #376 (but no changes to the current visual studio project files)

[Review on Reviewable](https://reviewable.io/reviews/plusvic/yara/408)

PR #410, opened 2016-02-22
Add support for VS2015 as a submodule

kallanreed:
After a long conversation about the best way to add VS2015 support in #377, we came to the conclusion that updating binaries in repo wasn't a good long term solution but neither was having to sort out dependencies just to get the project to build.

This is the compromise. New VS version support added via a submodule should keep the core repo light while still providing a decent pull and build story on Windows.

I added a build readme for Windows with instructions on how to pull the submodule. The submodule itself contains the project and binaries needed to compile. I also cleaned up the project files in the new repo. This change also includes PR #408.

The submodule is currently pointing to a repo I manage, but that can be moved under `plusvic` if that's desirable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,12 @@ stamp-h1
 # Linux and Mac files
 *.swp
 .DS_Store
+
+# Visual Studio files
+Release/
+Debug/
+x64/
+*.obj
+*.suo
+*.sdf
+*.opensdf
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "windows/yara-VS2015"]
+	path = windows/yara-VS2015
+	url = https://github.com/kallanreed/yara-VS2015.git
--- a/libyara/include/yara/strutils.h
+++ b/libyara/include/yara/strutils.h
@@ -25,7 +25,9 @@ limitations under the License.
 
 // Cygwin already has these functions.
 #if defined(_WIN32) && !defined(__CYGWIN__)
+#if defined(_MSC_VER) && _MSC_VER < 1900
 #define snprintf _snprintf
+#endif
 #define strcasecmp _stricmp
 #define strncasecmp _strnicmp
 #endif
--- a/libyara/libyara.c
+++ b/libyara/libyara.c
@@ -47,6 +47,43 @@ static int init_count = 0;
 char lowercase[256];
 char altercase[256];
 
+#ifndef HAVE_PTHREAD
+#ifdef _WIN32
+
+typedef HANDLE pthread_mutex_t;
+
+unsigned long pthread_self()
+{
+    return GetCurrentThreadId();
+}
+
+int pthread_mutex_init(pthread_mutex_t *mutex, void* attr)
+{
+    *mutex = CreateSemaphore(NULL, 1, 1, NULL);
+    return *mutex == NULL;
+}
+
+int pthread_mutex_destroy(pthread_mutex_t *mutex)
+{
+    BOOL result;
+
+    result = CloseHandle(*mutex);
+    *mutex = NULL;
+    return result == TRUE ? 0 : -1;
+}
+
+int pthread_mutex_lock(pthread_mutex_t mutex)
+{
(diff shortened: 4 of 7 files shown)


kallanreed:
cc: @keithjjones this is done if you want to give it a try.

keithjjones:
:+1: I'll give it a try this week.  Nice work!

plusvic:
To be honest, I agree that having large binary files in the repo is not a good idea, but I don't like the idea of submodules neither. 

After googling a little I found Git LFS (https://git-lfs.github.com/). I haven't tried it yet, but it seems to solve the problem of storing binary files without bloating the repository and without adding more complexity with submodules. So I'm going to give it a try and see how it works. If anyone want to test it and share his experiences, they're welcomed.

keithjjones:
@plusvic I played with LFS some.  It worked as advertised for me.  You shouldn't have too much trouble with it.  Have you thought about how you would like to handle supporting multiple versions of VS?

Thanks for a great tool!

plusvic:
Assuming that we use LFS for storing the binary files, I would simply create separate directories under yara/windows for VS2010 and VS2015 with their own projects and binaries.

keithjjones:
@plusvic Sounds great.

kallanreed:
Are you going to make the change or do you want me to make a new PR with LFS?

My only request is that the the VS2015 solution and project files in this PR end up in the repo. I hand edited them to make sure that all build configurations and targets are set up properly. The VS2010 projects are kind of a mess with different settings for every combination of x86/x64/Release/Debug.

plusvic:
Yes sure, go ahead.

kallanreed:
@plusvic  I started to give the a try and ran into a problem. It doesn't look like LFS can be used on "public forks".

```
$ git push -u origin lfs
Git LFS: (0 of 4 files) 0 B / 26.82 MB
Git LFS objects can not be uploaded to this public fork.
Docs: https://github.com/contact
error: failed to push some refs to 'https://github.com/kallanreed/yara.git'
```

[This thread](https://github.com/github/git-lfs/issues/773) seems to indicate that this expected behavior - at least for now. Effectively, forking is broken.

Also the LFS [billing page](https://help.github.com/articles/billing-plans-for-git-large-file-storage/) seems to indicate there are storage/bandwidth limits after which things break in unfortunate ways.

LFS might not be the best solution to this problem given these limitations.

kallanreed:
What about including the dependencies as NuGet packages. That way they don't have to be included in the repo and they can be updated (for fixes) without needing to push new bits into the repo.

I could look into building either binary libs or packaging the source so it can be built with yara.

plusvic:
That's bad news :(

I haven't used NuGet packages, can't tell if that's the right solution, but according to what I've read we need to have an account nuget.org and install a nuget client, which don't integrate with git. A bit cumbersome. This is why open source development in Windows is a pain. 

To be honest the Windows project in this repository is here for my own convenience while building the Windows binaries. I never intended to provide any support for them, and given the issues caused by the Windows toolchain I think I'm going to stick to the DYI philosophy for the time being. For users who don't need to build the code themselves we already provide binaries.

kallanreed:
FWIW you don't need a nuget account to pull packages and VS has nuget built in. It'll pull packages from nuget during the build.

I get your point. Are you willing to at least take fixes to make the code compile in newer versions of VS (e.g. #408) if not the project will have to be permenently forked which isn't great.

Also, how about a build read me in the Windows folder that talks about what's needed to get up and running? Maybe with a link to the VS project repo if people want it.

It's low overhead for you but it helps the Windows devs who are already have a tool chain handicap.

kallanreed:
@plusvic before you give up on this altogether, let me make a PR with the NuGet idea so you can see what it looks like. It would be really nice to have this building on Windows for other contributors. I was kind of stuck until I found PRs from @keithjjones and @mrexodia unblocking Windows building.

mrexodia:
@plusvic please be so kind to merge my pr that addresses compiler errors on Windows. I think once you have those out of the way it becomes a no brainer to slam some libraries in there by yourself. Implementing pthread_mutex all the time is kind of annoying...

kallanreed:
@plusvic I made PR #414 that uses NuGet to hold the dependencies which keeps the repo clean and provides a good build experience.

Issue #406, opened 2016-02-19
Yara doesn't allow negative integers in meta fields


Professor-plum:
This may be intended behavior, but if so please document it.
 meta fields cannot be negative values.

The attached rule will not compile
[test.txt](https://github.com/plusvic/yara/files/138824/test.txt)

hillu:
Are you seeing something like `syntax error, unexpected '-', expecting _NUMBER_ or _TEXT_STRING_ or _TRUE_ or _FALSE_`? This looks like an oversight in the lexer to me -- the  underlying datatype in `_YR_META` is an `int64_t`.

Cheers,
-Hilko

Professor-plum:
Yes, that's exactly the error I'm seeing, I also believe it is an oversight. Thanks for looking up the signedness of the underlying type.

PR #411, opened 2016-02-22
Accept signed numbers (integers, doubles)

hillu:
From what I can see, `_NUMBER_` tokens are used in meta declarations, (grammar.y:428) and primary expressions (grammar.y:1567). In both cases, an int64_t is used to store the value, so the sign can be preserved.

`_DOUBLE_` tokens are used in primary expressions (grammar.y:1577), the same applies here.

After the changes in this PR, the following works as expected:
<pre>
$ cat t.yar
rule t1 { meta: myint = +7 condition: true }
rule t2 { meta: myint =  7 condition: true }
rule t3 { meta: myint = -7 condition: true }
rule t4 { condition: -1.0 > -2.0 }
$ ./yara -m t.yar /dev/null 
debug enabledt1 [myint=7] /dev/null
t2 [myint=7] /dev/null
t3 [myint=-7] /dev/null
t4 [] /dev/null
</pre>

### DIFF
--- a/libyara/grammar.y
+++ b/libyara/grammar.y
@@ -438,6 +438,19 @@ meta_declaration
 
         ERROR_IF($$ == NULL);
       }
+    | _IDENTIFIER_ '=' '-' _NUMBER_
+      {
+        $$ = yr_parser_reduce_meta_declaration(
+            yyscanner,
+            META_TYPE_INTEGER,
+            $1,
+            NULL,
+            -$4);
+
+        yr_free($1);
+
+        ERROR_IF($$ == NULL);
+      }
     | _IDENTIFIER_ '=' _TRUE_
       {
         $$ = yr_parser_reduce_meta_declaration(


plusvic:
Plus and minus signs in front of numeric constants can't be handled in the lexer because those signs can be also arithmetic operations. Let's say we have the expression: `1-1==0`. You expect that expression to be syntactically valid and true, right? But if you include the minus sign as part of numeric constant this expression will be tokenized as: `1`, `-1`, ` ==`, `0` and the grammar will reject such an token sequence.

In fact I made the same mistake before, take a look at this commit when I solving the issue: https://github.com/plusvic/yara/commit/6cd9fad91545e19f3ecdda2f5736971bb715000e

hillu:
Thanks; Let's see if I can come up with a correct solution then. :-)

hillu:
Meh, for some reason I thought that the problem described in #406 was not limited to meta declarations. So, handling the unary minus in the grammar was actually not that difficult. (Although the near-duplication of code may not be the most elegant solution.)

PR #412, opened 2016-02-23
libyara/exec: avoid unaliged access to 64bit variable

sebastianas:
The derefence of an unaligned 64bit variable results in a SIGBUS abort
on 32bit SPARC. ARMv5 CPUs seem to perform the load but load garbish.
This memcpy() workaround forces the compiler to do something that works
on even if the data was not properly aligned. For X86 it means no
change. ARM on other hand will produce slightly different code depending
on the CPU used.

While at it, I coverted all (ip + 1) references.

Signed-off-by: Sebastian Andrzej Siewior <sebastian@breakpoint.cc>


plusvic:
Did you test YARA extensively on SPARC and ARM? I'm wondering if those are the only places where we have unaligned accesses.

sebastianas:
clamav in 0.99 is using yara however they pulled a modified version of it into their source. Before the patch clamav's testsuite failed on sparc:
https://buildd.debian.org/status/fetch.php?pkg=clamav&arch=sparc&ver=0.99%2Bdfsg-0%2Bdeb7u1&stamp=1455813120
After the patch, it passed:
https://buildd.debian.org/status/fetch.php?pkg=clamav&arch=sparc&ver=0.99%2Bdfsg-0%2Bdeb7u2&stamp=1456266227
Debian's ARM buildd are ARM v7 or better so they don't the problem with loading garbage which means I can't tell if the testsuite would fail otherwise.

If you have a testsuite or something you want me to run on sparc to verify I would try to give it a try.

plusvic:
There's a test suit in Python, but of course it requires building yara-python too. You can find it here:

https://github.com/plusvic/yara-python/blob/master/tests.py

plusvic:
Now we have the same tests in C (https://github.com/plusvic/yara/blob/master/tests/test-rules.c) and you can run them with "make check".

Could you verify if the tests pass in SPARC and ARM?

sebastianas:
if fails on the sparc and the arm box. ARM says
 test-rules: arena.c:469: yr_arena_coalesce: Assertion `page != ((void *)0)' failed.
and sparc throws SIGBUS at the earlier *reloc_address.
Even after I fixed those with memcpy() I saw no change in ARM but sparc went segfault in exec.c
It completed fine with my addons on x86_64. 
That is my feedback so far.

plusvic:
A lot of work as been done in #1724 that should fix most of the unaligned memory accesses.

PR #413, opened 2016-02-27
Port most of the the rule-related tests from yara-python

hillu:
The test code for parsing and behavior of the rule engine belongs with Yara itself rather than the the  Python bindings that have been split off. Now the tests can be run again when the library is built.


plusvic:
God job!. I wanted to do this since long ago.

PR #414, opened 2016-02-28
Add Visual Studio 2015 Support with NuGet

kallanreed:
Adding build support for VS2015. The binary dependencies are contained in a NuGet package. Visual Studio will download the package automatically when yara is built. The NuGet package is [here](https://www.nuget.org/packages/yara-vs2015-binary-dependencies/) and the package project is [here](https://github.com/kallanreed/yara-vs2015-binary-dependencies).

This solution keeps tool chain specific binaries out of the yara repo, doesn't complicate the repo with submodules and still provides a good "clone and compile" experience for VS2015 contributors. The NuGet package version is locked to the specific version referenced in the checked in project file so so future updates to the NuGet package are "opt in".

This hopefully addresses concerns raised in #410 and includes changes from #408 (pthread impl).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,16 @@ stamp-h1
 # Linux and Mac files
 *.swp
 .DS_Store
+
+# Visual Studio files
+Release/
+Debug/
+x64/
+*.obj
+*.suo
+*.sdf
+*.opendb
+*.opensdf
+
+# NuGet
+windows/*/packages/
--- a/libyara/include/yara/strutils.h
+++ b/libyara/include/yara/strutils.h
@@ -25,7 +25,9 @@ limitations under the License.
 
 // Cygwin already has these functions.
 #if defined(_WIN32) && !defined(__CYGWIN__)
+#if defined(_MSC_VER) && _MSC_VER < 1900
 #define snprintf _snprintf
+#endif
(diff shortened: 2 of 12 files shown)


kallanreed:
cc: @keithjjones

keithjjones:
I thought Nuget was mainly for .NET?  How does it work in this case?
:+1:

kallanreed:
Nuget added support for native targets around VS2012. Basically it's just the include headers and the libs packaged up and a targets file that is injected into the projects so that the projects can reference the package files. I linked to the nuget package project above if you're interested in how it's built.

keithjjones:
That is pretty cool!  Thanks for teaching me something.

keithjjones:
Apparently this PR has conflicts already.  Something must have changed since last night.

keithjjones:
I was hoping this would make it into master at some point.

plusvic:
Don't worry, this is going to be merged in the master branch, it's just that I haven't found the time to do it. That's why I'm only merging simple patches that don't require too much effort.

jacob-baines:
For what its worth, I've tested/used this successfully and am also looking forward to seeing it merged in.

plusvic:
Manually merged.

Issue #33, opened 2013-11-24
Optimize memory usage when scanning processes and files


plusvic:
_From [zeroStei...@gmail.com](https://code.google.com/u/111030040457214474971/) on December 20, 2011 19:01:35_

This is a feature request not a bug.

It would be helpful if Yara would not load entire files into memory to scan them but step through them with an sizeable overlap to avoid false negatives. It would also be very helpful if Yara would use a similar chunking method to scan processes because files can at least be scanned in as chunks of data by the controlling process.

Basically Yara's memory usage needs to have some limitation.

_Original issue: http://code.google.com/p/yara-project/issues/detail?id=33_

kallanreed:
I want to try a fix for process memory reading. Basically adding a new `SectionReader` type to keep track of the process handle, the buffer and the last read location of the section. Instead of allocating enough chunks for the whole section, one chunk would be reused.

The caller would call `OpenReader` to get a new reader and call a `ReadNext` function with the `SectionReader` which would either read the next chunk of the section or would return null or a failure code if the section is no longer readable. A `HasMore` flag would indicate that reading is incomplete and could be used in the caller's while condition.

When the caller is done reading the section, it would have to call a `CloseReader` which would free the internal buffer and detach from the process.

The benefit would be much smaller memory footprint during process scanning at the expense of potential memory consistency.

I need this behavior because the current memory characteristics will not work for the environment we want to use YARA in.

jhumble:
YARA does seem to be using memory mapped files for OSX and Linux, but the Windows implementation using ReadProcessMemory, does result in a a huge increase in physical memory usage if I'm not mistaken. 

Good luck, Kallenreed. This would be hugely helpful for me as well. I had started coding a POC implementation similar to what you've described for Windows. 

So far, my solution was to no longer read data into each block in yr_process_get_memory and instead just stuff the process handle into the data field in the block structure. Then, when iterating over blocks in yr_rules_scan_mem_blocks I can call ReadProcessMemory to load the data for that one block, scan it, release it, and move onto the next block. 

Your implementation using a SectionReader is much cleaner though, let me know if there is any way I can help!

kallanreed:
Yeah, you don't really want to leak the platform specific code into the rules handler. I'll give my design a try this weekend and see if I can make it work for all three platforms. I don't have a Linux build machine handy so I'll need some help testing once I get a PR ready.

Sent from my iPhone

[...]

PR #418, opened 2016-03-06
Read process memory in VM sections instead of all at once

kallanreed:
Fix for issue #33.

The gist of the change is to delay allocation and reading of memory sections until they are actually scanned. This reduces the memory footprint of yara during scanning at the expense of possibly having to read a given section multiple times (mainly when modules are used). Allocated sections are freed after they are scanned.

The main new concept is the `YR_BLOCK_ITERATOR` abstraction which encapsulates the state needed to read process memory (e.g. the process handle) between calls. `yr_open_process_iterator` in proc.h is used to create an iterator that iterates all memory sections in a process. `yr_close_process_iterator` must be called to detach and free the iterator. A simple implementation of is included to also allow single blocks to be iterated (which is the case for file scanning).

Some simple memory results:

    Test                                Before    After
    Notepad on Windows                  96MB      21MB
    Test app on OS X                    76MB      34MB
    Vim on Ubuntu                       17MB      7MB

This change includes PR #414 which I'm also hoping will get pulled, however this change does not require #414 and I could pull out the dependency if #414 isn't going to be pulled.


kallanreed:
@Plusvic this change still had the bug with the bad reads when presenting matches that we never figured out. I thought I had closed this PR.

I never figured out how to address that problem with the fix without breaking the API for modules or the contract with callers of the library.

plusvic:
You're right. I forgot that. But now I've fixed this problem in 89a305f91de07ebfbeaf56a228140e10f27a8623

PR #419, opened 2016-03-08
Split strings, hex-strings

hillu:
This series of commits extends the grammar (and in the case of hex strings) the lexer to accept strings and hex strings that are be split into multiple tokens, possibly with interwoven comments.

For hex strings, this is particularly useful because it is now possible to mix bytes with assembly mnemonics or annotated hex dumps.

For example:

    $str = "foo" "bar" "baz"
    $str = "foo" /* a comment */ "bar" // another comment
           "baz"
    $hex = { 66 6f 6f } { 62 61 72 }
           { 62 61 7a }
    $hex = { 66 6f 6f } /* a comment */ { 62 61 72 } // another comment
           { 62 61 7a }

### DIFF
--- a/libyara/grammar.y
+++ b/libyara/grammar.y
@@ -78,6 +78,8 @@ limitations under the License.
       yyerror(yyscanner, compiler, NULL); \
       YYERROR; \
     }
+
+  int inside_rule_body = 0;
 %}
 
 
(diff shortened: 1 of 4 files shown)


hillu:
This PR seems to be mostly obsolete with #455 having been merged.

Issue #399, opened 2016-01-30
Entrypoint bug (unknown)


Chandlr:
sample file: 
https://mega.nz/#!hdMVGKqC!zafNo7KHdyi4B-yLIxmGWFytl-454oTGfdBJQ00Tymw
just to test the entrypoint:
rule UnKnown
{
	meta:
		author="_pusher_"
	strings:
		$a0 = { 68 00 00 42 00 31 C0 40 EB 58 }
	condition:
		$a0 at pe.entry_point
}

wxsBSD:
I looked at this a bit tonight and there is what amounts to an off-by-one error in ```pe_to_rva_offset()```. In this particular case the proper entry point is at offset 2 but YARA calculates it to be 3 because ```section->PointerToRawData``` (https://github.com/plusvic/yara/blob/master/libyara/modules/pe.c#L396) is 1 which means the calculation done at https://github.com/plusvic/yara/blob/master/libyara/modules/pe.c#L416 is incorrectly adding one (rva - section_rva is 2). I'm going to read the PE specification a bit closer and look for any corner cases which indicate we should ignore (or properly align?) PointerToRawData, but that will have to wait until tomorrow.

wxsBSD:
I've got this one solved. I'll have a PR up shortly, but I want to see if I can quickly solve the other entrypoint bugs you raised.

Issue #373, opened 2015-11-20
Entrypoint bug (file packed with UPack)


Chandlr:
sample:
https://mega.nz/#!9YU1EbSI!w3Dh1w2FXvs4mV5BCzvTnTc8tCiVPn7i3iQX_DG2YZo

rule Upack_entrypoint_issue
{
      meta:
		author="_pusher_"
strings:
		$right = { BE B0 11 40 00 }
		$wrong = { 0B 01 4C 6F 61 64 4C }

condition:
		$right at pe.entry_point or $wrong at pe.entry_point
}

correct Entrypoint is: 18 (file offset)

wxsBSD:
I've also got this one fixed (same root cause as #399).

Chandlr:
Sounds good :)

PR #420, opened 2016-03-09
Fix RVA calculation corner cases.

wxsBSD:
When the PointerToRawData is less than 0x200 it is rounded down to 0x00. This
causes entry point calculation to be incorrect in some cases. For example, these
two issues highlight the problem:

https://github.com/plusvic/yara/issues/399
https://github.com/plusvic/yara/issues/373

Details of this are in
https://code.google.com/archive/p/corkami/wikis/PE.wiki#PointerToRawData

I tested this against 293 PE files and found one difference in entry point
calculation, which upon inspection turned out to be another case of this bug.

Fixes: #399
Fixes: #373

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -395,6 +395,13 @@ int64_t pe_rva_to_offset(
         section_rva = section->VirtualAddress;
         section_offset = section->PointerToRawData;
         section_raw_size = section->SizeOfRawData;
+
+        // If the section_offset is less than 0x200 it is rounded down to 0.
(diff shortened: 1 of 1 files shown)


plusvic:
Good job!

hillu:
Merging 3a91764bc978f5c70171280595f29697b1edd5be breaks this test, though. (Everything passes again after reverting):

    assert_true_rule_blob(
          "import \"pe\" \
           rule test { \
            strings: $a = { 6a 2a 58 c3 } \
            condition: $a in (pe.entry_point .. pe.entry_point + 1) }",
          PE32_FILE);

PR #421, opened 2016-03-09
Revert "Add support for compiling under cygwin."

cisval:
Reverts plusvic/yara#403


hillu:
What does this commit fix / achieve?

PR #422, opened 2016-03-11
Add Travis CI integration

hillu:
Note that the test currently fails due to a regression introduced by merging PR #420. (This is exactly the type of regression that a CI setup might prevent from happening in the future.)


SleuthKid:
to be really hip you should also embed one of these in the README with the pull:
https://docs.travis-ci.com/user/status-images/

PR #435, opened 2016-04-07
Add yr_filemap_unmap_fd() that leaves the filehandle open

hillu:


plusvic:
Could you add this new function to the API documentation?

hillu:
None of the other `yr_filemap_*` functions are documented so far, so I'll do a separate PR for that first.

plusvic:
Oh, that's right, ignore the previous comment then.

PR #437, opened 2016-04-10
Set a value of zero for pe.number_of_signatures by default

iangcarroll:
This sets a value of zero for `pe.number_of_signatures` when there are no Authenticode signatures. It is currently set to `UNDEFINED` when there are no signatures.


iangcarroll:
Is this an acceptable solution? Or does it need changes?

plusvic:
Sorry for not answering before, I missed this PR. This was already fixed in this other commit f38ac770b8e87548623cfbe3a67c6ff1ae68633b.

PR #438, opened 2016-04-25
pe: rva_to_offset: Stay within bounds if no valid sections are found

hillu:
out-of-bounds reads in pe_valid_dll_name() due to an invalid offset
could lead to segmentation faults for some malformed PE files.

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -441,7 +441,9 @@ int64_t pe_rva_to_offset(
 
   if (rva < lowest_section_rva)
   {
-    return rva;
+    section_rva = 0;
+    section_offset = 0;
+    section_raw_size = pe->data_size;
   }
 
   // Many sections, have a raw (on disk) size smaller than their in-memory size.


plusvic:
I'm not sure if this fix is correct. A PE could have a RVA higher than it's own raw size for all its sections. That's unusual, but could be possible in theory because the RVA is not related to the file layout, it's related to the in-memory layout.

Could you provide more details about the circumstances leading to the segmentation fault?

hillu:
I introduced `lowest_section_rva` in ba60f4e5edd4d1eca9ba2f1ad6f5543aa4d95153 to track the start of the first mapped section in order to catch unmapped data in front of the first section. (`pe.c:439`). If no valid sections are found `lowest_section_rva` stays at its initial value, so before this patch any rva would be considered valid and lead to OOB reads later on.

You can reproduce the problem by just taking the first 1024 bytes of `tests/data/tiny.exe` and scanning them using
<pre>
import "pe"
rule t { condition: pe.exports("foo") }
</pre>

plusvic:
I've truncated the tiny.exe file to 1KB and scan it with the provided rule. The pe_rva_to_offset function was called in this line: https://github.com/plusvic/yara/blob/master/libyara/modules/pe.c#L1080. It returned -1 and pe_parse_imports exited in https://github.com/plusvic/yara/blob/master/libyara/modules/pe.c#L1083 without even reaching pe_valid_dll_name.

However, after taking a look at the code I've noticed that pe_rva_to_offset could return a offset larger than the file size in this line: https://github.com/plusvic/yara/blob/master/libyara/modules/pe.c#L442. 

But the solution you propose seems incorrect to me because its possible for a PE file to have all its sections with a RVA larger than the file size. In that case, when you invoke pe_rva_to_offset with a valid RVA found within one of the the sections, the function is going to incorrectly exit in line 444, because the value for lowest_section_rva would be lower than than the rva passed to pe_rva_to_offset.

plusvic:
My proposed solution is replacing:

````
  if (rva < lowest_section_rva)
  {
    return rva;
  }
````

with:

```
  if (rva < lowest_section_rva)
  {
    section_rva = 0;
    section_offset = 0;
    section_raw_size = pe->data_size;
  }
```

In https://github.com/plusvic/yara/blob/master/libyara/modules/pe.c#L442

The subsequent checks are going to make sure that pe_rva_to_offset doesn't return a value larger than the file size. What do you think?

hillu:
@plusvic, please ignore the bit about `pe_valid_dll_name()`. This happened to me with another truncated sample, sorry for the confusion.

We seem to agree on the root cause (`pe_rva_to_offset` returning values that are larger than the file size) and I think your proposed fix is cleaner.

plusvic:
Fixed in 1497fbdf39f6db3e44344fffd2cb4b3a89c937c0

Issue #441, opened 2016-05-03
pe: Corner cases


hillu:
# Corner cases in importing or exporting names

I have been looking for corner cases in the PE module by auto-generating rules from [pefile](https://github.com/erocarrera/pefile)'s verdict and passing them to YARA. So far, I am using malware samples from Virusshare for this.

This issue is intended to track those differences.
## The set of valid characters is not complete.

An ASCII minus sign (2e) is a perfectly valid character for DLL names. Commit da6790b0b2d0896a4857ba193ad71b2990e47b79 addresses this.
## Trailing spaces in imported DLLs are ignored.

A simple "hello world"

<pre>
#include <stdio.h>
int main() { puts("hello world"); return 42 }
</pre>

(built using MinGW GCC) can still be loaded and executed after the following change, at least on XP:

<pre>
 00005570: 1460 0000 1460 0000 1460 0000 1460 0000  .`...`...`...`..
-00005580: 1460 0000 6d73 7663 7274 2e64 6c6c 0000  .`..msvcrt.dll..
+00005580: 1460 0000 6d73 7663 7274 2e64 6c6c 2020  .`..msvcrt.dll  
 00005590: 2020 2020 0000 0000 0000 0000 0000 0000      ............
</pre>

Things are not as simple as I initially thought: Chaniging `"msvcrt.dll"` to `"msvcrt"` apparently causes the PE loader to silently add the `".dll"` as the library is still found. In this case, adding a trailing space (resulting in `"msvcrt\x20"`) causes the PE loader to look for `"msvcrt\x20.dll"` -- which leads to an error message.
## File name without terminating null character at EOF

`VirusShare_00000/VirusShare_39440a85aa6e4556858110163bfe99d5` contains the filename `shell32.dll` right at the end of file, without a terminating null character. YARA ignores this name altogether, pefile thinks that the DLL name is `shell32.dl`, with the last letter mising.. The executable loads fine and it does import `shell32.dll`.

Commit c3eb21b0d6a44cb8f50b751c2c71ae69e0264792 addresses thiis.
## Strings (file / function name) without terminating null character at end of section

This is a variation on the previous observation.
`VirusShare_00000/VirusShare_0fd036bc0df937872913f0b9cc26dcd5` has a small overlay that becomes part of what YARA thinks is a function. (pefile gets this [wrong](https://github.com/erocarrera/pefile/issues/114), in a different way.)

<pre>
[...]

PR #442, opened 2016-05-03
Fixes for three more pe corner cases

hillu:
See #441

### DIFF
--- a/libyara/Makefile.am
+++ b/libyara/Makefile.am
@@ -84,6 +84,14 @@ lib_LTLIBRARIES = libyara.la
 
 libyara_la_LDFLAGS = -version-number 3:6:0
 
+BUILT_SOURCES = \
+  lexer.c \
+  hex_lexer.c \
+  re_lexer.c \
+  grammar.c \
+  hex_grammar.c \
+  re_grammar.c
+
 libyara_la_SOURCES = \
   $(MODULES) \
   grammar.y \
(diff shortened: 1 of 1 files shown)


libyara/modules/pe.c
```diff
@@ -1052,7 +1052,7 @@ int pe_valid_dll_name(
     }
   }
 
-  return (l > 0 && l < n);
+  return (l > 0 && l <= n);
```
plusvic: I think this introduces a potential buffer overrun in line 1117, where yr_strdup is called with dll_name as its argument. If dll_name is at the end of the file and without a null termination, pe_valid_dll_name will return true and yr_strdup can be invoked with a non-null-terminated string.

hillu: You're correct, of course. Turns out that the use of a `strdup` style function is wrong here. But then, we didn't know about the string-at-EOF corner case before...

libyara/modules/pe.c
```diff
@@ -438,22 +431,9 @@ int64_t pe_rva_to_offset(
...
-    section_raw_size = pe->data_size;
-  }
-
-  // Many sections, have a raw (on disk) size smaller than their in-memory size.
-  // Check for rva's that map to this sparse space, and therefore have no valid
-  // associated file offset.
-
-  if ((rva - section_rva) >= section_raw_size)
-    return -1;
+  // Apparently, the loader maps data that resides between sections in
```
plusvic: The returned offset is going to be equal to the passed RVA for any RVA not contained within a section. What if the value passed to pe_rva_to_offset is an offset within the PE overlay? The value is going to be accepted as a valid RVA and is going to be returned as the corresponding offset. But I think the PE overlay is not mapped into memory by the loader? I don't have a Windows machine around to verify it. Can you confirm if that's the intended behavior?

hillu:
Yeah, .I  going to do some experiments.

PR #447, opened 2016-05-13
Fix sized string

### LINKED ISSUE #446: SIZED_STRING layout causes incompatible compiled rules between 32 and 64 bit architectures
Consider the following simple rule

<pre>rule t { condition: str == "asdf" }</pre>

compiled like this (b32: i386, b64: amd64, both on Linux):

<pre>
b32/yarac -d str=fdsa tests/data/string.yar tests/data/string-32.yac
b64/yarac -d str=fdsa tests/data/string.yar tests/data/string-64.yac
</pre>

Both following tests should show a match for the rule `t`:

<pre>
b64/yara -d str=asdf tests/data/string-32.yac /dev/null
b32/yara -d str=asdf tests/data/string-64.yac /dev/null
</pre>

But they don't. The reason for this is that the layout for SIZED_STRING is different between i386 and amd64 architectures

hillu:
Closes #446

### DIFF
--- a/libyara/include/yara/sizedstr.h
+++ b/libyara/include/yara/sizedstr.h
@@ -18,6 +18,7 @@ limitations under the License.
 #define _SIZEDSTR_H
 
 #include <stddef.h>
+#include <inttypes.h>
 
 //
 // This struct is used to support strings containing null chars. The length of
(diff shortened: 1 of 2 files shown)


hillu:
@plusvic I'd love some feedback about these changes.

plusvic:
Your changes are ok, the only improvement I'm considering is moving the definition of SIZED_STRING to the types.h file and use explict packing for that structure (i.e. putting it inside the #pragma pack(8) block)

hillu:
Renaming `SIZED_STRING` to `YR_SIZED_STRING` and moving it to `types.h` seems sensible to me.

PR #453, opened 2016-05-23
create cache layer for non-string input hashing functions

bmagistro:
When trying to run rule sets that contain thousands of hashes, processing time can significantly increase.  Allowing the hash module to cache the last computed value, offset, and length it can quickly return the already computed hash value avoiding the cost of recomputing for each rule.

This doesn't add any value to folks with various offsets/lengths.  For users with the same offset + size for rules, this can decrease processing time.

### DIFF
--- a/libyara/modules/hash.c
+++ b/libyara/modules/hash.c
@@ -25,9 +25,32 @@ limitations under the License.
 #endif
 
 #include <yara/modules.h>
+#include <yara/mem.h>
+#include <stdbool.h>
 
 #define MODULE_NAME hash
 
+
+typedef struct _CACHED_HASH {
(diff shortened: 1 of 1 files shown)


plusvic:
This solution only stores a single hash of each type (sha256/sha1/md5). A solution using a hash table is implemented in 22ce5e09dd49ccfb175524e8c6066984d1ecfb63

bmagistro:
We debated trying to do something like this but decided against it on the assumption rules likely wouldn't have that many offset/length variations.  There was also the part of figuring out how to implement the hash table.  Thanks for putting something together and pulling it into the main branch!

PR #463, opened 2016-06-21
Update README.md

sierkb:
Add Apple in section "Who's using YARA", since YARA rules are also used and provided in Apple's recent XProtect (file: XProtect.yara).

PR #464, opened 2016-06-21
Update README.md

sierkb:
Add Apple in section "Who's using YARA", since YARA rules are also used and provided in Apple's recent XProtect (file: XProtect.yara).


sierkb:
Please revert/remove #464 because of dupe of  #463.

PR #467, opened 2016-06-23
Travis: Show test-suite.log on failure

hillu:


hillu:
Should `ARENA_FILE_VERSION` be increased because of the different `SIZED_STRING` layout?

plusvic:
> Should ARENA_FILE_VERSION be increased because of the different SIZED_STRING layout?

Yes, we should

hillu:
`ARENA_FILE_VERSION` is taken care of in #468

PR #472, opened 2016-06-27
Changes needed to build with OpenSSL 1.1

hillu:
~~Please don't merge -- I am not sure about the `i2c_*` / `i2d_*` change yet.~~

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -42,6 +42,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <openssl/bio.h>
 #include <openssl/pkcs7.h>
 #include <openssl/x509.h>
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#define X509_get_signature_nid(o) OBJ_obj2nid((o)->sig_alg->algorithm)
+#endif
 #endif
 
 #include <yara/pe.h>
@@ -1267,7 +1270,7 @@ void pe_parse_certificates(
           pe->object,
           "signatures[%i].version", counter);
 
-      sig_alg = OBJ_nid2ln(OBJ_obj2nid(cert->sig_alg->algorithm));
+      sig_alg = OBJ_nid2ln(X509_get_signature_nid(cert));
 
       set_string(sig_alg, pe->object, "signatures[%i].algorithm", counter);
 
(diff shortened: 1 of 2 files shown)


wxsBSD:
Thank you! I started down this path a few months ago but ended up getting sidetracked onto some other things and never got back to this!

hillu:
@wxsBSD can you shed some light on what `i2c_ASN1_INTEGER()` is supposed to do here?

wxsBSD:
It was used to convert the internal representation of the serial number to the human readable string form.

On a related note, there are a couple of places in that file where i2c_ is mentioned in the comments, you will want to clean those up too.

hillu:
@wxsBSD I am not even sure if using `i2d_ASN1_INTEGER()` is the correct replacement -- do we want a hex dump of the DER-encoded INTEGER?

wxsBSD:
It has been too long and I don't remember the terminology, so I can't say. We want the format that VirusTotal uses, which is also the format most other tolls use. I can double check this patch later today to verify it doesn't break some of my old signatures.

hillu:
Apparently, `i2c_ASN1_INTEGER` puts the "raw data"  into a buffer while `i2d_ASN1_INTEGER` prefixes that data with DER type and length information. I have updated my patch to use `i2d_ASN1_INTEGER`. In my tests, I got the same result as before for `4bfe05f182aa273e113db6ed7dae4bb8`  (negative serial) and for a legit psexec binary.

This change can be merged, as far as I'm concerned.

libyara/modules/pe.c
```diff
@@ -1336,6 +1344,10 @@ void pe_parse_certificates(
                     (char*) serial_ascii + 3 * j, 3, "%02x", serial_bytes[j]);
               }
 
+              fprintf(stderr, "signature[%i].serial = '%s'\n", counter, serial_ascii);
```
wxsBSD: Probably want to remove this. :)

wxsBSD:
Thank you! I was going to look into this today (been backlogged at $job) but you seem to have covered my thoughts well. As long as serials are matching before and after this change I'm happy with it.

libyara/modules/pe.c
hillu: done.

wxsBSD:
I haven't tested on an OpenSSL 1.1 system (the VM has gone missing), but I did test on my OS X install running OpenSSL 0.9.8 (plus whatever patches Apple has applied) and tested it against a signed sample I had laying around. No regressions AFAICT. Thanks for looking into this, it would be great to get it into 3.5.0 as we will be seeing these OpenSSL changes become more widespread.

PR #501, opened 2016-08-12
Add number_of_imports to PE.

wxsBSD:
Expose the number of imports parsed by YARA in the PE module. Sometimes this can
be useful to know, especially if there are zero imports as that may be a sign
that something is obfuscated. For example,
f0501b0b3990dab0d8644729da02acb8ea5f39765b820626dc8873f60f787980 has zero
imports because the import table has been nulled out.


plusvic:
@wxsBSD Could you add it to the documentation?

PR #505, opened 2016-08-16
Add pe.overlay.offset and pe.overlay.size (closes #432)

### LINKED ISSUE #432: Add ability to process overlays in PE Module
It would be neat to have the ability to write rules that can identify appended data in a pe file (overlays)
something like:

```
rule overlay
{
   condition:
     pe.has_overlay
}

```

or access to overlay information such as offset and length (pe.overlay.length and pe.overlay.offset)

merces:
An overlay is data appended to the end of a file. This PR adds the ability to use pe.overlay.offset and pe.overlay.size in Yara rules. For example:

```
import "pe"

private rule has_overlay {
	condition:
		pe.overlay.offset
}

rule big_overlay {
	condition:
		pe.overlay.size > 1024
}

rule overlay_bytes {
	strings:
		$bytes = { 90 41 42 50 56 }
	condition:
		$bytes at pe.overlay.offset
}
```

It closes the issue #432 (also discussed at https://groups.google.com/forum/#!topic/yara-project/5AWlwXyy1oY)

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -224,6 +224,22 @@ Reference
 
     *Example: pe.sections[1].characteristics & SECTION_CNT_CODE*
 
+.. c:type:: overlay
+
+    .. versionadded:: 3.6.0
+
+    A structure containing the following integer members:
+
+    .. c:member:: offset
+
+        Overlay section offset.
+
+    .. c:member:: size
+
+        Overlay section size.
+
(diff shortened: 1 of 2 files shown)


plusvic:
Could you add this new feature to the documentation in docs/modules/pe.rst?

merces:
Certainly. It is done. I just need to confirm what the new release version is going to be? 3.6.0? I noticed this information is essential in the documentation.

Thanks,
Fernando

plusvic:
Yes, it will be released in version 3.6.0

PR #513, opened 2016-09-06
Allow MAX_THREADS to be set arbitrarily

ewil:
This allows the 32-thread upper limit to be adjusted by converting the int32_t thread mask to a byte array. A potential drawback is that if the rule structure is ever serialized it will be binary incompatible when MAX_THREADS is altered due to the differing structure.

### DIFF
--- a/libyara/compiler.c
+++ b/libyara/compiler.c
@@ -590,7 +590,7 @@ YR_API int yr_compiler_get_rules(
   yara_rules->match_table = rules_file_header->match_table;
   yara_rules->transition_table = rules_file_header->transition_table;
   yara_rules->code_start = rules_file_header->code_start;
-  yara_rules->tidx_mask = 0;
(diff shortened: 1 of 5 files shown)


plusvic:
Hi @ewil, Coverity found a subtle bug in your commit that I fixed in 89c70ae6e05ddc2d79d7ee0f6a4de10f3cd36539. The `tidx < MAX_THREADS` check must be executed before `YR_BITARRAY_TEST(rules->tidx_mask, tidx)` to avoid an out-of-bound access in the tidx_mask array.

Those bugs are hard to detect by reading the code, I introduced a similar one in the past and spent a few hours pulling out my hair before finding the problem (wasn't using Coverity back then). Fortunately Coverity does a very good job finding those subtle bugs.

ewil:
Nice, I should've seen that, thanks!

PR #521, opened 2016-09-17
Make our memmem implementation available in dotnet.c

hillu:


wxsBSD:
Why is this needed? I don't think I'm using mermen there?

hillu:
https://github.com/VirusTotal/yara/blob/master/libyara/modules/dotnet.c#L54

wxsBSD:
Oh, right. I forgot about that. 😉

PR #525, opened 2016-09-21
Travis: Cross compile with MinGW

hillu:


plusvic:
Travis-CI builds are failing after this change.

This failure is particularly intriguing:

*** Error in `/usr/bin/x86_64-w64-mingw32-ld': free(): invalid pointer: 0x0000000001ec97f8 ***
collect2: error: ld terminated with signal 6 [Aborted], core dumped

A core dump in the linker itself?

hillu:
Strange. It clearly worked before... https://travis-ci.org/VirusTotal/yara/builds/161631311

plusvic:
The problem was apparently related to binary files generated by previous builds. A "make clean" before "make" solved the issue (see cfecd1105394d2697c95911786c947737e477bac). I thought that each build was performed in an isolated and clean environment, but doesn't seem to be the case.

Issue #523, opened 2016-09-20
No CI for Windows


kallanreed:
The Windows project is somewhat broken right now - *ix specific code introduced by contributors and it looks like there was an update to OpenSSL that may have caused some import regressions.

Ideally Windows would be CI, Travis doesn't support Windows. I'm taking a look at AppVeyor which is free for open source projects.

If I can get it working on my fork, it would be useful to add AppVeyor CI to keep this project from regressing on Windows. Also, AppVeyor could be used to build Windows releases and deploy NuGet packages.

https://ci.appveyor.com/project/kallanreed/yara

plusvic:
That would be a great contribution. Most people don't have a Windows setup
to test their changes and it would be nice if we can detect this kind of
issues promptly.

hillu:
Related to this: Cross-building using MinGW on Travis. #525

kallanreed:
@plusvic check out https://ci.appveyor.com/project/kallanreed/yara/history and https://ci.appveyor.com/project/kallanreed/yara/build/1.0.9 (for details) and see if this is sufficient. I'm trying to figure out PR integration, tests and even have it publish binaries to the feed.

If you want to integrate, you'll need to set up an account on AppVeyor with an account that can admin the yara repo.

plusvic:
I've created an account but need to take a look on how to configure everything. Do you have any guidelines?

kallanreed:
I'll send the yaml file from my set up. It's also possible to add a yaml file to the root (like travis) and have it "just work" but I haven't tried that yet.

kallanreed:
I've added the appveyor.yml to PR #526 with info on getting everything set up.

Issue #514, opened 2016-09-07
assert failure


mccoysc:
assert failure at line 467 in file exec.c of yara version 3.5 when rule is:
rule fire2013 : webshell
{
    meta:
        author      = "Vlad https://github.com/vlad-s"
        date        = "2016/07/18"
        description = "Catches a webshell"
    strings:
        $a = "eval(\"\x65\x76\x61\x6C\x28\x67\x7A\x69\x6E\x66\x6C\x61"
        $b = "yc0CJYb+O//Xgj9/y+U/dd//vkf'\x29\x29\x29\x3B\")"
    condition:
        all of them
}

and the file content to scan is :
eval(\"\x65\x76\x61\x6C\x28\x67\x7A\x69\x6E\x66\x6C\x61

the code which bug occoured like this 

457      case OP_OBJ_LOAD:
458        identifier = _(char__)(ip + 1);
459        ip += sizeof(uint64_t);
460
461        r1.o = (YR_OBJECT_) yr_hash_table_lookup(
462            context->objects_table,
463            identifier,
464           NULL);
465
467        assert(r1.o != NULL);
468        push(r1);
469        break;

plusvic:
I'm unable to reproduce this issue. With that rule YARA 3.5 is working fine for me. Are you able to reproduce it with this single rule? Or are you running YARA with more rules besides this one?

mccoysc:
https://github.com/Yara-Rules/rules



364589886@qq.com
 
From: Victor M. Alvarez
Date: 2016-09-07 15:32
To: VirusTotal/yara
CC: mccoysc; Author
Subject: Re: [VirusTotal/yara] assert failure (#514)
I'm unable to reproduce this issue. With that rule YARA 3.5 is working fine for me. Are you able to reproduce it with this single rule? Or are you running YARA with more rules besides this one?

mccoysc:
https://github.com/Yara-Rules/rules 

all rules dwonload from this URL.



364589886@qq.com
 
From: Victor M. Alvarez
Date: 2016-09-07 15:32
To: VirusTotal/yara
CC: mccoysc; Author
Subject: Re: [VirusTotal/yara] assert failure (#514)
I'm unable to reproduce this issue. With that rule YARA 3.5 is working fine for me. Are you able to reproduce it with this single rule? Or are you running YARA with more rules besides this one?

plusvic:
Still unable to reproduce it. A stack trace would be great.

mccoysc:
[...]

PR #526, opened 2016-09-21
Update VS2015 project to set module variable definitions

kallanreed:
Some additional fixes for VS compilation as well.


kallanreed:
This and #527 will address the assert logged in #514

kallanreed:
@plusvic, this should also take care of the basic Windows CI #523 as well. AppVeyor supports nicer GitHub integration (like Travis) but that requires tokens and things that I didn't want to bother with. This will simply add a comment to the PR thread indicating if the build passed or failed.

All you should have to do is create the yara project in AppVeyor. It will pick up the build config from the yml file.

From the documentataion:
>Project builds can be configured by either appveyor.yml or on the user interface. ...

>If you have appveyor.yml in your repo it will override all settings made on the UI unless explicitly disabled by Ignore appveyor.yml. The only exceptions are environment variables and notification settings. Environment variables defined on UI are getting merged with those ones defined in appveyor.yml. Variable values with the same names are getting overridden with values from UI.

plusvic:
Already addressed in other commits.

PR #531, opened 2016-10-04
dotnet user strings and assembly refs

wxsBSD:
Add support for parsing assembly ref information and user strings.

Assembly references are references to external assemblies, including version information and a public key or token which identifies the author.

User strings are exactly what they sound like. Strings in the binary are put into a special stream (#US) and we can now parse them. Note that they are all wide character strings so they are currently slightly cumbersome to use but I'm working on improving that.


wxsBSD:
Any plans on merging this? It adds some interesting features for dotnet analysis and will serve as the basis for some other stuff I have in mind. In particular we discussed adding support for wide modifier in conditions and I'd also like to explore string deobfuscation techniques in a module. For example, I think it would be great to be able to say:

for any i in (0..dotnet.number_of_user_strings):
  (string.incrementing_xor(dotnet.user_strings[i], 0x41) == "SECRET STRING" wide)

Having the user strings parsed out is the first step in this.

plusvic:
Yes, the plan is merging this module before launching version 3.6.0. I'm
just merging the low-hanging fruits first.

wxsBSD:
Awesome! I'll start doing some work on the stuff which is building upon this.

plusvic:
@wxsBSD some test cases for the dotnet module would be great. Can you include a small .NET file in /tests/data and create some simple tests for it?

Issue #516, opened 2016-09-13
Segfault on invalid imports


fumfel:
Invalid import `"\x00"` causes a **segfault**.

Tested on yara 3.5.0 (cloned today from master).

Test data:
`import "\x00"`

Valgrind output: https://gist.github.com/fumfel/497b0863667d4cd4a1819872328e5b6d

Bug exists in **yara/libyara/hash.c**:
- https://github.com/VirusTotal/yara/blob/f502bdf4903545cfc75dfe15bed650138da53a81/libyara/hash.c#L73
- https://github.com/VirusTotal/yara/blob/03c2aac2c1253b096342cf4b745d66e448c7d110/libyara/hash.c#L183
- https://github.com/VirusTotal/yara/blob/03c2aac2c1253b096342cf4b745d66e448c7d110/libyara/hash.c#L93

Bug found with american fuzzy lop.

fumfel:
Problem still exists, tested repo HEAD: a0bb3836f16e3c5d0c2a1da097a1ebacbebc3a94.

wxsBSD:
Fix is up. I'd be curious if you are able to trigger this any other way, especially via modules parsing files. :)

PR #533, opened 2016-10-06
Fix issue #516.

wxsBSD:
When attempting to import "\x00" there is a crash due to an integer underflow.
The "\x00" is passed to yr_parser_reduce_import() as a SIZED_STRING, with a
c_string member pointing to "\x00" and a length of 1. Eventually
yr_hash_table_lookup() is given the c_string pointer, which then calls strlen()
on it and passes the length (0) into yr_hash_table_lookup_raw_key(). So at this
point yr_hash_table_lookup_raw_key() is given a pointer to a string containing
"\x00" and a length of 0, which it then passes to hash() which ultimately
subtracts 1 from the 0 length and then looks that value up in the byte_to_int32
array, which is obviously bogus.

This fixes the problem by checking for a key_length of 0 in
yr_hash_table_lookup_raw_key() before attempting to hash it and returns NULL.
The NULL value is now checked for in yr_parser_reduce_import() and a compiler
error is now thrown.

I did a quick check for other places where a similar problem may pop up,
modules are especially concerning, but I haven't found any that don't look right
to me.

### DIFF
--- a/libyara/hash.c
+++ b/libyara/hash.c
@@ -180,9 +180,6 @@ YR_API void* yr_hash_table_lookup_raw_key(
   YR_HASH_TABLE_ENTRY* entry;
   uint32_t bucket_index;
 
-  if (key_length == 0)
-    return NULL;
-
   bucket_index = hash(0, key, key_length);
 
   if (ns != NULL)
--- a/libyara/hash.c
+++ b/libyara/hash.c
@@ -180,6 +180,9 @@ YR_API void* yr_hash_table_lookup_raw_key(
   YR_HASH_TABLE_ENTRY* entry;
   uint32_t bucket_index;
 
+  if (key_length == 0)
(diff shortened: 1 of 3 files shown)


wxsBSD:
I'll look into the test failures tonight.

wxsBSD:
Ugh, that was a silly bug for me to introduce. I've got a better fix in place that doesn't break the tests (thank you for having them) and also adds a test to ensure we don't regress in the future.

edeca:
Unsure if you also fixed but importing an empty string has the same effect, likely for the same reason.

wxsBSD:
Are you sure about that? I just tested with latest master and can not replicate the problem importing an empty string.

plusvic:
I've fixed this issue in 2a36c168bb2097c0831d4ddf24a6c04bafe28dca

Issue #493, opened 2016-08-04
Yara crashes on big-endian architectures


hillu:
Builds of Yara 3.5.0-2 on Debian's build servers fail their tests (segfault), see https://buildd.debian.org/status/package.php?p=yara

hillu:
This is not a regression: Yara 3.4.0 also crashes on Debian unstable/powerpc architecture when checking a trivial rule `rule test { condition: true }`.

hillu:
It looks like something is seriously wrong with encoding/decoding pointers for byteode and with storing pointers on the stack.

`test-rules`:
<pre>
Program received signal SIGSEGV, Segmentation fault.
yr_execute_code (rules=rules@entry=0x1013ea78, 
    context=context@entry=0xfffee7a8, timeout=timeout@entry=0, 
    start_time=start_time@entry=1470348868) at exec.c:448
448               rule->t_flags[tidx] |= RULE_TFLAGS_MATCH;
(gdb) bt
#0  yr_execute_code (rules=rules@entry=0x1013ea78, 
    context=context@entry=0xfffee7a8, timeout=timeout@entry=0, 
    start_time=start_time@entry=1470348868) at exec.c:448
#1  0x10015c48 in yr_rules_scan_mem_blocks (rules=rules@entry=0x1013ea78, 
    iterator=iterator@entry=0xfffef3d0, flags=flags@entry=0, 
    callback=callback@entry=0x10008aa0 <count_matches>, 
    user_data=user_data@entry=0xfffef408, timeout=timeout@entry=0)
    at rules.c:470
#2  0x10015f30 in yr_rules_scan_mem (rules=rules@entry=0x1013ea78, 
    buffer=buffer@entry=0x10045d08 "dummy", buffer_size=buffer_size@entry=5, 
    flags=flags@entry=0, callback=callback@entry=0x10008aa0 <count_matches>, 
    user_data=user_data@entry=0xfffef408, timeout=timeout@entry=0)
    at rules.c:582
#3  0x10008d2c in matches_blob (
    rule=0x10042ec4 "rule test { condition: true }", blob=<optimized out>, 
    len=<optimized out>) at tests/util.c:119
#4  0x10001a9c in test_boolean_operators () at tests/test-rules.c:37
#5  main (argc=<optimized out>, argv=<optimized out>)
    at tests/test-rules.c:1350
</pre>

`test-pe`:
<pre>
Program received signal SIGSEGV, Segmentation fault.
[...]

PR #535, opened 2016-10-09
Big endian fixes

hillu:
This set of patches fixes a number of issues on big-endian architectures as described in #493. I have successfully built and tested these with Debian/unstable on powerpc and ppc64 (not ppc64el).

The pe module has not been fixed to work on big-endian, so far; I have disabled the pe-specific tests for the time being. I'm going to fix the module code if these fixes get merged.


hillu:
Apparently, the endianess check does not work on Mac OS X.

hillu:
Build and tests pass on OSX now.

Issue #350, opened 2015-10-01
Thread-safe setting of variables


hillu:
If I read `libyara/rules.c` correctly, when there are multiple threads
sharing a ruleset for scanning, there is no way to safely define
external variables.

I use variable definition to pass trivial information such as file or
process names that are used within some of my rules. Of course, I
could use one copy of the ruleset per thread, but in my use case that
would mean a much larger memory footprint. (Also, this would defeat
the purpose of YARA's support for multiple threads.)

The `yr_rules_define_*_variable` functions operate on a global
structure hanging off the `YR_RULES` struct without any locking. Even
though `yr_rules_scan_mem_blocks` eventually copies everything into
`context.objects_table` before starting the actual scan, I see no way
to ensure that the variables hanging off `rules->externals_list_head`
are concsistent -- even if I protect the individual
`yr_rules_define_*_variable` calls with a mutex.

Am I missing something?

plusvic:
You are right, yr_rules_define_*_variable aren't thread-safe. You can't assign a thread-specific value to a given variable, all threads share the same value which should be assigned by a single thread and shared by all threads calling yr_rules_scan_*. 

As you said, the most important limitation here is that you can't pass values that depends on the file being scanned, like filename, or whatever.  This could be solved by providing a separate structure to each thread to store variable values, but that would break existing code that assumes that once the main thread assigns value to a variable all other threads inherit that value.

bmagistro:
I have hit this issue as well and will likely just be making multiple copies of the rules.  Is there any chance the yr_rules_scan_* functions can be extended to include an (optional) variable list?  Under the hood it may be possible to keep the calls the same and point this new argument to the rules/compiler structure's external variable list when it isn't present.  The thinking being this might avoid breaking existing code while allowing an additional path to be added where variables could be defined per scan call.  Yes this means that all variables would need to be defined at all times even if not in use but I see this as a issue for the caller and not an issue for yara to solve.

Thoughts?  If this sounds like a possible approach I'd be willing to try and code some of this up but not sure how successful I'd be.

hillu:
@bmagistro What you describe is pretty much the path I had in mind.

bmagistro:
@hillu I think I have something, I'm just now starting to test things using the new functions and providing the vars to the scan function.

PR #539, opened 2016-10-15
Ext var scan

bmagistro:
Believe this provides a solution to #350.  Needs additional testing/review and likely more documentation.


plusvic:
In this solution the `yr_rules_scan_*_e` functions completely replace all variable definitions with the ones provided in the `external_vars` argument. Any variable previously defined at compile time or with `yr_rules_define_*` is lost. All external variables associated to the `YR_RULES` struct are completely ignored. I would prefer a solution where each thread could override the value for some external variable, but the variables defined during compile time are still valid.

Besides, having to expose a new structure `YR_EXT_VARS` in the API seems like unnecessary and confusing. People would ask what's the difference between YR_EXT_VAR* and YR_EXTERNAL_VARIABLE*.

The solution I have for this is adding a set of functions `yr_rules_define_*_variable_ex` that would receive an additional argument indicating whether the variable definition is for all threads using the `YR_RULES` struct or only of the current thread. The existing  `yr_rules_define_*_variable` would simply call `yr_rules_define_*_variable_ex`. That's the only change from the API standpoint, no need to define and document new data structures.

bmagistro:
That is correct about replacing all of them, the case of a union of the two did not occur to me when coming up with this solution.  I don't disagree on the additional structure - it did feel a little clunky, but it was what made sense to me at the time.  In that regard being able to override only select variables does make sense.

I was under the impression that you could not define a new variable with "yr_rules_define_*", as it returns "ERROR_INVALID_ARGUMENT" today if it doesn't already exist/wasn't added by "yr_compiler_define_*_variable" when building the rule set.

The rules structure is still shared by all threads, you would need something else to make that association.  Going off your description, perhaps expanding "yr_external_variable" to include "tidx" to be -1 or any thread else value of "tidx" then when calling scan_mem_blocks the loop could pull in any that are -1 and any that match that thread id.  There would need to be an ordering so that you get all -1 first followed by the overrides. Is that more along the lines of what you were thinking?

hillu:
The idea to have optional thread-specific variable overrides sounds elegant at first, but you can't tie it to `tidx` because the thread<->tidx association only exists as long as `yr_rules_scan_mem_blocks` runs.

How about this:
- Add a limited number of slots of `YR_EXTERNAL_VARIABLE` chains for overriding the globally-set variables.
- Use some bits of the `flags` argument to the `yr_rules_scan_*`  functions to indicate which **one** of these chains should be used. 0 means that only the global set should be used.

This would require no extra data structures, no extra variants of the scan functions, just variants of the `yr_rules_define_*_variable` that use an extra parameter to specify the slot. And probably functions for allocation and cleanup of the local variable chains.

bmagistro:
I was looking at tidx the other day and wanted to wait till someone could chime in.  That is exactly what I thought I saw but wanted confirmation.

What I am thinking about now is adding the actual thread id or -1 for global to `YR_EXTERNAL_VARIABLE`.  If you call `yr_rules_define_*_variable` it would set thread to -1.  If you call `yr_rules_define_*_variable_override` it would use the thread id, this means you can't build up the overrides in a different thread than the one that is going to call scan.  When it goes through `yr_rules_scan_mem_blocks` it would load all globals then apply any thread specific overrides.  The call to 'yr_finalize_thread` would be updated to include looping through and removing obsolete variables (don't remember if there is/was a lock on rules to support this).  Thoughts?

plusvic:
What @hillu proposes have one advantage: you can override the variable in one thread and use the overriden value in another thread. The disadvantage is that the user must deal with allocating and de-allocating `YR_EXTERNAL_VARIABLE` structures, and take care of passing the correct values to
`yr_rules_scan_* `  for the slot. It's also harder to understand.

The solution proposed by @bmagistro requires that variables are overriden from the thread that is going to use them, but yara takes care of all the allocation/deallocation of  `YR_EXTERNAL_VARIABLE` structures. The concept is pretty straightforward: you can define variables for all threads using the `YR_RULES`, or just for the current thread. I'm more inclined towards this approach.

However, what @hillu pointed out is true, as the thread<->tidx association is not done until `yr_rules_scan_mem_blocks` is called, we can't rely in the tidx. We could use thread local storage (TLS) instead. But where do you store the TLS index? In `YR_RULES`? In that case you can't free the resources in `yr_finalize_thread`, you need to get a function called by each thread they are done with the `YR_RULES` struct. A sort of per-thread `yr_rules_destroy`, not very elegant.

I'll keep thinking about this tomorrow. Any idea is welcomed.

bmagistro:
I think I am going to try and start on this using the thread id (posix thread id `pthread_self`) and see where things go.  I think the two (three?) possible solutions that don't involve a per thread rule set are listed (at least the ones I can think of).

bmagistro:
Finally getting back to working on this and think I'm missing something with `yr_arena_*`.  The `YR_EXTERNAL_VARIABLE` starts as an arena and when calling compile, the base address is type casted and stored.  What I was envisioning was being able to append another entry to this structure when doing the override but it's not clear to me how (not sure it's even possible) to append another entry after calling compile without changing what is stored to the arena.

bmagistro:
I believe I got this working.  If you guys say okay, I'll update the docs and add those into this pr.

hillu:
If I understand your code right, you are pinning local variables to the current thread. I do not particularly like that approach. It makes the feature hard to use in environments where threads are hidden behind an abstraction layer (e.g. Golang's goroutines).

hillu:
If I understand the patch correctly, the set of local variables is tied
to the thread from which the yr_rules_scan_* function is called. I think
that this approach has at least two problems:

1. The user has to go through extra steps to be able to use the same
variable-set from two different threads.

2. Yara is commonly used from higher-level languages where the runtime
may have abstracted away the concept of threads, e.g. Golang's
goroutines. I wrote and maintain a set of Yara bindings for Golang
(https://github.com/hillu/go-yara) and I do not know if I would be able
to make thread-bound local variables usable to users of my library at
all.

Cheers,
-Hilko

bmagistro:
Any updates/plans on this or #559 being merged in?

plusvic:
@bmagistro the API seems reasonable, but the implementation has a number of issues:

* The variable overrides for all threads are mixed in a single list, which means that when a thread wants to override a new variable it needs to spend time iterating a long list where most of the items will belong to another thread. This can be easily avoided by using thread local storage.

* The yr_compiler_define_x_variable functions are setting the current thread id to defined variables. This means that the only thread that will see those variables is the one compiling the rules, any other thread won't see those variables as they belong to another thread. The desired behaviour is that any variable defined during compile time are seen by all threads using the rules.

* Functions yr_rules_define_local_x_variable are very similar, but the same code is repeated over and over.

Besides, I would like to discuss what @hillu mentioned in:

> 1. The user has to go through extra steps to be able to use the same
> variable-set from two different threads.

> 2. Yara is commonly used from higher-level languages where the runtime
> may have abstracted away the concept of threads, e.g. Golang's
> goroutines. I wrote and maintain a set of Yara bindings for Golang
> (https://github.com/hillu/go-yara) and I do not know if I would be able
> to make thread-bound local variables usable to users of my library at
> all.

I don't see any issue with point 1. For using the same variable-set in two different threads the only thing you need to do is defining the variables in any thread using yr_compiler_define_x_variable functions at compile-time or yr_rules_define_x_variable from any thread.

But point 2 is really important, is true that higher-level languages could have a thread-like abstraction not backed by OS threads, it could be an event loop running coroutines. So, probably it makes more sense to use an approach based in passing a set of variable definitions to yr_rules_scan_*

bmagistro:
@plusvic  regarding the implementation

- If you want to add another variable it does require iterating over a potentially long list of variables.  There was one version committed that allowed each thread to have their own structure but development moved away from that due to it having an additional structure that would need to be in the api and exposed.  See comments around Oct 16, but it had several issues related to its implementation.
- This point is not correct.  External variables defined either in the compiled rule set or using the `yr_compiler_define_x_variable` are stored in `external`.  Variables defined by the `yr_rules_define_local_x_variable` are stored in `external_overrides`.  Since they share the same variable structure, and while not required, variables defined with `yr_compiler_define_x_variable` insert that thread's id but the id is never checked when iterating through.  I do see where having that line in there could cause some confusion.  Reference libyara/rules.c line 401 (original) or 808 (new) in the diff.  You will then see a second section that loads in the overrides on top of the global external variables leaving you with the ability to have a variable shared across all threads.
- I agree the code between `yr_rules_define_local_x_variable` is very similar, the delta is in the variable types and if a malloc call is needed (strings).  It might be possible to abstract that out a little and make it cleaner.  At the time of writing, I was after function.

Regarding comments from @hillu 

- I agree with you on point 1.  My thinking was the whole point of creating thread local variables was that you needed unique sets between threads.  For my use case, there is/was never a scenario where I would create one set of variables and use that same set in threads 1 and 2.  There might be some variables shared but they would be defined globally and not be overridden.
- For point 2, this is one that I struggle with...Yara is a C based library/tool so use of threads and associated ids doesn't seem unreasonable to me.  I guess I would want to know/understand when Yara is being used in Go or Python, are threads even in play?  I know there are wrappers around Yara for various languages but don't know to what extent if any threads are in play when they are being used.  All of my use has been in a C/C++ context.

Another solution might be to allow the caller to pass in an identifier to the `yr_rules_define_local_x_variable` and associated `scan` calls.  Going from memory that would look closer to what @hillu provided in #559 and allow for the same override set to be used in multiple places.  The posix thread id could not be used in that scenario as there is a specific compare function that should be used.  What it doesn't allow for (and neither did this implementation) was variable1 to be used in in contexts 2 and 3 but not 4, you would need to define it in both places.  When using this additional parameter, it would be dependent upon the user to keep these ids used to link variable lists and scans consistent/coordinated.

Maybe the right answer is a mix of the two approaches.  Expose the option to provide an arbitrary external var index, what @hillu calls a context.  If one isn't provided it defaults to the posix thread id.  On the scan calls if the context isn't provided, it defaults to using the posix thread compare function else it does a numerical compare of the values or skips all override checks if the value is null.   This should allow either approach to work.

plusvic:
@bmagistro you're right, I missed the detail that you are not actually using the thread ids stored in `external`, I got confused by the unnecessary initialization in `yr_rules_define_local_x_variable`.

At this point the strongest argument against using a per-thread list of variable overrides is @hillu's point 2. That's something I didn't think about when advocating for this solution, but it's a very good point. To fully understand the problem you be familiar with coroutines. Which are basically routines that run independently, with their own execution flow, but not in parallel. In fact they usually run within the same OS thread. Coroutines usually yield the CPU control at certain points in their execution flows to a higher level entity (event loop) which orchestrates all the currently running coroutines, and then regain the CPU control back and keep running at the same point where they yielded before. Even Windows has a little known feature called [fibers](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682402(v=vs.85).aspx) which is basically the same concept: functions that can run "concurrently" by cooperating with each other.

In such a context all coroutines would share the same set of variables for a `YR_RULES` structure, because all coroutines run in the same thread, but conceptually couroutines are independent pieces of code.

So, I think I'm going to use #559 as a base for this feature.

bmagistro:
I haven't had time to look at coroutines or fibers, but with that aspect being a driving factor I'm okay with closing this one as not compatible with other aspects of the project.

I guess I'll have to spend a little more time looking at #559.

plusvic:
A solution for #350 was provided in https://github.com/VirusTotal/yara/commit/be0fc3f52f140d38a6a45db006ddc2b71d4de825.

PR #558, opened 2016-11-11
Windows mutex replaced with CriticalSection

CaldurG:
I replaced Mutex with CriticalSection object. It could speed up scanning a tiny bit on Windows.


hillu:
> I replaced Mutex with CriticalSection object. It could speed up
> scanning a tiny bit on Windows.

Nice. The use of GetLastError() after InitializeCriticalSection() seems
wrong, though. Just return 0.

Cheers,
-Hilko

hillu:
Also, have you done any measurements when performing single threaded and multi-threaded scans?

plusvic:
Yes, [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683472(v=vs.85).aspx) doesn't mention that InitializeCriticalSection() change last error. Returning 0 should be fine.

CaldurG:
I fixed return value and used __try __except around critical section in case of STATUS_NO_MEMORY on WinXP. But compiler doesn't seem to recognize __try :(
And here is a performance test someone did on SO http://stackoverflow.com/questions/800383/what-is-the-difference-between-mutex-and-critical-section 
From the SO answer: "On my system for 1,000,000 uncontended acquires and releases, a mutex takes over one second. A critical section takes ~50 ms for 1,000,000 acquires."

plusvic:
__try and __except are a Microsoft's extension, gcc doesn't support them. I'll simply return 0 and forget about the exception.

PR #559, opened 2016-11-13
Add APIs for defining contexts for local external variables

hillu:
This is an alternative approach to #539 for providing a solution #350. See the commit messge for a detailed description.


stefan-baranoff:
I'm a bit concerned about the move away from Yara's memory management model (arenas and yr_* functions for allocating/freeing) and the use of malloc/free directly. There's nothing directly wrong with this approach but is in conflict with other similar code. Is there a reason you took this approach?

hillu:
* stefan-baranoff:

> I'm a bit concerned about the move away from Yara's memory management
> model (arenas and yr_* functions for allocating/freeing) and the use
> of malloc/free directly. There's nothing directly wrong with this
> approach but is in conflict with other similar code. Is there a reason
> you took this approach?

There's no specific reason. Thank you for bringing this up.

When looking at how the variables are read in rules.c, it had not
occured to me at all that all that memory is backed by an arena --
probably because malloc is actually used for
"EXTERNAL_VARIABLE_TYPE_MALLOC_STRING" types.

I could, of course, allocate the memory for local variable sets
on-demand in rules->arena and overwrite the first entry's type field
when freeing the set. @plusvic, what do you think?

Cheers,
-Hilko

hillu:
@stefan-baranoff I have pushed a new version that allocates local external variable spaces within the arena instead of using `malloc()`.

plusvic:
A solution for #350 was provided in https://github.com/VirusTotal/yara/commit/be0fc3f52f140d38a6a45db006ddc2b71d4de825.

PR #565, opened 2016-11-22
-r safe argument added

CaldurG:
Added argument that allows me to treats compiler warnings as errors. We use yara to scan files with 1 big rule file. Rules can be added to this file by many users and we want to prevent users from breaking the file / hurting scan performance.


plusvic:
Implemented in 00f3e1a80876a638ec9288a455c42c441d5a7c37 without changing the API

PR #567, opened 2016-11-22
Rename _CONCAT to YR_CONCAT_IMPL.

### LINKED ISSUE #564: _CONCAT is a reserved identifier
yara/libyara/include/yara/modules.h defines `_CONCAT` which is a `_Leading_underscore_capital` identifier reserved for use by the compiler and standard library. MSVC's STL actually defines its own `_CONCAT` macro (now in `<type_traits>` in MSVC 2017).

It appears that yara's `CONCAT` is the public interface, while the reserved `_CONCAT` is an implementation detail. Therefore, the fix should be simple: rename `_CONCAT` to anything non-reserved, like `CONCAT_IMPL`.

kobykahane:
`_CONCAT` is a reserved identifier name and conflicts with a macro in the Visual C++ headers (`type_traits` in Visual Studio 2017 RC).

This fixes #564.

### DIFF
--- a/libyara/include/yara/modules.h
+++ b/libyara/include/yara/modules.h
@@ -45,8 +45,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // Concatenation that macro-expands its arguments.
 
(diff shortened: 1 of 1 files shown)


plusvic:
Already fixed in bf32e1e27a75459821bb41285517f245875260ba

PR #572, opened 2016-11-29
Exports

wxsBSD:
Support exports in PE module better.

Parse exports and store them for later use in the exports() function. This is similar to how we do import parsing.

Handle parsing ordinals. The PE specification says every export must have an ordinal and the name is optional (though most DLLs do export by name it is not technically required).

Rename the IMPORTED_FUNCTION structure to be IMPORT_EXPORT_FUNCTION as it is essentially identical (the only difference is exports don't have a "has_ordinal" flag).

Implement an exports_ordinal() function which allows conditions to search for exports by ordinal.

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -486,6 +486,21 @@ Reference
 
     *Example:  pe.exports("CPlApplet")*
 
+.. c:function:: exports(ordinal)
+
+    .. versionadded:: 3.6.0
+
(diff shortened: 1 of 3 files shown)


wxsBSD:
An example DLL which has 5 exports all of which have an empty string name but different ordinals is e9a1487826889db9be298ca57e8c2c7b. It's worth noting that it looks like this same kind of bug was fixed in IDA in August.

PR #589, opened 2016-12-28
Add missing NULL check, causing crash on truncated samples.

wxsBSD:
Add missing NULL pointer check, which was causing crashes when run against corrupt binaries where the GuidAttribute string is not actually in the binary. Only times I've seen this is when the binary has been truncated.


wxsBSD:
I also did a quick sweep over the rest of the dotnet module and this is the only place I missed this check.

PR #607, opened 2017-02-13
Fix endian issues with section name extraction

jacob-baines:
Line **167** in _elf.c_ doesn't use an endian function in order to access _section->offset_. This is problematic when investigating a binary that has a different endianness than the host.

In the diff, I've added yr_##bo##bits##toh around the access to offset. I also updated the ELF unit tests to include tests for a big endian ELF binary. I also added a test to ensure the section name is still properly extracted from the x64 test binary.


hillu:
* Jacob Baines:

> Line 167 in elf.c doesn't use an endian function in order to access
> section->offset. This is problematic when investigating a binary that
> has a different endianness than the host.

Could you also provide one or two small examples that could be used in
the tests?

Cheers,
-Hilko

jacob-baines:
@hillu I did provide a binary for the tests (via blob.h). I also updated the unit tests to test this scenario.

Did you mean you'd like to test this yourself? If so, I'm happy to provide extra binaries. Attached you'll find a zip called "test.zip". It contains three items:

* **small-mips** - a big endian executable
* **small-x64** - a little endian executable
* **match_section_name.yar** - a yara rule

[test.zip](https://github.com/VirusTotal/yara/files/772632/test.zip)

The rule is fairly simple. It searches for _".text"_ in the section names.

```sh
import "elf"

rule match_section_name
{
    condition:
        for any i in (0..elf.number_of_sections):
        (
            elf.sections[i].name == ".text"
        )
}
```

Quick examination of both binaries with _readelf_ shows that they each have a section named _".text"_:

```sh
albinolobster@ubuntu:~/test$ readelf -S ./small-mips | grep text
  [ 9] .text             PROGBITS        00400600 000600 0002f0 00  AX  0   0 16
albinolobster@ubuntu:~/test$ readelf -S ./small-x64 | grep text
  [ 8] .text             PROGBITS         0000000000400400  00000400
```

When scanning these files with the _section_name_big_endian_ branch I get the following results:

```sh
albinolobster@ubuntu:~/yara$ ./yara ~/test/match_section_name.yar ~/test/
match_section_name /home/albinolobster/test//small-x64
match_section_name /home/albinolobster/test//small-mips
albinolobster@ubuntu:~/yara$
``` 

However, scanning with the _master_ branch I get:

```sh
albinolobster@ubuntu:~/yara$ ./yara ~/test/match_section_name.yar ~/test/
match_section_name /home/albinolobster/test//small-x64
albinolobster@ubuntu:~/yara$ 
```

Issue #570, opened 2016-11-24
Rules namespace not mentioned in API documentation 


noxdafox:
Greetings,

I am integrating Yara C API within a project and I would like to report the rule namespace when I encounter a match. 

The Python APIs include the namespace in the matching callback. 

Looking at their implementation, I notice they refer to a field of the YR_RULE struct:
https://github.com/VirusTotal/yara/blob/master/libyara/include/yara/types.h#L242

Yet such field is not mentioned in the documentation:
http://yara.readthedocs.io/en/v3.5.0/capi.html#c.YR_RULE

Is this a documentation issue or that field is not supposed to be accessed (example: it's under work and might change at any point in time)?

noxdafox:
I there anyone whoc can answer?

wxsBSD:
Given that the python module uses it, I think it is safe to use in your program. Victor has always been good about giving advance notice when structures are changing.

noxdafox:
I can provide a patch for the documentation if needed. I'd like the OK from the developers though to be sure not to document internal logic or non exposed APIs.

wxsBSD:
I would say it isn't internal, given that yara.c uses it.

plusvic:
@noxdafox I've merged your PR. The namespace is part of the public attributes of a rule, it just slipped through the cracks while writing the documentation.

PR #616, opened 2017-03-01
Document YR_NAMESPACE structure in C API

noxdafox:
It is referenced in Python documentation but not mentioned in the C
API one.

Signed-off-by: Matteo Cafasso <noxdafox@gmail.com>


noxdafox:
This addresses issue #570

PR #618, opened 2017-03-08
Fix ELF module issues on big endian system

jacob-baines:
I created a build of YARA for a mips (big endian) system. Included in that build were the slightly more expanded unit tests for the ELF module that I included in #610. A couple of the tests were failing. There turned out to be two reasons:

1. A handful of places were using raw access of the ELF header values instead of first calling the endian.h functions.

2. Access to section name offset was using yr_##bo##bits##toh but this field is always 32 bits.

Changes were simple. All tests now pass.

### DIFF
--- a/libyara/exefiles.c
+++ b/libyara/exefiles.c
@@ -157,11 +157,12 @@ static uint64_t yr_elf_rva_to_offset_32(
 {
   // if the binary is an executable then prefer the program headers to resolve
   // the offset
-  if (elf_header->type == ELF_ET_EXEC)
+  if (yr_le16toh(elf_header->type) == ELF_ET_EXEC)
   {
     int i;
     elf32_program_header_t* program;
-    if (elf_header->ph_offset == 0 || elf_header->ph_entry_count == 0)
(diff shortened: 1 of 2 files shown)


hillu:
leXXtoh(0) is still 0, so all those changes for `if (... == 0)` checks are not really needed. Otherwise, this looks good, though.

Thanks for testing this (so I won't have to deal with it in Debian when I upload the next stable release).

jacob-baines:
@hillu You are obviously correct about the == 0. I added the lexxtoh there more for consistency since people like me make silly mistakes by accidentally leaving the endian accessors off. So its more for consistency of appearance.

I do appreciate that it is useless though. I'm happy to revert those if you'd like.

hillu:
* Jacob Baines:

> @hillu You are obviously correct about the == 0. I added the lexxtoh
> there more for consistency since people like me make silly mistakes by
> accidentally leaving the endian accessors off.

Yeah, I see what you mean. People like me make such mistakes, too. :-)

PR #621, opened 2017-03-15
Use set_sized_string for setting version in dot net module.

plusvic:
The existing code was going a very long path to do the same thing.


plusvic:
@wxsBSD would you mind taking a look at this change? I think set_sized_string can be used for setting the "version" attribute. The only situation where my change behaves different is if the version string contains a NULL character, is that the reason for doing it like that? or is just that you missed set_sized_string?

wxsBSD:
I think I just missed it. The documentation says the version is a UTF-8 encoded null-terminated string of a specified length, so I think this change is the right thing.

plusvic:
Ok. Great!

Issue #632, opened 2017-04-10
YR_TRYCATCH installs an exception handler for entire process, not handling access violations


rspeaker:
The YR_TRYCATCH macro defined in libyara/exception.h installs a vectored exception handler using AddVectoredExceptionHandler.  This exception handler applies to the entire process that is hosting the libyara module and not just the thread that is currently using yara to scan a memory buffer.

The exception handler itself is defined like this:

static LONG CALLBACK exception_handler(
    PEXCEPTION_POINTERS ExceptionInfo)
{
  int tidx = yr_get_tidx();

  switch(ExceptionInfo->ExceptionRecord->ExceptionCode)
  {
    case EXCEPTION_IN_PAGE_ERROR:
    case EXCEPTION_ACCESS_VIOLATION:
      if (tidx != -1 && exc_jmp_buf[tidx] != NULL)
        longjmp(*exc_jmp_buf[tidx], 1);

      assert(FALSE);  // We should not reach this point.
  }

  return EXCEPTION_CONTINUE_SEARCH;
}

The tidx variable will be set to -1 whenever yara is not currently scanning a memory buffer.   This means that if some other thread that is not currently processing a yara request triggers an EXCEPTION_IN_PAGE_ERROR or EXCEPTION_ACCESS_VIOLATION, this exception_handler will hit the assert(FALSE) block which in debug mode will result in immediate termination even if there is another exception handler present.

A secondary problem with this is that because tidx will be set during the entirety of yr_rules_scan_mem_blocks, and that function makes callbacks that are external to the library

e.g.:
callback(CALLBACK_MSG_SCAN_FINISHED, NULL, user_data);

This means that any EXCEPTION_IN_PAGE_ERROR/EXCEPTION_ACCESS_VIOLATION exception that is triggered by the *callback* will be silently ignored as well.  This can cause problems if the callback was trying to handle its own exceptions.

Its not immediately obvious to me why a process-wide vectored exception handler is needed as opposed to just a normal try/catch to handle the paging exceptions.

plusvic:
@hillu, I think @rspeaker is right, and this issue applies to Linux too. Any thoughts?

hillu:
[...]

PR #639, opened 2017-04-14
Implement YR_TRYCATCH in terms of __try/__catch when using the Microsoft compiler

hillu:
This fixes part of #632

### DIFF
--- a/libyara/exception.h
+++ b/libyara/exception.h
@@ -57,7 +57,27 @@ static LONG CALLBACK exception_handler(
   return EXCEPTION_CONTINUE_SEARCH;
 }
 
-#define YR_TRYCATCH(_do_,_try_clause_, _catch_clause_)                  \
+#ifdef _MSC_VER
+
+#include <excpt.h>
+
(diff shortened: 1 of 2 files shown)


libyara/exception.h
```diff
@@ -57,7 +57,27 @@ static LONG CALLBACK exception_handler(
...
+#include <excpt.h>
+
+#define YR_TRYCATCH(_do_,_try_clause_,_catch_clause_)           \
+  do                                                            \
+  {                                                             \
+    if (_do_)                                                   \
+    {                                                           \
+      __try                                                     \
+      { _try_clause_ }                                          \
+      __except(exception_handler(GetExceptionInformation))      \
```
rspeaker: I believe this needs to be "GetExceptionInformation()" since GetExceptionInformation is a function that you need to invoke in order to get the _EXCEPTION_POINTERS*

After making that change I can confirm the code compiles on VS2015.  Will look at testing the actual exception handling later today.

hillu: of course it needs to be a function call.

PR #643, opened 2017-04-21
YR_TRYCATCH: Don't abort process if unrelated code triggers an exception while scanning. (Win32)

hillu:
As described in #632


plusvic:
What about POSIX systems? I mean, I believe this issue also applies to POSIX. We are setting a signal handler for SIGBUS and SIGSEGV during the duration of the scanning, if another thread causes such a signal our handler is going to be called and the assertion will fail. Am I right?

hillu:
Right. We should get rid of the `assert(false)` there, too.

hillu:
Ideally, our `exception_handler` function should invoke a signal handler that the application may already have installed (or invoke the default action for the signal), but I am not quite sure how to do that.

Issue #650, opened 2017-05-05
Multiple bugs in .NET parsing module


flandr:
Several parsing bugs exist in the .NET module. Thus far:

1. Malformed stream header records w/o null termination are silently accepted, leading to assumption of name length == 32 and garbage reads of remaining stream header records

2. Tilde (`#~`) parsing assumes that `#Strings` and `#Blob` sections were found, leading to null pointer dereference if this is not the case.

3. Large blob (4-byte blob length encoding) parsing has an off-by-one error on the blob size descriptor, leading to erroneous advancement of the offset for later structures.

This list is probably not comprehensive. In general, sanity checking reserved fields and version numbers as well as computed offsets early could ameliorate some of these problems.

wxsBSD:
Can you share samples which trigger these bugs? I'm happy to address them if I can reproduce them.

wxsBSD:
Should probably clarify that I can fix the second one without a sample, but having one which triggers it would help.

flandr:
Sure, I emailed you a sample that exercises 1 & 2. I'll send a pull request for 3 in a moment.

wxsBSD:
I think first one you bring up (stream headers without null termination are silently accepted) is within what I would consider a implementation. The spec says they must be null-terminated and are limited to 32 characters. With no null terminator I have no other option but to take all 32 bytes.

As for issue 2, I'm putting up a pull request right now for those. Thanks again for bringing these to my attention. Very much appreciate them.

plusvic:
I guess this issue can be closed as all points have been addressed. Right?

wxsBSD:
I believe this can be closed. Thank you for bringing it up!

flandr:
Sorry for my delayed response; I was busy this weekend.

> I think first one you bring up (stream headers without null termination are silently accepted) is within what I would consider a correct implementation. The spec says they must be null-terminated and are limited to 32 characters. With no null terminator I have no other option but to take all 32 bytes.

The spec is poorly worded but I disagree:

> Name of the stream as null-terminated variable length array
of ASCII characters, padded to the next 4-byte boundary
with \0 characters. The name is limited to 32 characters.

The primary property of the stream name array is that it is null-terminated; the fact that it is limited to 32 characters is a secondary property. When parsing, the absence of a null character within the 32 byte array should mean that the stream header is malformed.

There are other sanity checks that could be useful when validating these variable-length headers, e.g.
 
[...]

PR #651, opened 2017-05-05
Fix variable length blob size offset

flandr:
Refs #650


wxsBSD:
Sigh, good catch. Hopefully @plusvic merges before he tags 3.6.0!

PR #652, opened 2017-05-05
Fix dereference of uninitialized pointer

flandr:
One of these things is not like the others...

Refs #650


wxsBSD:
Sigh... Thank you @flandr!

flandr:
Misread `set_integer`; this is just a dereference of uninitialized pointer. Still a bug, but less severe.

PR #653, opened 2017-05-06
Handle malformed streams better.

wxsBSD:
When parsing the #~ stream it will require references to the #US and #Blob
streams, so we need to make sure those are not NULL before attempting to parse
the stream. It is possible to only do these checks in the place we need them but
I'd rather declare the whole thing as malformed once and early.

Fixes part of #650.


wxsBSD:
@flandr - Let me know if this is roughly what you had in mind. I decided to do the check once, early in the process, instead of sprinkling it around at all the places inside the function which handles that stream, which seems more sane to me. We should bail early if we can...

flandr:
LGTM

PR #675, opened 2017-06-01
Increase RE_MAX_SPLIT_ID to allow complex rules.

cblichmann:
Some of our internal rules are rather complex and don't work with the default setting of 128.


plusvic:
Are you getting the error with latest version? Notice that version 3.5.0 was wasteful with split IDs, but that was improved in version 3.6.0.

cblichmann:
Indeed, the 128 default seems to be enough now. Closing.

PR #691, opened 2017-06-28
Fix issue of opening file with reparse point on Windows

lgsonic:
There is error using CreateFile to open file with reparse point on Windows. For instance, CreateFile returns INVALID_HANDLE_VALUE when trying to opening the symbolic link file whose target does not exist .

If FILE_FLAG_OPEN_REPARSE_POINT flag is specified, the issue is resolved because CreateFile will attempt to open the symbolic link itself instead of opening the target, 

Refer to MSDN on CreateFile and reparse point:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365505(v=vs.85).aspx

Following is a real archive file from Linux system. Decompress this file and scan with Yara. Error is reported when scanning the file "proc/net".
[debug-low-memory-SBx908-main-20170310-1-1-1489108193.zip](https://github.com/VirusTotal/yara/files/1110309/debug-low-memory-SBx908-main-20170310-1-1-1489108193.zip)


plusvic:
Why would you want to read the reparse point data instead of the file being pointed to? 

When you have a symbolic link in a file system and you open it, you usually expect to open the linked file. If the link is broken you expect an error. By adding the FILE_FLAG_OPEN_REPARSE_POINT you won't be able to use symbolic links for scanning the linked file anymore.

lgsonic:
In our case, we'd like to skip any symbolic link whose target doesn't exist and won't want to get any error. Perhaps this is a bit special other than normal case. 
Just ignore this issue please.

PR #701, opened 2017-07-04
filenames should be usable in conditions

teeshop:
Certain Malware creates files with specific names. I'd like to be able to detect such files without using external variables.

Example:
```
rule filename_sample {
  condition:
    filename matches /\.exe$/
}
```


plusvic:
YARA can scan a buffer in memory, in those cases a file name won't be available. Passing a file name as an argument to `yr_rules_scan_mem` doesn't look like a good API design, and breaks backward compatibility.

PR #705, opened 2017-07-12
Crash due to memory uninitialized

alvarofe:
The fix might be wrong since I am not familiar with the code so point me out in the right direction. I think the first `reloc_offset` should starts at zero but again not familiar with the format

The issue resides in the function `yr_arena_load_stream` specifically in the following lines

```
961   if (yr_stream_read(&reloc_offset, sizeof(reloc_offset), 1, stream) != 1)
962   {
963     yr_arena_destroy(new_arena);
964     return ERROR_CORRUPT_FILE;
965   }
966
967   while (reloc_offset != 0xFFFFFFFF)
968   {
969     if (reloc_offset > header.size - sizeof(uint8_t*))
970     {
971       yr_arena_destroy(new_arena);
972       return ERROR_CORRUPT_FILE;
973     }
974
975     yr_arena_make_relocatable(new_arena, page->address, reloc_offset, EOL);
976
977     reloc_address = (uint8_t**) (page->address + reloc_offset);
978     reloc_target = *reloc_address;
979
980     if (reloc_target != (uint8_t*) (size_t) 0xFFFABADA)
981       *reloc_address += (size_t) page->address;
982     else
983       *reloc_address = 0;
984
985     if (yr_stream_read(&reloc_offset, sizeof(reloc_offset), 1, stream) != 1)
986     {
987       yr_arena_destroy(new_arena);
988       return ERROR_CORRUPT_FILE;
989     }
990   }
```

`reloc_offset` is read from the file which is under the control of the user. In the first iteration it holds a value of 32, which will make the first 32 bytes of `page->address` to be uninitialized. However, later on the first 32 bytes of `page->address` will be dereferenced at line 401 within `rules.c` file.

```
399   external = rules->externals_list_head;
400
401 while (!EXTERNAL_VARIABLE_IS_NULL(external))
```

Here is the file that produces the issue (just remove .png extension)

yara-crash

```
==2194==ERROR: AddressSanitizer: SEGV on unknown address 0x000000001aa0 (pc 0x000000545baf bp 0x74b15eab3c10 sp 0x74b15eab2f40 T0)
==2194==The signal is caused by a READ memory access.
    #0 0x545bae in yr_rules_scan_mem_blocks /home/alvaro/fuzzers/yara/yara/libyara/rules.c:401:11
    #1 0x54771a in yr_rules_scan_mem /home/alvaro/fuzzers/yara/yara/libyara/rules.c:586:10
    #2 0x54771a in yr_rules_scan_file /home/alvaro/fuzzers/yara/yara/libyara/rules.c:610
    #3 0x50ffb7 in main /home/alvaro/fuzzers/yara/yara/yara.c:1227:14
    #4 0x65a326cbc439 in __libc_start_main (/usr/lib/libc.so.6+0x20439)
    #5 0x41a179 in _start (/home/alvaro/fuzzers/yara/yara/yara+0x41a179)
```

### DIFF
--- a/libyara/arena.c
+++ b/libyara/arena.c
@@ -963,6 +963,10 @@ int yr_arena_load_stream(
     yr_arena_destroy(new_arena);
     return ERROR_CORRUPT_FILE;
   }
+  if (reloc_offset) {
+    yr_arena_destroy(new_arena);
+    return ERROR_CORRUPT_FILE;
+  }
 
   while (reloc_offset != 0xFFFFFFFF)
   {


alvarofe:
The path is wrong because it does not protect in the following iteration of the loop so is incomplete. I will wait until you take a look to come up with a better solution.

plusvic:
The problem with the file you sent is that the `size` field of `ARENA_FILE_HEADER` (see: https://github.com/VirusTotal/yara/blob/master/libyara/arena.c#L58) is incorrect. The current value is 0x5CF8 but it should be 0x5CE8. 

YARA compiled files has the following layout:

<--header--><-- data --><-- relocs -->

As the `size` field in the header is 16 bytes larger than expected the first 4 relocation offsets are considered part of the data and not processed as they should. This means that there are 4 pointers in YARA data structures that weren't properly relocated and they are pointing to wrong data.

The right solution for avoiding this kind of issues is implementing some kind of integrity check, like calculating some hash for the file and append it.

Did you manually changed the header? Or it just happened?

alvarofe:
I discovered this while fuzzing yara.

If at some point the code interprets that there is more data that it should, shouldn't be this propagated throughout the code base to avoid use those relocs that have not been initialized? I've might understand it wrong, I didn't spend enough time with the code.

plusvic:
The program doesn't have a way for knowing that those relocations weren't applied, it relies on the value for `size` to know where the data ends and the relocations begin, if `size` is modified there's little that the program can do to figure out where the relocations actually begins. As I said before the solution for avoiding this kind of issues is implementing an integrity check and reject any file that has been tampered.

If you start fuzzing compiled rules you are going to find a lot of crashes, and that's expected because YARA trusts the content of those files and assume they are correct. For that reason you shouldn't use YARA with a compiled rules file coming from a untrusted source. For using compiled rules coming from the outside world an integrity checking mechanism is a must.

If you want to fuzz YARA you should fuzz rules in source form, not in binary form.

plusvic:
An integrity check has been implemented in 361f77457becd4e18136caf7fc46f5b7d52da778. Now YARA does not load any compiled rules that has been corrupted. Notice however that this isn't a security mechanism, a maliciously crafted file can still crash YARA.

alvarofe:
I think this is a dangerous statement

"_If you start fuzzing compiled rules you are going to find a lot of crashes, and that's expected because YARA trusts the content of those files and assume they are correct. For that reason you shouldn't use YARA with a compiled rules file coming from a untrusted source_"

Given the popularity and nature of Yara, trust is something can't be always assured. Therefore, I think security measures should be taken. Independently, file integrity is a step forward.

plusvic:
I understand your point, but at the end everything depends on which warranties you are willing to make or where you put the walls for your walled garden. YARA guarantees that it will be safe to compile a rule coming from an untrusted source and that you won't be able to crash or exploit the program by using a malformed/malicious source file. That's a very important warranty that YARA can't ignore. 

However, compiled rules are not necessarily in the same case. The ability to write/read compiled rules to/from a file was put there as a convenience feature intended to avoid compiling the same rules over and over again, it was not designed with the idea of sharing compiled rules with untrusted sources. That's why YARA assumes that the compiled rules are correct and it doesn't defend itself from malformed files. To put it in another way, compiled rules are not considered user input, the actual user input is the source file, compiled rules are just an intermediate representation that you can reuse.

If your intention is receiving compiled rules from untrusted sources then additional warranties should be made, but with the current ones that's a very bad idea.

plusvic:
BTW: If I didn't manage to convince you drop me an email and we can discuss it at length in Spanish ;)

alvarofe:
As long as there is not decompiler for yara rules (correct me if I am wrong) there would be an scenario where the shared rules could be compiled ones instead of sources files due to commercial reasons (I don't know if this a common scenario or if it does make sense at all tbh). But given that scenario,  an user unwittingly could get tricked and make use of a compiled yara which could lead to a crash/exploit yara somehow. 

I agree with you that after all this is about design and what you can guarantee. However, about _don't run untrusted compiled rules_, is like the trust with TLS and the certificates where there have been examples where were totally broken for different reasons. 

What I am raising here is a shifting on the rationale, in order to provide the same strength against compiled rules. If this project wouldn't be used or popular I wouldn't mind, but the reality is quite the opposite. 

Independently is up to you ;) I will report and try to fix them as I find them so we can discuss it one by one. After all I am neither familiar with the code nor YARA's design.

plusvic:
I agree on fixing any bug that may cause a buffer overflow, but that's the only guarantee that it can be made when loading compiled rules from untrusted sources. Guaranteeing that YARA won't crash in other forms (like invalid pointer dereferences) is imposible due to the way in which compiled rules are saved. A malicious attacker will always be able to craft a compiled rules file that would make the program crash.

Distributing rules in compiled form makes sense in certain escenarios, like for example vendors selling appliances or products integrating YARA and providing the rules for them, but they don't want the users to have access to the rules. That's a legit use case, but in those cases the vendors should be implementing a digital signature mechanism in which they can send the rules signed to the appliance/product, so the product is able to verify that the rules are authentic before loading them.

PR #708, opened 2017-07-14
Json module

CaldurG:
I created module for matching json data. It exports function "has_key(str)" and string dictionary "keys". So for json {"hello": "world"} you can just write rule:

import "json"

rule test {
conditon:
json.has_key("hello") and json.keys["hello"] == "world"
}


metthal:
Pardon me my question, but I wonder, what is the use case for this?

CaldurG:
We are using yara not only to search in executables, but also in malware configs obtained from honeypots and in feed of newly registered certificates. These are all json formatted data so I assumed that maybe this module could also help others.

cneberg:
Would you consider adding some json list operations?    like "len" for lengths of json lists and "contains"  for whether a certain string is a json list?

plusvic:
This module looks interesting, but I've finally decided not to merge it.

PR #714, opened 2017-07-24
Leet

CaldurG:
Added "leet" keyword (just like nocase, wide). It matches strings with common substitutions of numbers and letters. So string "$="Microsoft" leet" matches for example "M1cros0ft".


plusvic:
Closing this PR as it doesn't look that there's too much demand for this feature.

PR #727, opened 2017-08-28
Callback on include

edhoedt:
Adds the possibility to fetch 'include' rules files from external data sources (eg: database or remote server) by allowing developers to set a custom callback for the 'include' directive. 
This update allows developers to create interfaces between yara and other services without requiring write access on the disk to store rules source files.
If the callback function is not set, yara behaves as in previous versions and tries to get the requested file from the disk. If it is set, yara will not try to read the file from disk.

Setting a callback is done via:
`void yr_compiler_set_include_callback(YR_COMPILER* compiler, YR_COMPILER_INCLUDE_CALLBACK_FUNC callback, void* user_data)
`

The callback itself must have the following signature:
```
const char* include_callback(
    const char* include_name,
    const char* calling_rule_filename,
    const char* calling_rule_namespace,
    void* user_data);
```
Parameters
  * ``include_name``: name of the requested file as stated by the 'import' directive.
  * ``calling_rule_filename``: the requesting file name (NULL if not a file, ie: compiling from string).
  * ``calling_rule_namespace``: namespace (NULL if undefined).
  * ``user_data``: additional user data.

Returns:
  * The content of the file to include as a single string


edhoedt:
Any comments @plusvic ? Anything I should improve to have it merged?

edhoedt:
Thanks for the feedback, I believe I fixed everything accordingly. Could you please tell me if you see any other issues? If everything is OK, I will proceed with the corresponding fixes in yara-python.

edhoedt:
@plusvic I followed your suggestion but I changed the free function signature a little bit:
` void include_free_func (char* callback_return_ptr, void* user_data)`
*callback_return_ptr* is a pointer to the value returned by the callback. *user_data* is the same as passed to the callback.
I also updated yara-python: https://github.com/VirusTotal/yara-python/pull/58

adulau:
The patch is quite neat and nifty for https://github.com/CIRCL/yara-validator. Will you plan to merge it soon? Thank you.

libyara/compiler.c
```diff
@@ -182,38 +183,119 @@ YR_API void yr_compiler_set_callback(
...
+
+    fseek(fh, 0, SEEK_END);
+    file_length = ftell(fh);
+    fseek(fh, 0, SEEK_SET);
+    file_buffer = (char*) yr_malloc(file_length+1);
+    if(file_buffer)
+    {
+      if(file_length != fread(file_buffer, 1, file_length, fh))
+      {
+        return NULL;
```
plusvic: There's a memory leak here. If the file can't be fully read for some reason `file_buffer` is never freed. Add a `yr_free(file_buffer)` before the `return` statement. A call to `fclose` is also required to prevent leaking file handles.

libyara/lexer.c
```diff
@@ -1439,19 +1384,42 @@ YY_RULE_SETUP
...
+      if(compiler->include_callback == _yr_compiler_default_include_callback)
+      {
+        base_err_msg = "can't open include file: %s";
+      }
+      else
+      {
+        base_err_msg = "callback failed to provide include resource: %s";
+      }
+      size_t err_buff_len = sizeof(base_err_msg) + sizeof(yyextra->lex_buf);
+      char* error_buffer = (char*) yr_malloc(err_buff_len);
```
plusvic: There's a memory leak here, `error_buffer` is never freed.

plusvic: `sizeof(base_err_msg)` returns the pointer's size, not the buffer's size.

plusvic:
Already merged in c0fa0bf760fe2fe82a7216daed34feb2c4641307

PR #744, opened 2017-10-01
Fix the build on 10.13. Missing LDFLAGS for OpenSSL from brew.

wxsBSD:
The build is broken on OS X 10.13 because they finally removed the OpenSSL libraries they had been shipping. They stopped shipping the headers back in 10.11 but have now stopped shipping the libraries. This diff adds the libraries from brew to LDFLAGS. We were already adding the headers from brew to CFLAGS for a while now.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,7 +32,8 @@ case $host_os in
     darwin*) CFLAGS="$CFLAGS -I/opt/local/include"
              # Starting with Mac OS X 10.11 (El Capitan) the OpenSSL headers
              # are in /usr/local/opt/openssl/include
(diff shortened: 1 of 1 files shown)


wxsBSD:
The failure presents as:

`
checking for openssl/safestack.h... yes
checking for MD5_Init in -lcrypto... no
checking for MD5_Update in -lcrypto... no
checking for MD5_Final in -lcrypto... no
checking for SHA256_Init in -lcrypto... no
checking for SHA256_Update in -lcrypto... no
checking for SHA256_Final in -lcrypto... no
`

Applying this patch and doing the usual `autoreconf --force --install` will fix it.

PR #750, opened 2017-10-06
Yara elf hardening

flandr:
Fixes several more errors in ELF parsing where various references were not validated.


flandr:
Years of GCC use have lulled me into a false sense of void*-arithmetic security 😢

PR #787, opened 2017-11-08
Improve test_process_scan

hillu:
- The shell may exec() the last command, but we won't find our test
  string in the sleep process...
- Kill shell by sending a signal after scanning


hillu:
(This seems to have been the real issue for failures in my FreeBSD code.)

PR #792, opened 2017-11-14
Exports regexp

wxsBSD:
Per request from @Neo23x0, allow searching for PE exports by regular expression.


wxsBSD:
I'm assuming this, if accepted, will go in 3.7.1 whenever that happens. Please let me know if I should update the docs!

PR #798, opened 2017-11-26
Added parsing of unoptimized metadata stream in dotnet module

metthal:
This pull request adds a support for unoptimized metadata stream in dotnet module. @wxsBSD initially added a comment that he saw such a sample, but chosen no to parse it because it is undocumented in ECMA-335. I tried to look it up on the Internet and found a book .NET IL Assembler which describes this stream as **unoptimized (uncompressed)**. It has completely the same structure as optimized one, but may contain some additional tables which does not affect us at all.


wxsBSD:
I haven’t tested this, but if your description is indeed accurate then this looks good to me. Thanks for the contribution!

PR #804, opened 2017-12-07
modules/pe: Fixed possible invalid memory read when parsing certificates

metthal:
Attribute Length in `WIN_CERTIFICATE` structure is length of `bCertificate` buffer,
not the length of the whole structure. Every time we want to work with the `bCertificate` buffer,
we need to subtract size of the header. If the security directory is the
very last thing in the PE file, we can possibly end up with invalid pointer.

### DIFF
--- a/libyara/include/yara/pe.h
+++ b/libyara/include/yara/pe.h
@@ -472,6 +472,8 @@ typedef struct _VERSION_INFO {
 #define WIN_CERT_TYPE_RESERVED_1       0x0003
 #define WIN_CERT_TYPE_TS_STACK_SIGNED  0x0004
 
+#define WIN_CERTIFICATE_HEADER_SIZE    8
+
 typedef struct _WIN_CERTIFICATE {
     DWORD Length;
     WORD  Revision;
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -1134,14 +1134,14 @@ void pe_parse_certificates(
         yr_le16toh(win_cert->CertificateType) != WIN_CERT_TYPE_PKCS_SIGNED_DATA)
     {
       uintptr_t end = (uintptr_t)
-          ((uint8_t *) win_cert) + yr_le32toh(win_cert->Length);
+          ((uint8_t *) win_cert) + yr_le32toh(win_cert->Length) - WIN_CERTIFICATE_HEADER_SIZE;
 
       win_cert = (PWIN_CERTIFICATE) (end + (end % 8));
       continue;
     }
 
     cert_bio = BIO_new_mem_buf(
-        win_cert->Certificate, yr_le32toh(win_cert->Length));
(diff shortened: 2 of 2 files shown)


wxsBSD:
Do you have a sample which triggers this? There's a comment right near this code which I wrote which says the length is actually for the entire structure, including the certificate itself, which is contradictory to the documentation but matches every sample I've seen. I ran approximately 700 signed binaries through this code and could not hit the condition which would trigger your patch, so it's hard to say if your patch is right or not.

metthal:
Which documentation do you refer to? Because _Windows Authenticode Portable Executable Signature Format_ [(http://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx)](http://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx) which states:

> **dwLength** is set to the length of **bCertificate**

Or do you mean length as a data directory length?

I do not have sample which triggers this BUT this is the main reason why I think this pull request should be accepted:

It passes either unallocated memory to OpenSSL (if security directory is the last thing in the file) or memory that does not belong to security directory at all (if there is some overlay after security dir) and relies on the fact that PKCS#7 is properly encoded. OpenSSL basically obtains wrong size of the buffer and you can't be sure what exactly it is going to do with it. I may be a little bit paranoid here, but we cannot be sure OpenSSL is not going to access it. I am probably not the first who opened this code and saw that the wrong size is being passed around wondering why. It may not trigger the bug on 99% of files you analyze, but there still may be that 1% of files where this would cause troubles and in that case, I would rather be safe than sorry.

wxsBSD:
I'm reading the same documentation and I think the documentation is wrong. The length is actually the length of the entry in the directory, not the data directory length as a whole.

d65071a494726473c6f530e06fe9e211 is an interesting sample to examine because it has multiple entries in the directory. In that sample the certificate directory starts at offset 0x150ae0, and the first entry has a length of 0x2200. If we treat that length as the length of the entire WIN_CERTIFICATE structure (ie: the entire entry, including padding) it should end at offset 0x152ce0 (0x150ae0 + 0x2200). If we look at offset 0x152ce0 we see the start of another WIN_CERTIFICATE structure.

I'm not opposed to hardening this if it is indeed incorrect memory access, I just want to be sure we aren't going to lower fidelity of the parser to fix a bug which may not actually exist.

metthal:
Sorry for misunderstanding, mea culpa. I have been reading that line I quoted wrong the whole time. You are right, it is length of the whole record. I have also fixed the calculation of next signature in that while loop. That `- WIN_CERTIFICATE_HEADER_SIZE` was meant only for calling `BIO_new_mem_buf`. I have corrected my mistake and made a new commit.

Thank you for the hash, I will look into it tomorrow. Also thanks for review. :)

wxsBSD:
Yeah, I'm in agreement that this is a good fix. It's not likely to trigger any bugs but we should be as strict as possible with what we pass into OpenSSL. Thanks for the fix!

Issue #809, opened 2018-01-10
error in opening file when including file with CRLF line endings


jakubkroustek:
Hi guys,

I encountered a strange error on a Windows machine. When I include a Yara file with Windows line endings (CR LF), Yara scanner ends with a following error:

`$ cat index.yar`
```
include "unix.yar"
include "win.yar"
```

`$ ./yara64.exe index.yar .`
`index.yar(2): error: can't open include file: win.yar`

`$ ./yara64.exe win.yar .`
`(this is ok)`

`$ dos2unix win.yar`
`$ ./yara64.exe win.yar .`
`(this is ok)`

`$ ./yara64.exe --version`
`3.7.0`

Thanks,
Jakub
[crlf_issue.zip](https://github.com/VirusTotal/yara/files/1618686/crlf_issue.zip)

PR #810, opened 2018-01-10
Open files in default include callback with binary mode on Windows

metthal:
Default include callback obtains the file size of included file in order to allocate sufficiently large buffer. It then compares the file size from `_filelength()` with the return value of `read()` call. However, `_filelength()` most probably just read the metadata of the file and returns the amount of bytes in the file. In contrary, `read()` on Windows transforms CRLF line endings to LF line endings when file is open in text mode and thus `read()` returns different value than the actual file size. Opening the file in binary mode should solve this problem.

This fixes bug #809.

### DIFF
--- a/libyara/compiler.c
+++ b/libyara/compiler.c
@@ -84,7 +84,7 @@ const char* _yr_compiler_default_include_callback(
   int fd = -1;
 
   #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


hillu:
This does not fix the issue if yara is built using MinGW (or any other non-MSVC compiler). See my PR #811.

metthal:
You are right. I did not take that into consideration since I am not using MinGW and often forget that it is a thing. I am for merging your PR.

PR #830, opened 2018-02-16
Implement xor modifier.

wxsBSD:
I've implemented an xor modifier, which allows searches for every one byte xor modification to a string. Right now it is only implemented for text strings, but I'm debating on implementing it for hex strings too, though that may be significantly more difficult. Here's some examples of how it would work:

```
// This will match on "Uihr!qsnfs`l!b`oonu" which is "This program cannot" xor'ed with 0x01
rule a {
  strings:
    $a = "This program cannot" xor
  condition:
    $a
}


// This will match on "U\x01i\x01h\x01r\x01!\x01q\x01s\x01n\x01f\x01s\x01`\x01l\x01!\x01b\x01`\x01o\x01o\x01n\x01u\x01" which is "This program cannot" xor'ed with 0x01
rule b {
  strings:
    $a = "This program cannot" xor wide
  condition:
    $a
}

// Match on either ascii or wide single byte xor versions of "This program cannot".
rule c {
  strings:
    $a = "This program cannot" xor ascii wide
  condition:
    $a
}

// A case-insensitive single byte xor search. It generates all the strings first, then applies every xor to each.
rule d {
  strings:
    $a = "This program cannot" xor nocase
  condition:
    $a
}
```


wxsBSD:
Thanks! I’ll write up some docs soon and send those in!

PR #832, opened 2018-02-18
Document the xor modifier.

wxsBSD:
Add documentation for the xor modifier.

### DIFF
--- a/docs/writingrules.rst
+++ b/docs/writingrules.rst
@@ -63,7 +63,7 @@ keywords are reserved and cannot be used as an identifier:
      - uint16be
      - uint32be
      - wide
-     -
+     - xor
      -
 
 Rules are generally composed of two sections: strings definition and condition.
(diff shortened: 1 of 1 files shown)


wxsBSD:
I asked this question to someone who reverses a lot of malware, because I wasn’t sure what the right choice was. The behavior I coded up is based upon his response. His response is, and I’m paraphrasing, that the malware he sees most often for which this would be beneficial are just passing a wide character string through a function which does the single byte xor and is not cognizant of wide character strings at all.

I’m trying to get a hash which shows this, but that might take a bit. In the meantime I’ll update the docs to reflect this interaction more clearly.

wxsBSD:
Would it be better to generate atoms for both ways of doing it? That would cover both possibilities at the expense of more atoms to search for, though I don’t think it would be that bad?

plusvic:
Sounds reasonable. Probably the most common scenario is passing the wide character string to some function that does the xoring as it would be a buffer of bytes.

wxsBSD:
So go ahead and generate all atoms for both ways of doing it? Or is the current method good enough and just document it only?

plusvic:
> Would it be better to generate atoms for both ways of doing it? That would cover both possibilities at the expense of more atoms to search for, though I don’t think it would be that bad?

I don't think so, I would leave it as it is now unless we see some evidence that both ways are equally common.

wxsBSD:
Ok, I’ll update this PR to clarify things as soon as i can.

PR #842, opened 2018-03-19
Add mach_absolute_time()-based stopwatch for macOS

droe:
Unbreak on macOS releases before 10.12 Sierra, lacking clock_gettime(),
by adding an mach_absolute_time()-based implementation of stopwatch.
Prefer the mach_absolute_time()-based implementation to clock_gettime()
for now.  Only check for clock_gettime() in ./configure on non-Mach
platforms.


droe:
The Travis failure seems to be unrelated to the content of this pull request.

PR #854, opened 2018-04-06
Add xordiff module to match XOR differentials with known period

droe:
New module for matching XOR differentials with known period, useful for searching for known plaintext in XOR ciphertext with known key length but unknown key. I'll add documentation if this is of interest to get merged. Usage see included tests.


Agusherman69:
on kan terus sinyal nya

plusvic:
This module breaks YARA's philosophy of scanning the data once in order to be fast. Each time the   `xordiff_match` function is called it iterates over the file's content.

Issue #840, opened 2018-03-15
Bug: If a module has no imports, YARA's PE module reports UNDEFINED instead of 0


ladislav-zezula:
I have a sample EXE file (SHA256: 00131d62f4c27788da7a5e1a324b82aab61fe32442b1ceb649ab76626209db4e) which has 0 imports. This is done by IMAGE_OPTIONAL_HEADER32::DataDirectory[1].VirtualAddress being 0.
```
Microsoft Windows [Version 10.0.16299.192]
(c) 2017 Microsoft Corporation. All rights reserved.

C:\yara-bug2>dumpbin 00131d62f4c27788da7a5e1a324b82aab61fe32442b1ceb649ab76626209db4e /imports
Microsoft (R) COFF/PE Dumper Version 14.00.23918.0
Copyright (C) Microsoft Corporation.  All rights reserved.

Dump of file 00131d62f4c27788da7a5e1a324b82aab61fe32442b1ceb649ab76626209db4e
File Type: EXECUTABLE IMAGE

  Summary

        2000 .rsrc
      4F4000 .text
```
However, YARA's PE module (YARA version 3.7.1) reports UNDEFINED instead of 0:
```
pe
	number_of_signatures = 0
	signatures
	number_of_resources = 2
	resources
		[0]
			offset = 5192864
			length = 4264
			type = 3
			id = 1
			language = 1033
			type_string = UNDEFINED
			name_string = UNDEFINED
			language_string = UNDEFINED
		[1]
			offset = 5197128
			length = 20
			type = 14
			id = 1
			language = 1033
			type_string = UNDEFINED
			name_string = UNDEFINED
			language_string = UNDEFINED
	resource_version
		minor = 0
		major = 4
	resource_timestamp = 0
	number_of_exports = UNDEFINED
	number_of_imports = UNDEFINED
```
Given that 0 imports in a PE module is very suspicious, it would be nice to have way to check that in YARA.

wxsBSD:
I’ll take a look at this today.

tlansec:
Hi,

For what its worth - it looks like this is resolved now and can be closed, I downloaded the bin mentioned in this request and the newest YARA correctly identifies 0 imports.

Cheers,
Tom

PR #898, opened 2018-06-21
PE module extensions:

campt:
Porting forward to 3.7.1 some internal changes that I feel the community might find useful.

ispe - explicit 0 or 1 if the file has a PE header
has_exports - explicit 0 or 1 if the file has a exports table
imports_regex - mirror of the exports_regex for imports

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -883,6 +883,9 @@ IMPORTED_DLL* pe_parse_imports(
 
   PIMAGE_IMPORT_DESCRIPTOR imports;
 
+  /* default to 0 imports until we know there are any */
+  set_integer(0, pe->object, "number_of_imports");
+
   PIMAGE_DATA_DIRECTORY directory = pe_get_directory_entry(
       pe, IMAGE_DIRECTORY_ENTRY_IMPORT);
 
@@ -977,6 +980,9 @@ EXPORT_FUNCTIONS* pe_parse_exports(
 
   if (pe == NULL)
     return NULL;
+
+  /* default to 0 exports until we know there are any */
+  set_integer(0, pe->object, "number_of_exports");
 
   directory = pe_get_directory_entry(
       pe, IMAGE_DIRECTORY_ENTRY_EXPORT);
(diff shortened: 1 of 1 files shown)


wxsBSD:
I think a `imports_regex(dll_regex, function_regex)` would be nice to have. As you have it right now you're only checking if it imports a particular DLL and there's no way to check for functions inside it too. Finding a way to express `imports_regex(NULL, function_regex)` as a way to say "I don't care about the DLL, just regex the function names please" would be awesome too!

libyara/modules/pe.c
```diff
@@ -1340,6 +1340,10 @@ void pe_parse_header(
   uint64_t section_end;
   uint64_t last_section_end;
 
+  PIMAGE_DATA_DIRECTORY directory;
+
+  set_integer(1, pe->object, "ispe");
```
plusvic: Rename to `is_pe` for keeping consistency with `is_dll`.

libyara/modules/pe.c
```diff
@@ -1416,6 +1420,10 @@ void pe_parse_header(
       OptionalHeader(pe, DllCharacteristics),
       pe->object, "dll_characteristics");
 
+  directory = pe_get_directory_entry(pe, IMAGE_DIRECTORY_ENTRY_EXPORT);
+  set_integer((directory && directory->VirtualAddress != 0) ? 1 : 0,
+     pe->object, "has_exports");
```
plusvic: Instead of defining `has_exports` I would prefer fixing issue #840 and using `pe.number_of_exports > 0` for checking if the PE has exports.

plusvic:
> I think a imports_regex(dll_regex, function_regex) would be nice to have. As you have it right now you're only checking if it imports a particular DLL and there's no way to check for functions inside it too. Finding a way to express imports_regex(NULL, function_regex) as a way to say "I don't care about the DLL, just regex the function names please" would be awesome too!

I agree with this too, as is a lot more flexible. You can use it for matching a DLL name, no matter the function, and vice versa. You don't really need to pass a NULL though,  `imports_regex(/.*/,  function_regexp)` and `imports_regex(dll_regexp, /.*/) `

campt:
CI - sure its great...  Anybody got any theories on why yacc would complain about a file not changed by my patch that it didn't complain about in the other 8 builds of the same source....?

Making all in libyara
  YACC     grammar.c
/Users/travis/build/VirusTotal/yara/libyara/grammar.y:115.14-22: syntax error, unexpected string, expecting =

plusvic:
I think this is ready to be merged, I just miss the documentation for the new features.

wesinator:
Thanks for adding these - seems PE module `is_pe` is undocumented ? 
https://yara.readthedocs.io/en/latest/modules/pe.html

r0ny123:
There's a pending PR for this. #1280

Issue #913, opened 2018-07-12
Add printf format attribute to yara_yywarning?


fishilico:
Hi,
While compiling yara with ``gcc -Wsuggest-attribute=format``, the compiler reports the following warning:
```
lexer.l:762:3: warning: function ‘yara_yywarning’ might be a candidate for ‘gnu_printf’
format attribute [-Wsuggest-attribute=format]
```

Would it be acceptable to add this attribute, for example with the following patch?
```diff
--- a/libyara/include/yara/lexer.h
+++ b/libyara/include/yara/lexer.h
@@ -127,7 +127,7 @@ void yyerror(
 void yywarning(
     yyscan_t yyscanner,
     const char *message_fmt,
-    ...);
+    ...) __attribute__((format(printf, 2, 3)));
 
 void yyfatal(
     yyscan_t yyscanner,
```

This will help catching issues in format string usages when ``yara_yywarning`` is called (there are currently no such issues, but they might be introduced in the future).

plusvic:
Good idea, just make sure it will compile with compilers that don't support that.

PR #914, opened 2018-07-12
Add gcc printf format attribute to yara_yywarning

fishilico:
This attribute helps catching issues related to printf-like format string when `yara_yywarning` is called. As it only exists in gcc and clang (not MSVC), add a suitable macro in `utils.h`.

Fixes: https://github.com/VirusTotal/yara/issues/913

### DIFF
--- a/libyara/include/yara/lexer.h
+++ b/libyara/include/yara/lexer.h
@@ -127,7 +127,7 @@ void yyerror(
 void yywarning(
     yyscan_t yyscanner,
     const char *message_fmt,
-    ...);
+    ...) YR_PRINTF_LIKE(2, 3);
 
 void yyfatal(
     yyscan_t yyscanner,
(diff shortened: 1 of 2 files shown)


libyara/include/yara/lexer.h
```diff
@@ -127,7 +127,7 @@ void yyerror(
 void yywarning(
     yyscan_t yyscanner,
     const char *message_fmt,
-    ...);
+    ...) YR_PRINTF(2, 3);
```
plusvic: Something like `YR_PRINTF_LIKE` is more self-explanatory. `YR_PRINTF` suggests that it's a wrapper around `printf` similarly as how `yr_malloc` wraps `malloc`.

fishilico:
Thanks for your review. I have updated my commit.

PR #924, opened 2018-08-02
Use YR_BUILDING_DLL and YR_IMPORTING_DLL for controlling the semantics of YR_API

plusvic:
For compiling YARA in Windows as a DLL you must define YR_BUILDING_DLL. For using it that DLL in your project you must define YR_IMPORTING_DLL. This allows to use the same header files in both cases. If YR_BUILDING_DLL is defined, YR_API translates to __declspec(dllexport), if  YR_IMPORTING_DLL is defined  YR_API translates to __declspec(dllimport).


ApoIIy0n:
why am I getting linker errors when following the current steps? Am I missing something? the linker errors both contain the packages included from nuget (jansson and openssl) they are present and detected by files within the project itself.

PR #936, opened 2018-08-17
Fix bool macro redefinition warning in VS 2013 and higher and a typo

zeffy:
When including the yara headers in a C program or library that has already included `stdbool.h`, you get this compiler warning in Visual Studio 2013 and up, because it does not define `HAVE_STDBOOL_H`:
```
yara\utils.h(45): warning C4005: 'bool': macro redefinition
```

I fixed it by adding a secondary check for `_MSC_VER >= 1800`, which is the earliest version that supports C99's `stdbool.h`.


zeffy:
I also just happened to notice there was a typo(?) in the same file. It had `__attribute_` with one _ at the end, where there should be two __.

PR #940, opened 2018-08-23
Fix bug with multiple signatures.

### LINKED ISSUE #515: Parsing of multiple Authenticode signatures does not work
Portable executable files with multiple signatures are not currently parsed correctly.  This was tested on `msvcr120_clr0400.dll` (SHA256: `d29e93c0fe4f108fa063e1a9692559a4278a0d51ab4feabbb231907dffaeb019`).  In my testing Yara can only ever extract a maximum of one signature.

The underlying reason for this is that multiple signatures are nested.  Extracting additional signatures after the first will require parsing OID 1.3.6.1.4.1.311.2.4.1 (likely because [X.509 only supports one signature](http://www.ietf.org/mail-archive/web/pkix/current/msg33266.html), but this is not an area of expertise).

The current code attempts to iterate `WIN_CERTIFICATE` entries in [pe.c](https://github.com/VirusTotal/yara/blob/eeeb703f1049796274197b4a831da84ccfbf3f63/libyara/modules/pe.c#L972).  This is superfluous as there will only ever be one entry.

The current code also attempts to iterate `SignerInfo` entries in [pe.c](https://github.com/VirusTotal/yara/blob/eeeb703f1049796274197b4a831da84ccfbf3f63/libyara/modules/pe.c#L1018).  From testing I do not believe there can be multiple entries, but would welcome information about files with multiple `SignerInfo` entries.

References:
- https://github.com/SummitRoute/osslsigncode-fork/blob/master/osslsigncode.c#L1166
- http://stackoverflow.com/questions/28777582/reading-certificate-data-from-signed-executables-with-multiple-signatures?rq=1

wxsBSD:
This fixes a bug with files that have multiple signatures. Turns out there are two ways to have multiple signatures.

- Multiple WIN_CERTIFICATE structures. We currently support this.
- A "hidden" nested signature inside the first signature. This patch fixes it so we properly handle this case.

The way it works is it makes the code which converts a PKCS7 structure into YARA module data into a function `_parse_pkcs7()` and then at the very end it checks if there is a "hidden" nested signature and recursively calls itself if one is found. That part of it is on line 1265, the rest of this diff is mostly just a copy/paste of the rest of the code into the recursive function.

I've tested this with the sample mentioned in #515 (d29e93c0fe4f108fa063e1a9692559a4278a0d51ab4feabbb231907dffaeb019) and it does properly parse both of them. I've also verified it by comparing it to what Windows says about the binary and it does indeed match.

Fixes #515.


edeca:
One thought - if you are calling `_parse_pkcs7()` recursively you might like to check the value of `counter` for something sane (e.g. 16) to prevent malicious binaries triggering resource exhaustion.

wxsBSD:
I plan to implement a limit (16 seems reasonable) and may have uncovered a bug in it. I'll hunt it down and update this PR soon.

wxsBSD:
I've uncovered a weird case here which I haven't had the time to fully understand, and may not be able to get to it this week. If anyone wants to take a look here's what I know. `b605b216588fe454b76c51dbb407cc9e71f7b1932367d5557fcc296bd9f28d34` is a PE with 3 signatures (according to Windows). The three serial numbers are `‎30 63 b3 a7 40 c1 cd fd f8 bb 9e 6c 33 1a d7 de`, `‎06` and `‎38 bf a6 1b 82 b8 0f 60 57 15 e4 8a a1 0d e1 53`. This pull request is able to find the first two signatures but not the third one.

I'm hoping to get to this by the end of this week but if someone knows what is going on with this file I'd appreciate a pointer. :)

plusvic:
@wxsBSD is this PR ready to be merged?

wxsBSD:
I’ve got some more work to do on this. Can you give me a couple of days, please? I’ll dedicate some time to it and hopefully have a better PR by Friday morning!

wxsBSD:
Sorry for the force push, but after spending too long fighting with openssl I'm not in the best of moods. ;)

This branch is ready to merge. It now parses the files with nested signatures better (tested on d29e93c0fe4f108fa063e1a9692559a4278a0d51ab4feabbb231907dffaeb019) and will stop parsing at 16 nested signatures. I also moved away from using the BIO interface in OpenSSL in favor of d2i_PKCS7 as it's less cumbersome to work with.

I spent entirely too long trying to get b605b216588fe454b76c51dbb407cc9e71f7b1932367d5557fcc296bd9f28d34 to parse correctly, and I'm giving up on it. As far as I can tell this is not a very common format and was only used by MSFT for a while, but I've been told they are moving away from it in favor of the multiple WIN_CERT format (see https://twitter.com/vathpela/status/1045389079473459200 for details). I'll give a rundown of the file in case anyone wants to pick up where I left off, but I'm not planning on fixing this problem because it's a pain and not likely to be problematic IMO.

The start of the WIN_CERT structure is at 0xe3000. The certificate contained in there has serial `30:63:b3:a7:40:c1:cd:fd:f8:bb:9e:6c:33:1a:d7:de` and has a nested signature (OID 1.3.6.1.4.1.311.2.4.1). This diff allows parsing of the nested signature, which is actually an ASN1 sequence (basically a list?) of signatures. The sequence starts at 0xe48cc, with the first signature at 0xe48e0. This second signature has a serial of `06` and is parsed correctly with this patch. The problem I'm having is finding a way to parse the rest of the sequence. I can only ever get OpenSSL to recognize the first item in the sequence. The third signature (second in the sequence, the missing one I can't parse) starts at 0xe6000. According to Windows this should have a serial of `‎38:bf:a6:1b:82:b8:0f:60:57:15:e4:8a:a1:0d:e1:53`. If anyone can figure out how to get that to parse I'd love to see how, but as I said above I'm frustrated enough with this and am not going to deal with it. :)

wxsBSD:
Thank you to @recvfrom for showing me how to walk the X509 attributes! I'm able to now parse the 3rd signature from the file in question. This PR is now complete and brings YARA more in line with how MSFT is parsing authenticode signatures.

plusvic:
@wxsBSD it looks like this is not compiling in Windows.

```
..\..\..\libyara\modules\pe.c(1128): error C2065: 'signer_info': undeclared identifier [C:\projects\yara\windows\vs2015\libyara\libyara.vcxproj]
```

Maybe the version of OpenSSL we are using in Windows doesn't have the `PKCS7_SIGNER_INFO` type.

wxsBSD:
I’ll take a look tonight.

wxsBSD:
I don't have a copy of VS 2017 handy to test on Windows, but I did manage to confirm the `PKCS7_SIGNER_INFO` structure is declared in the version of OpenSSL used by the Windows build. I did it by installing nuget on my OS X laptop and then fetching the OpenSSL source with `nuget install YARA.OpenSSL.x64 -Version 1.1.0 -ConfigFile windows/vs2017/NuGet.Config` and then checked and the `PKCS7_SIGNER_INFO` structure is declared in `include/openssl/PKCS7.h`. I'm not sure what is going on here...

Is it possible to get better build logs from appveyor? In particular can we get whatever the equivalent of "V=1 make" is on there? Maybe that will point us in a better direction for debugging?

plusvic:
It looks like this PR makes harder a future use of BoringSSL as a replacement of OpenSSL. BoringSSL doesn't implements any of the PKCS7 APIs and they don't plan to do it. So, we either forget about BoringSSL or keep this unfixed :(

wxsBSD:
I’d rather see this fixed than switch to BoringSSL. I think accuracy in parsing is the more important thing here.

recvfrom:
It'd be worth investigating how difficult it'd be to re-implement the functionality in this patch without using the PKCS7 functions... I'll aim to take a look sometime in the next few days unless someone beats me to it.

plusvic:
> It'd be worth investigating how difficult it'd be to re-implement the functionality in this patch without using the PKCS7 functions... I'll aim to take a look sometime in the next few days unless someone beats me to it.

I don't think it's worth the effort, PKCS7 seems to be a very complicated format and parsing it is error-prone.

plusvic:
@wxsBSD I found out why the build is breaking in Appveyor. It turns out that wincrypt.h has the following macro definition which conflicts with the type defined by OpenSSL:

```c
//+-------------------------------------------------------------------------
//  Predefined PKCS #7 data structures that can be encoded / decoded.
//--------------------------------------------------------------------------
#define PKCS7_SIGNER_INFO                   ((LPCSTR) 500)
```
Looking for the most elegant way for fixing this.

plusvic:
It looks like Wine had the same issue, they simply undefined those macros:
https://www.winehq.org/pipermail/wine-patches/2007-August/042510.html

wxsBSD:
Ah, thanks for digging into this and merging it. As I’ve said before, I’m all for removing OpenSSL but I don’t want to sacrifice useful functionality or parsing accuracy.

Issue #943, opened 2018-08-29
Isn't yr_stream_write supposed to return the total size of data written?


jhyeom26:
In yr_arena_save_stream function, v3.8.1

there are validation code like below.
```c
...
if (yr_stream_write(&header, sizeof(header), 1, stream) != 1)
  return ERROR_WRITING_FILE;

if (yr_stream_write(page->address, header.size, 1, stream) != 1)
  return ERROR_WRITING_FILE;
...
```
But, In documentation for v3.8.1
yr_stream_write function should return total size of data written.
but the validation code checks the number of the written element.

Is the documentation wrong? or just validation code mistake?

plusvic:
The documentation is wrong, `yr_stream_write` behaves similarly to [fwrite](http://www.cplusplus.com/reference/cstdio/fwrite/).  Fixed in 97ddc18cb23bafa0717eb594e67022ed5dde036e.

PR #944, opened 2018-08-29
Fix yr_stream_write validation check (#943)

jhyeom26:


plusvic:
The validation was correct, it was the documentation which was wrong. See #943.

PR #957, opened 2018-09-21
Add `icons` array to PE module

jgarman:
Add a new `icons` field to the PE module to retrieve "displayable" icons in the PE file for hashing, matching, or another analysis. Icons are included in this list if they are considered as an 'application icon' by Windows Explorer (for example, when viewing the directory containing the file or displaying the file's properties page).

Also:
* slight fix to the `yr_object_print_data` function to properly escape output of arbitrary binary strings
* add Carbon Black as a company using Yara and myself as a contributor

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,6 +29,7 @@
 Anthony Desnos <adesnos@google.com>
 Christian Blichmann <cblichmann@google.com>
 Hilko Bengen <bengen@hilluzination.de>
+Jason Garman <jason.garman@gmail.com>
 Joachim Metz <joachim.metz@gmail.com>
 Karl Hiramoto <karl.hiramoto@virustotal.com>
 Mike Wiacek <mjwiacek@google.com>
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ awesome list of [YARA-related stuff](https://github.com/InQuest/awesome-yara).
 * [BinaryAlert](https://github.com/airbnb/binaryalert)
 * [Blue Coat](http://www.bluecoat.com/products/malware-analysis-appliance)
 * [Blueliv](http://www.blueliv.com)
+* [Carbon Black](https://www.carbonblack.com)
 * [Conix](http://www.conix.fr)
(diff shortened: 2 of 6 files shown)


metthal:
Hey, great PR, we've found ourselves in the situation where we need exactly this. I just have a one thing that could potentially be improved. What about storing parsed icons in `struct PE` as a list (or anything else) in a similar way how imports and exports are stored there? This would allow us to write additional functions that somehow process the icons (hashing, ...) lazily.

We don't want to compute hash if it's not used in the rules but we also don't want to reparse resources each time the hash computation function is called. Having these icons parsed and stored in `struct PE` would help a lot there.

I know that asking someone else to do something because you need it is not nice so we are also offering our help with providing you a patch that would do it. I would just like to ask you before we start with it, if you are willing to update your PR. We would like to avoid creating another PR since you did the most of the work yourself.

plusvic:
I'm merging this PR, but the `data` field containing the raw data for each icon doesn't convince me. This field encourages people to use conditions like `pe.icons[0].data matches /some regular expression/` which is far from efficient as the icon data can be a relatively large string.

So my question is, what's the primary use case for this feature? Is searching for some pattern inside the icon data? Or is searching for PE files that have a certain number of icons or icons with certain characteristics?

merces:
Hi @plusvic. From my experience the primary case is to match PE files that use **the same icon**, but this is already possible if we use the `hash` module:

```python
import "pe"
import "hash"

rule crime_win_remcos_backdoor : rat
{
	condition:
		hash.sha256(pe.resources[0].offset, pe.resources[0].length) == "587d7c591b7db831f378bf973c5f0e79f532c458e739ffd9e8d4d332c6007526"
}
```
We can also use a loop if we don't know the exact position of the icon.

camptatopenmars:
@plusvic - by way of explanation... the rational since I happened to be the instigator of this patch at CB (though not the author).  I am using yara as a metadata/feature extraction tool/library rather than as a detection engine.  Yara rules provide a nice environment to detect and extract features of interest.  In this case I'm seeking to build a data set of icons used by malware rather than directly detecting malware.  Thus, the raw data field is important.

PR #958, opened 2018-09-27
Update yara cuckoo module to support cuckoo 2.0.6 + format

wroersma:
I updated this module to support the current cuckoo 2.0.6 report.json format as mutex, file_access and registry_access didn't work before. I would be happy to add more fields if that's something that would be of value to the public yara repository. 

Please let me know if you need me to change anything or you want more information from me. I would be happy to provide example report.json files and test yara/files.

Edit: fix yara version listing. I would be happy to some how provide backwards compatibility as well if this is something the repo would like.

### DIFF
--- a/libyara/modules/cuckoo.c
+++ b/libyara/modules/cuckoo.c
@@ -246,6 +246,249 @@ define_function(sync_mutex)
   return_integer(result);
 }
 
+uint64_t summary_parser(
+    YR_SCAN_CONTEXT* context,
+    YR_OBJECT* summary_obj,
+    RE* regexp_value,
+    json_t* value_json)
+{
+  json_t* value;
+
+  uint64_t result = 0;
+  size_t index;
+
+  json_array_foreach(value_json, index, value)
+  {
+    if (yr_re_match(context, regexp_value, json_string_value(value)) > 0)
+    {
+      result = 1;
+      break;
+    }
+  }
+
+  return result;
+}
+
+define_function(summary_registry_key_access)
+{
+  YR_OBJECT* summary_obj = parent();
+  json_t* keys_json = (json_t*) summary_obj->data;
+  json_t* value_json = json_object_get(keys_json, "key_access");
+  return_integer(
+      summary_parser(
+          scan_context(),
+          parent(),
+          regexp_argument(1),
+          value_json));
+}
+
+define_function(summary_filesystem_file_access)
+{
+  YR_OBJECT* summary_obj = parent();
+  json_t* keys_json = (json_t*) summary_obj->data;
+  json_t* value_json = json_object_get(keys_json, "file_access");
+  return_integer(
+      summary_parser(
+          scan_context(),
+          parent(),
+          regexp_argument(1),
+          value_json));
+}
+
+define_function(summary_mutex)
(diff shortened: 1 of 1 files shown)


plusvic:
Backward compatibility is important, as this module is still used with reports generated by previous cuckoo versions. Would you add this changes in backward-compatible manner?

wroersma:
I will get that added in. Is there any list of versions it's intended to support? I couldn't find anything on the wiki but I could just not be able to read :0

wroersma:
So it's updated and retains the old style while adding a new cuckoo.summary path that supports key_access,file_access,file_created,file_recreated,file_failed,mutex,dll_loaded,directory_created,command_line,regkey_opened,file_written,file_deleted,file_exists,regkey_read,directory_enumerated,regkey_written and file_read. If this is something that is useful I'd be happy to add the rest of the options in summary and support things like process list options and more.

metthal:
I'm not the one who makes calls what to merge and what not, but to give you my feedback. I really think that the interface of `cuckoo` module you propose is not in a great shape. With your changes, there would be functions `cuckoo.filesystem.file_access` and then `cuckoo.summary.file_access`. Both express the same thing but they are looking at different JSON keys. Wouldn't it be possible to have a single `cuckoo.filesystem.file_access` which would work with both JSONs from cuckoo v1 and v2 simultaneously? It would also make a sense to move all filesystem related functions to `cuckoo.filesystem` and keep them together (like `cuckoo.summary.file_created`, ...). I understand why you went with `cuckoo.summary`, because that's where they are kept in JSON, but we also have mutexes in `behavior`->`summary`->`mutexes` and they have function `cuckoo.sync.mutex`.

I don't know if you are willing to change the interface, but I would like to ask you to at least think about what I just wrote.

wroersma:
I'm pretty happy to really do about anything that the project would rather
have for sure! One thing is I still don't know what version this was built
on. I should get that up or at least a sample report from it.

For me it was very confusing with the other syntax because it doesn't
follow the path of what it actually is in the report. The actual cuckoo
report itself should be more flat as well and not have so many nested
dictionary objects.

I'll try to play around with seeing if I could somehow also support the
older format as well as using the newer format. I'm not amazing with C so
if I'm doing anything wrong or stupid just let me know. I didn't want to
submit something totally radically different in my first PR to this project.

Please let me know if anyone else has any other thoughts or suggestions!

plusvic:
@wroersma the cuckoo module was written when Cuckoo was still at version 0.1. We expect backward compatibility with that ancient version because at VirusTotal we still have tons of behavior reports in that format. I can share with you some reports in that format if needed. 

I agree with @metthal in keeping a single `cuckoo.filesystem.file_access` that is able to work with all versions of cuckoo reports. The module's naming is not expected to reflect the path of the JSON fields, they are just like namespaces that group functions together according to their functionality:  `filesystem`, `network`, `sync` and `registry`.

wroersma:
Thanks for the information! I'll get it updated tonight and get that pushed
out. It would definitely help if I could get some of the older reports to
test against and ensure everything works. I have access to full file
reports of anything in the past couple years but nothing much older than
that.

PR #975, opened 2018-11-02
Add StaticRelease|x64 config with NDEBUG to libyara.vcxproj

packetzero:
I was building libyara on Windows for osquery and there were two things that I could not work around:
1. Needed NDEBUG=1 defined (compile out asserts)
2. Needed RuntimeLibrary to be MultiThreaded (not DLL) for static


packetzero:
Addresses #978

Issue #979, opened 2018-11-07
*Title Edit* File mapping reading locked file = Crash


Tigzy:
Here's a snapshot of a crash dump I have. It seems at some point the yr_scanner_scan_mem function iterates index for block_data a little bit too much. The array position is pointing in a unavailable memory address and the program crashes. Any direction for me to look at? Do you want the dump and symbols?

![image](https://user-images.githubusercontent.com/9531438/48117934-f8009a80-e26a-11e8-8b40-cf80cc7c1f40.png)

Tigzy:
I cannot reproduce the crash myself, it's a dump uploaded from a client.

Tigzy:
Just to add a little bit of context, and I think this might be useful, at the moment of the crash I have 2 other worker threads: 
- One iterating strings (yr_rule_strings_foreach)
- Another one blocked the EXACT SAME line as the crashing one, so I'm pretty sure it's a thread-safety issue. Another weird thing, the i variable has the same value (even though the block_data (mapped file pointer) is different:

0x000001724f8f0000 + x2600
0x000001724f9e0000 + x2600

![image](https://user-images.githubusercontent.com/9531438/48118529-d4d6ea80-e26c-11e8-9b38-a7f8bf8c26f2.png)

Tigzy:
Up? I've dozens of dumps with the same pattern. 
i grows up until a value where the memory cannot be read. I suspect the boundaries in the block struct are wrong, or badly calculated.

Joseph-Tortorelli:
Wait, so you are getting multiple dumps with the same pattern? Something really weird has recently been going on. People just found malware in some big repositories. I wonder if something is preventing YARA from working. What if someone was able to break YARA and therefore, stop their own malware's signature from uploading? Maybe I am overthinking this but so much weird stuff has just been going on

Tigzy:
> Wait, so you are getting multiple dumps with the same pattern? Like from different people/clients? Something really weird has recently been going on. People just found malware in some big repositories. I wonder if something is preventing YARA from working. Maybe I am overthinking this but so much weird stuff has just been going on

Na I don't think so. It looks really typical of a thread-safety issue. I think there's some shared variable being corrupted and not guarded because the the crash context is very similar everytime, with 2 threads entering the same code area at the same time

Tigzy:
Adding some more context: 
[...]

PR #991, opened 2018-11-27
Guarding file mapped buffer read against page exception

Tigzy:
Context: https://github.com/VirusTotal/yara/issues/979

### DIFF
--- a/libyara/scanner.c
+++ b/libyara/scanner.c
@@ -83,7 +83,14 @@ static int _yr_scanner_scan_mem_block(
       match = match->next;
     }
 
-    index = block_data[i++] + 1;
+    __try
+	{
+		index = block_data[i++] + 1;
+	}
+	__except (GetExceptionCode() == EXCEPTION_IN_PAGE_ERROR ? EXCEPTION_EXECUTE_HANDLER : EXCEPTION_CONTINUE_SEARCH)
+	{
+		return ERROR_READ_FAULT;
+	}
     transition = transition_table[state + index];
 
     while (YR_AC_INVALID_TRANSITION(transition, index))


libyara/scanner.c
```diff
@@ -83,7 +83,14 @@ static int _yr_scanner_scan_mem_block(
       match = match->next;
     }
 
-    index = block_data[i++] + 1;
+    __try
+	{
+		index = block_data[i++] + 1;
```
plusvic: This is not the only point where `block_data` is accessed. The call to `yr_scan_verify_match` can touch that memory too.

Tigzy: Ok thanks. It should be much less critical cause I haven't seen any crash dump so far since the fix I made (local fork). Do you want me to fix or do you want to fix in your own branch?

plusvic: I just remembered that YARA already provides a try-catch (see: https://github.com/VirusTotal/yara/blob/master/libyara/scanner.c#L418). In your code, are you calling yr_scan_xxx with the `SCAN_FLAGS_NO_TRYCATCH` flag?

Tigzy: Nope, I didn't even know about this flag.

Tigzy: I'm afraid that exception handler isn't called in that case...

Tigzy: It does, that weird. However I can confirm I haven't had any single crash dump since the fix, whereas I had at least 5-6 of them every single day. There must be something we don't see

Tigzy: It's hard to "try", since I'd need to retrieve a customer's machine having that very specific driver-locking mechanism (it's tied to hardware AFAIR). 
But yeah I think you found it, EXCEPTION_CONTINUE_SEARCH will try to find a try/catch block in the frames above and probably crash if none is found. I'll let you know if I get the problem again after a new Yara update in our product.

Tigzy: Thanks for looking at it :)

Tigzy: Just so you know I found a crash dump failing for the same reason but this time in yr_get_pe_header. I think it's crashing when accessing the casted PE header members. It should be covered by your fix though.

PR #998, opened 2018-12-06
Fix an OOM condition in dex module.

wxsBSD:
When scanning certain files the dex module will OOM. The file seems to be
obfuscated somehow and it causes the parser to misbehave. I haven't completely
narrowed down what is going on, but this change at least allows the parser to
complete without excessive memory usage and also quickly.

This is easily triggerable with
c8a3172a42cd9270db562b8fd0fddc15d35511483818bfc01c508fdb39f01635.

This change does two things:

1. Move the check for name_idx on both fields and methods to earlier in the
function. If the name_idx is UNDEFINED return early and indicate to the caller
that it should stop parsing that piece because it's not going to work.

2. The parser now support abandoning parsing certain pieces early if the
corresponding function (load_encoded_field or load_encoded_method) returns 0.
The parser would otherwise spend A LOT of time parsing files that are obviously
"wrong."

It does cause some method names (and possibly other things) to be incorrectly
parsed. For example, there's this output now when running the module on the
above sample:

[1]
		class_name = "Landroid/animation/AnimatorSet;"
		name = "le\x00\x11setExistBitmapImg\x00\x0bsetExpanded\x00\x09setExport\x00\x0dsetExportPath\x00\x0fsetExtActionBar\x00\x13setExtendedProperty"
		proto = "LJ"
		direct = 0
		virtual = 1
		method_idx_diff = 17
		access_flags = 1
		code_off = 17
		code_item
				registers_size = 52336
				ins_size = 46768
				outs_size = 24434
				tries_size = 46705
				debug_info_off = 1781676661
				insns_size = 74018772
				insns = UNDEFINED
				padding = UNDEFINED
				tries
				handlers

### DIFF
--- a/libyara/modules/dex.c
+++ b/libyara/modules/dex.c
@@ -374,6 +374,16 @@ uint32_t load_encoded_field(
 
   *previous_field_idx = encoded_field.field_idx_diff + *previous_field_idx;
 
+  int name_idx = (int) get_integer(
+      dex->object, "field_ids[%i].name_idx", *previous_field_idx);
+
+  if (name_idx == (int)(UNDEFINED & 0xFFFFFFFF))
+    return 0;
+
+  #ifdef DEBUG_DEX_MODULE
+  printf("[DEX]\tNAME_IDX 0x%x\n", name_idx);
+  #endif
+
   set_integer(
       encoded_field.field_idx_diff,
       dex->object,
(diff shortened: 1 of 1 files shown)


wxsBSD:
I don't know enough about the dex format to know if this change is even close to correct. I'm hoping the original author (@adesnosgoogle) can provide some guidance.

wxsBSD:
I just ran this PR on a handful of files which are not obfuscated and the results of the parser from before this PR (using `yara -D`) exactly match the results generated with this PR. I'm more confident my change is fine, despite the weirdness mentioned above.

Issue #999, opened 2018-12-08
Compiled rules can execute malicious code regardless of PARANOID_EXEC


bnbdr:
A malicious **compiled** rule file could escape YARA's virtual machine and execute arbitrary code despite the added checks introduced by `PARANOID_EXEC`.

This is possible due to the design of the virtual machine, which stores and operates on pointers to `YR_OBJECT`s from the virtual stack, as well as the following bugs:

- `OP_COUNT` can be used to read a DWORD from any arbitrary address.
- the VM scratch memory is uninitialized and could leak important addresses from the real stack.

I have requested CVE-IDs for theses issues and will update here once they are assigned.

bnbdr:
> using `OP_COUNT` to read the objects canary

Yes. My PoC uses a fake function object and then executes code using `OP_CALL`.  

> more detailed explanation of your exploit would be helpful.

I have written a technical write-up that delves into the VM architecture and explains said PoC. I can publish it shortly.

bnbdr:
@plusvic 
the write-up is here: https://bnbdr.github.io/posts/extracheese/
the PoC is here: https://github.com/bnbdr/swisscheese

plusvic:
Vary good job in finding and explaining this issue. You have done an impressive job in understanding YARA internals. This commit should solve the problem with OP_COUNT: https://github.com/VirusTotal/yara/commit/6acc08d7329413f60e0976be017e18a581450d7a

Regarding the initialization of `vmem`, it's now fully explained in your write-up but I guess your point is that you can read values from the the uninitialized `vmem` that are return addresses left in the stack, and they can be used for determining the base address where the library was loaded. Is that what you meant?

bnbdr:
Thanks.

>...from the the uninitialized vmem that are return addresses left in the stack, and they can be used for determining the base address where the library was loaded 

Yeah, among other things.

plusvic:
#1001 should solve this issue. @bnbdr, your feedback is welcomed.

plusvic:
@bnbdr would you mind reviewing https://github.com/VirusTotal/yara/pull/1061? The idea is storing the canary in a structure in the heap, instead of using a global variable, this way it should be harder to determine the canary's address, at least in theory. Your feedback will be appreciated.

PR #1001, opened 2018-12-13
Fix issue #999

plusvic:


bnbdr:
I would make sure all other opcodes don't suffer similar issues.

plusvic:
`OP_LENGTH` (and `OP_OFFSET` with a bit more difficulty) could be used to read the canary provided that you can write a `YR_MATCH` structure somewhere and get its address in order to push it into the virtual stack. Writing the `YR_MATCH` structure in the virtual stack is possible, but I didn't find a way of knowing its address.

bnbdr:
>...I didn't find a way of knowing its address.

Faking `YR_MATCH` structure is possible because the bytecode can still infer the location of the `vstack` as I showed with the PoC and the function object.

Regardless, I did find the following:

- It's possible to read the canary value using `OP_OFFSET` **without** finding the `vstack` and faking a `YR_MATCH` struct on it.  I can explain further if needed.

- `OP_MATCH_RULE` could theoretically be used to set the LSB of any address.

- The canary value is limited to a 15-bit unsigned value (at least on windows):
```C
// stdlib.h
#define RAND_MAX 0x7fff

```

plusvic:
> It's possible to read the canary value using OP_OFFSET without finding the vstack and faking a YR_MATCH struct on it. I can explain further if needed.

I'm curious about this.

bnbdr:
Basically pushing a shifted pointer to a **real** `YR_OBJECT` before issuing `OP_OFFSET` can cause the VM to read that object's parent's **identifier** and **canary** and put the addition of those on the virtual stack.

Then it's a matter of doing some calculations to reach the original canary value.

I've written a rule ([yarasm](https://github.com/bnbdr/swisscheese/blob/master/op_offset_poc.yarasm), [compiled](https://github.com/bnbdr/swisscheese/blob/master/op_offset_poc.rule)) that compares the canary using both methods (`OP_OFFSET` and `OP_COUNT`). 

It should exit cleanly if they match (i.e. no assertions).

bnbdr:
@plusvic I can elaborate if necessary.

plusvic:
Any additional info would be helpful. I haven't enough time to look into this in detail. However I'm wondering, given the design of the VM, if is going to be actually possible to close all the holes.

bnbdr:
Easiest "patch" would be to break compatibility by requiring a command-line switch to allow running compiled rules together with classifying compiled rules as "potentially dangerous".

This prevents uninformed users from running possibly malicious rules without their intention and shifts the responsibility. 



> additional info would be helpful

I'll see what I can conjure up

plusvic:
Yes, that's probably not vey hurtful to users and at least prevents using
compiled rules inadvertently.

PR #1005, opened 2019-01-07
Fix asterisk bug.

### LINKED ISSUE #1003: Asterisks in comments cause a syntax error
This compiles fine:
```
rule AsteriskBug
{
    strings:
        $my_string = { 12 /* this is fine */ 34 }

    condition:
        all of them
}
```

This doesn't:
```
rule AsteriskBug
{
    strings:
        $my_string = { 12 /* this is *not* fine */ 34 }

    condition:
        all of them
}
```

It says `asteriskbug(4): error: syntax error, unexpected '{', expecting text string`.

wxsBSD:
Fix a bug with asterisk characters in multiline comments in hex strings.

Add a test for this.

Fixes #1003.


plusvic:
Fixing #1003 is going to be harder than it seemed at first, see #1011. The problem is that flex regexp are greedy, and try to match as much as possible.

wxsBSD:
Sigh. Not sure I'll be able to properly fix this anytime soon. I'll see if I can take another shot at it but I've got some other things I want to tackle first. If someone with more knowledge can do it before me I'd love the help! :)

PR #1016, opened 2019-01-18
Fix #1015

### LINKED ISSUE #1015: Only one hex string can contain a comment
```
rule MyRule
{
    strings:
        $a = { 41 42 /* ABC */ 43 }
        $b = { 44 45 /* DEF */ 46 }

    condition:
        any of them

}
```
The above rule file fails to compile:
```
r(6): error: invalid hex string "$a": invalid character in hex string
```
If I remove either of the two comments, or if I move them into the same hex string, then it works fine.

(Unlike my previous issues about comments, this one really is a regression. It was caused by 3620f20.)

josephcsible:


plusvic:
I must admit that I thought at first that this regexp was probably failing at some edge case, but it works well. I had to entere it in https://regex101.com/ and study it carefully to actually understand the underlying idea. Good job @josephcsible! 

However, in the long term I should consider integrating the parsing of hex strings in the main parser, instead of having  a separate one. That would simplify things and avoid hairy regexps which are hard to maintain.

PR #1021, opened 2019-02-06
Add helper methods to dex module.

wxsBSD:
Add the following helper methods to the dex module:

dex.has_method("foo") - Returns true if any method has the literal name "foo".

dex.has_method(/foo/) - Returns true if any method name matches the regex.

dex.has_class("foo") - Returns true if any class has the literal name "foo".

dex.has_class(/foo/) - Returns true if any class name matches the regex.

dex.has_method("foo", "bar") - Returns true if any class has the literal name
"foo" and any method has the literal name "bar".

dex.has_method(/foo/, /bar/) - Returns true if any class name matches the regex
/foo/ and any method name matches the regex /bar/.


libyara/modules/dex/dex.c
```diff
@@ -37,6 +37,165 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+  int i;
+  SIZED_STRING* parsed_name;
+  SIZED_STRING* method_name = sized_string_argument(1);
+  YR_OBJECT* module = module();
+  int64_t number_of_methods = get_integer(module, "number_of_methods");
+
+  if (number_of_methods == UNDEFINED)
+    return_integer(UNDEFINED);
+
+  for (i = 0; i < number_of_methods; i++)
```
plusvic, line 51: We can start using C99 syntax and define the `i` identifier here `for (int i = 0 ..`

plusvic (approved):

PR #1028, opened 2019-02-28
Fix NULL pointer usage

Tigzy:
I've got crash dumps from clients where that function was called with NULL automaton pointer. 

RogueKiller64.exe!yr_ac_automaton_destroy(YR_AC_AUTOMATON * automaton=0x0000000000000000) Ligne 814	C++
RogueKiller64.exe!yr_compiler_destroy(_YR_COMPILER * compiler=0x000001d996a999e0) Ligne 269	C++
RogueKiller64.exe!yr_compiler_create(_YR_COMPILER * * compiler=0x000001d991c6c100) Ligne 246	C++

### DIFF
--- a/libyara/ahocorasick.c
+++ b/libyara/ahocorasick.c
@@ -811,6 +811,7 @@ int yr_ac_automaton_create(
 int yr_ac_automaton_destroy(
     YR_AC_AUTOMATON* automaton)
 {
+  if (automaton == NULL) return ERROR_SUCCESS;
   _yr_ac_state_destroy(automaton->root);
 
   yr_free(automaton->t_table);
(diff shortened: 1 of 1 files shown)


libyara/ahocorasick.c
```diff
@@ -811,6 +811,7 @@ int yr_ac_automaton_create(
 int yr_ac_automaton_destroy(
     YR_AC_AUTOMATON* automaton)
 {
+  if (automaton == NULL) return ERROR_SUCCESS;
```
plusvic: Return `ERROR_INVALID_ARGUMENT` instead.

Tigzy: Done, I was unsure this would have side effect since it's a failure code.

PR #1030, opened 2019-03-01
Guard against divide by 0

Tigzy:
Got a crash dump with QuadPart == 0

RogueKiller64.exe!yr_stopwatch_elapsed_us(_YR_STOPWATCH * sw=0x000002b93cbdbe90) Ligne 51	C++
RogueKiller64.exe!yr_scanner_scan_mem_blocks(YR_SCAN_CONTEXT * scanner=0x000002b93cbdbe30, YR_MEMORY_BLOCK_ITERATOR * iterator=0x000000e10d7fe390) Ligne 484	C++
RogueKiller64.exe!yr_rules_scan_mem(YR_RULES * rules, const unsigned char * buffer, unsigned __int64 buffer_size, int flags=0x00000001, int(*)(int, void *, void *) callback=0x00007ff796357920, void * user_data=0x000000e10d7fe4d0, int timeout=0x00000005) Ligne 255	C++
RogueKiller64.exe!yr_rules_scan_fd(YR_RULES * rules=0x000002b935f32970, void * fd, int flags=0x00000001, int(*)(int, void *, void *) callback=0x00007ff796357920, void * user_data=0x000000e10d7fe4d0, int timeout=0x00000005) Ligne 318	C++

### DIFF
--- a/libyara/stopwatch.c
+++ b/libyara/stopwatch.c
@@ -48,6 +48,7 @@ uint64_t yr_stopwatch_elapsed_us(
 
   QueryPerformanceCounter(&li);
 
+  if (sw->frequency.QuadPart <= 0) return 0;
   return (li.QuadPart - sw->start.QuadPart) * 1000000L / sw->frequency.QuadPart;
 }
 


libyara/stopwatch.c
```diff
@@ -48,6 +48,7 @@ uint64_t yr_stopwatch_elapsed_us(
 
   QueryPerformanceCounter(&li);
 
+  if (sw->frequency.QuadPart <= 0) return 0;
```
plusvic: This is just avoiding the crash, but is not solving the problem because all calls to `yr_stopwatch_elapsed_us` are going to return 0. My guess is that this happens with the 32bit YARA, is that correct? Did you reproduce the issue?

plusvic: From the name `RogueKiller64.exe` I would say that you are running a 64bit YARA, it would be very useful if you look at `sw->frequency` and tell me what you see. `sw->frequency` is a `LARGE_INTEGER` which is something like: 

```C
typedef union _LARGE_INTEGER {  
    struct {
        DWORD LowPart;  
        LONG  HighPart; 
    };
    LONGLONG QuadPart;
} LARGE_INTEGER;
```

It's very weird that QueryPerformanceFrequency returns 0.

Tigzy: Here's full context: 
image

Tigzy: Sorry for the delay, I'll take a look and let you know. 
I just need to accumulate more crash dumps with that bug.

Tigzy: Hello, sorry for the delay, that's the issue I think:
In yr_scanner_scan_mem_blocks (the frame above), I have ret = 1 so I think it exists on line 385 and skips initialization. Is this fixed yet ?

Tigzy: image

Tigzy:
Thanks, I'll look forward for the next release !

PR #1031, opened 2019-03-01
Optimize integer range loops.

wxsBSD:
If a loop can not possible evaluate to true, stop processing it. For example:

for all i in (0..100): (false)
for any i in (0..100): (true)

Both of these loops will stop iterating after the first time through.

Also, loops which use a primary expression are also handled (they are just a
special case of "any" loops).

### DIFF
--- a/libyara/grammar.y
+++ b/libyara/grammar.y
@@ -61,6 +61,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define INTEGER_SET_ENUMERATION   1
 #define INTEGER_SET_RANGE         2
 
+#define FOR_EXPRESSION_ALL 1
+#define FOR_EXPRESSION_ANY 2
+
 #define fail_if_error(e) \
     if (e != ERROR_SUCCESS) \
     { \
@@ -208,6 +211,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 %type <integer> string_modifiers
 
 %type <integer> integer_set
+
+%type <integer> for_expression
 
 %type <integer> rule_modifier
 %type <integer> rule_modifiers
@@ -1126,6 +1131,9 @@ expression
       '(' boolean_expression ')'
       {
         int mem_offset;
+        YR_FIXUP* fixup;
+        void* jmp_destination_addr;
+        uint8_t* nop_addr;
 
         compiler->loop_depth--;
         mem_offset = LOOP_LOCAL_VARS * compiler->loop_depth;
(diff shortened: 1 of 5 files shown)


wxsBSD:
I'm not happy with the fact I have no tests for this, but I couldn't come up with an easy way to test it. I'm thinking of adding in a "debug" option to the VM that will dump the state of memory after it's run so we can use that to write assertions. Thoughts on that?

libyara/grammar.y
```diff
@@ -1538,13 +1641,26 @@ string_enumeration_item
 
 for_expression
     : primary_expression
+      {
+        int mem_offset = LOOP_LOCAL_VARS * compiler->loop_depth;
+        yr_parser_emit_with_arg(yyscanner, OP_CLEAR_M, mem_offset + 4, NULL, NULL);
+        yr_parser_emit_with_arg(yyscanner, OP_ADD_M, mem_offset + 4, NULL, NULL);
+        yr_parser_emit_with_arg(yyscanner, OP_PUSH_M, mem_offset + 4, NULL, NULL);
```
plusvic: Here we have a push that didn't exist before and it should be pulled somewhere, however `for_expression` is used in loops of the form `for_expression _OF_ string_set` for which nothing else has changed. What happens with that extra push in those loops? 

The test cases run ok, but maybe it's because we are not covering that case.

plusvic: I didn't noticed that `OP_ADD_M` pops an item from the stack, so the `OP_PUSH_M` is just to put it in the stack again. So this code doesn't leave an item in the stack as I thought at first. However, it adds unnecessary extra instructions when `for_expression` is used in `for_expression _OF_ string_set` and `for for_expression _OF_ string_set : ` statements. For example:

`4 of ($a, $b)` 

results in ...
```
01.   OP_INIT_RULE
02.   OP_PUSH 4      
03.   OP_CLEAR_M 4
04.   OP_ADD_M 4
05.   OP_PUSH_M 4     
05.   OP_PUSH UNDEF 
06.   OP_PUSH $a   
07.   OP_PUSH $b
08.   OP_OF 
09.   OP_MATCH_RULE
```

Instructions 3, 4, and 5 are not required at all in this case.

My proposal is leaving the `for_expression` production as:

```
for_expression
    : primary_expression
      {
        $$ = FOR_EXPRESSION_ANY;
      }
    | _ALL_
      {
        $$ = FOR_EXPRESSION_ALL;
      }
    | _ANY_
      {
        $$ = FOR_EXPRESSION_ANY;
      }
    ;
```

And emit the required instructions in `_FOR_ for_expression _IDENTIFIER_ _IN_`.

wxsBSD:
I knew those instructions were unnecessary but hadn’t decided to fix them yet. I’ll do that shortly. What do you think about replacing them with a new SET_M which does the equivalent in a single instruction?

plusvic:
> What do you think about replacing them with a new SET_M which does the equivalent in a single instruction?

Makes perfect sense, and it's more efficient.

wxsBSD:
I added OP_SET_M. It doesn't check for setting UNDEFINED though. I noticed OP_ADD_M will not add if the operand is UNDEFINED. That makes sense, but I think setting it to UNDEFINED is acceptable.

plusvic:
@wxsBSD take a look at my latest commits in this branch. If they look good to you we can merge this branch into master.

wxsBSD:
They look good to me. Thanks!

wxsBSD:
Actually, the comments are slightly wrong. I’ll have a fix up in a few hours (stuck on mobile right now).

wxsBSD:
This looks good to me. Thanks for putting these in! :)

PR #1034, opened 2019-03-05
Optimize loops by emitting less instructions

plusvic:
Added instructions JLE_P and JFALSE_P, which are similar to their counterparts JLE and JFALSE, but pop their arguments from the stack instead of leaving the stack unmodified. These new instructions allow generating tighter loops and simplify the code a little bit.

### DIFF
--- a/libyara/exec.c
+++ b/libyara/exec.c
@@ -324,6 +324,18 @@ int yr_execute_code(
         mem[r1.i] = r2.i;
         break;
 
+      case OP_TOP_M:
+        r1.i = *(uint64_t*)(ip);
+        ip += sizeof(uint64_t);
+        #if PARANOID_EXEC
+        ensure_within_mem(r1.i);
+        #endif
+        pop(r2);
+        push(r2);
+        if (!is_undef(r2))
+          mem[r1.i] = r2.i;
+        break;
+
       case OP_SET_M:
         r1.i = *(uint64_t*)(ip);
         ip += sizeof(uint64_t);
@@ -371,6 +383,13 @@ int yr_execute_code(
         ip = jmp_if(r1.i <= r2.i, ip);
         break;
 
+      case OP_JLE_S:
+        pop(r2);
+        pop(r1);
+
+        ip = jmp_if(r1.i <= r2.i, ip);
+        break;
+
       case OP_JTRUE:
         pop(r1);
         push(r1);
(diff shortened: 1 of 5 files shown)


wxsBSD:
I haven't looked at this in detail yet, but I like what I've seen as I skimmed through it.. I'm hoping to get to it in the next day or two.

wxsBSD:
```
        // for all i in (N..M) : (<expr>)
        //
        // 1       PUSH UNDEF  ; "all"
        // 2       PUSH UNDEF  ; "end of list"
        // 3       PUSH N      ; integer range lower bound
        // 4       PUSH M      ; integer range upper bound
        // 5       POP_M 3     ; store range upper bound
        // 6       POP_M 0     ; store range lower bound
        // 7       CLEAR_M 1   ; clear <expr> result accumulator
        // 8       CLEAR_M 2   ; clear loop iteration counter
        // 9    .->INCR_M 2    ; increment loop iteration counter
        //      |  <expr>      ; here goes the code for <expr>, its result will
        //      |                be at the top of the stack
        // 10   |  SET_M 4     ; store boolean expression result in memory 4
        // 11   |  ADD_M 1     ; add boolean_expression result to accumulator
        // 12   |  INCR_M 2    ; increment loop iteration counter
        // 13   |  INCR_M 0    ; increment range lower bound (more like current bound)
        // 14   |  PUSH_M 4    ; boolean expression result
        // 15 .-+--JFALSE_P    ; jump out of loop if last result is false
        // 16 | |  PUSH_M 0    ; lower (current) bound
        // 17 | |  PUSH_M 3    ; upper bound
        // 18 | `--JLE_P       ; jump to start of loop if we haven't iterated enough
        // 19 `--->POP         ; pop end of list
        // 20      SWAPUNDEF 2 ; swap the UNDEF ("all") with loop iteration
        //                       counter (memory 2)
        // 21      PUSH_M 1    ; push the boolean_expression accumulator
        // 22      INT_LE      ; compare boolean_expression accumulator to loop
        //                       iteration counter
```

I believe the instruction on line 12 is never actually emitted and you just left it in the comment by mistake. Other than that, this looks right to me.

Also, as I haven't come up with a good way to make sure the optimizations are working as expected I've resorted to modifying `jmp_if()` in `exec.c` to print out a message if the condition is true or not. It's a cheap way to tell if any jump is taken. ;)

plusvic:
> I believe the instruction on line 12 is never actually emitted and you just left it in the comment by mistake. Other than that, this looks right to me.

Fixed.

> Also, as I haven't come up with a good way to make sure the optimizations are working as expected I've resorted to modifying jmp_if() in exec.c to print out a message if the condition is true or not. It's a cheap way to tell if any jump is taken. ;)

Yes, it's hard to create a test case for that.

PR #1052, opened 2019-04-16
Cmake

citronneur:
This pull request intends to introduce the CMake build system on the YARA project: The primary goal of this PR is to have a better build system for Windows.

Once merged, the `windows` folder can be removed: there is no more need to maintain one SLN file per version, per SDK and per target... This should significantly improve Windows developer life!

We did not change the original build system because both can be used separately.
All files are stored under the "cmake" folder.
CMake can handle the build system, test management, and packaging purposes. See README.md under the "cmake" folder for more details.


metthal:
Hey there. I am working on CMake build system for YARA in my free time too ([link to my fork](https://github.com/metthal/yara/tree/cmake)) and I think there are some things which we can share among your and my work. Here are some notes I made after taking look at your changes:

1) I would suggest not using `add_definitions` but rather [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html). `add_definitions` populates global definitions whereas `target_compile_definitions` is tied to target and then populated transitively according to the dependencies specified by `target_link_libraries`.
2) Same with setting `CMAKE_C_FLAGS`. Prefer [target_compile_options](https://cmake.org/cmake/help/latest/command/target_compile_options.html) which behaves the same as `target_compile_definitions` but for compiler flags.
3) It's also usual to specify [install(EXPORT ...)](https://cmake.org/cmake/help/latest/command/install.html#installing-exports) to generate CMake scripts during installation process which can be used from `find_package(... CONFIG)` in other projects that depends on YARA.
4) There are some things which you omitted from your CMake like possibility to build with jemalloc, tcmalloc or gperf.
5) I would suggest also running `find_package()` for Flex and Bison and verifying their version.
6) There should also be possibility to specify whether `libyara` is built as static or shared library.
7) `target_include_directories` should distinguish between build environment and install environment, see [this](https://github.com/metthal/yara/blob/cmake/libyara/CMakeLists.txt#L137).
8) It's rather unusual to have just single `CMakeLists.txt` and then include other `*.cmake` scripts into it. The more usual approach is to have recursive tree of `CMakeLists.txt` files because then you can directly use `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` to refer to their respective folder in the source-tree/build-tree.
9) `libmagic` is not specified in the list of linked libraries when magic module is enabled.
10) Prefer usage of modern CMake imported targets and rather than using `target_link_libraries(libyara ${OPENSSL_LIBRARIES})` consider using `target_link_libraries(libyara OpenSSL::Crypto OpenSSL::SSL)`
11) Same as 10 also applies to threading libraries, consider using `find_package(Threads REQUIRED)` and then refer to it as `Threads::Threads`.
12) Consider looking for jansson in the system and fallback to ExternalProject if it's not found. It's pretty usual that Linux machines will have it installed from through the package manager of their distribution.
13) Man pages are not installed from your install target.

My personal opinion here is that this is good starting point for CMake build system but it would need a lot of additional work. What I am aiming towards is a bit different though, because I wanted to completely replace build system of YARA with CMake. My CMake build already works with Linux, macOS, Windows MSVC, should be working on BSD but what I still don't support is Cygwin, MSYS2 and other *nix like environments for Windows and I don't have much time to fix it as of right now. Having at least some CMake instead of none is better than waiting for my slow ass to finish it. It's up to you @plusvic which way you want to go.

metthal:
By the way, here's how I managed installation of man pages (You'll also need to `include(FindUnixCommands)` because of `${GZIP}`): https://github.com/metthal/yara/blob/cmake/CMakeLists.txt#L182.

Here are my custom finders for third party libraries:
* [Jansson](https://github.com/metthal/yara/blob/cmake/cmake/FindJansson.cmake)
* [GPerfTools](https://github.com/metthal/yara/blob/cmake/cmake/FindGPerfTools.cmake)
* [Jemalloc](https://github.com/metthal/yara/blob/cmake/cmake/FindJemalloc.cmake)
* [Magic](https://github.com/metthal/yara/blob/cmake/cmake/FindMagic.cmake)

Here's how I link them and also some additional flags I have found out are needed for successful compilation on all systems:
https://github.com/metthal/yara/blob/cmake/libyara/CMakeLists.txt#L167

Feel free to use any of it. I don't have much time right now to dedicate my time to this so I'll be more than happy if it will be useful.

citronneur:
Thanks very much for all of your works and remarks, we really appreciate, and we are always searching to improve our works.
All your remarks are pertinents :
    1: Done
    2: Ok but i wanted to set this flag for all projects
    3: Done (Thanks for the tip)
    4: TBD
    5: TBD (i would like to use your macros)
    6: TBD (already done for another project internally)
    7: Done
    8: I do this to be the lower intrusive as possible (see protobuf)
    9: Done
    10: Done
    11: Done
    12: TBD Same as 5 (i would like to use your macros)
    
For all items mark as TBD, I'm waiting that first pull request will be accepted to work on it.

metthal:
> 8: I do this to be the lower intrusive as possible (see protobuf)

Sounds fair. Thank you very much for taking your time to work on it.

citronneur:
I signed it!

citronneur:
hi,
The error on appveyor is caused by the cygwin build that didn't found flex:
/cygdrive/c/projects/yara/build-aux/missing: line 81: flex: command not found

Is there something to do to fix that?

metthal:
I did it like [this](https://github.com/metthal/yara/blob/cmake/appveyor.yml#L49) in my fork. You should also rewrite travis.yml like [I did](https://github.com/metthal/yara/blob/cmake/.travis.yml). I am still busy and have no time to finish my CMake fork so take it all and use it. I really want this to merged.

citronneur:
OK all is fixed and ready to merge. Thanks to concider this pull request.
Thanks for your work on yara.

Have a nice day!

solvingj:
very much looking forward to building this library with cmake.

shrgabor:
I would love to use yara through CMake, and for me it seems that, all is good.
Why isn't this PR merged already? I mean almost a year has passed by.
Are there any impediments on this?

plusvic:
The reason for not merging this is that maintaining multiple build systems is painful. In top of the standard automake-based build we also support Bazel, so adding CMake is yet another thing to maintain.

metthal:
Thanks for response. I would like to ask you what do you think about getting rid of autoconf/makefile + Visual Studio solutions and let those be generated from CMake? Everyone can generate solution for their own VS version. Linux most probably won't be affected, instead of `./configure` you run `cmake`. Not sure about not-so-mainstream systems but I think there's strong community behind YARA and for example I personally would be glad to help anyone with CMake issues.

citronneur:
Hi,
I'm full agree with @metthal and we can full support this feature.
We can easily drop VIsual Studio solution in place of cmake.

Have a nice day,
Sylvain

plusvic:
I'm cleaning up old PRs. The final decision is continuing with the current `automake`, as migrating to `cmake` would involve a lot of additional work.

PR #1073, opened 2019-06-03
Properly set DEX string size

CalebFenton:
It looks like some code was copy / pasted and the wrong variables were changed for setting dex string sizes. The code now is using the string size as the _index_ into the string table and the value its setting is the offset.

### DIFF
--- a/libyara/modules/dex.c
+++ b/libyara/modules/dex.c
@@ -775,8 +775,8 @@ void dex_parse(
         "string_ids[%i].offset", i);
 
     set_integer(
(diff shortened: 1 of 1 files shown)


CalebFenton:
This was causing some really odd behaviors such as this rule failing:
```rule somehow_fails {
    condition:
        dex.string_ids[0].size == 0x1
        or dex.string_ids[0].size != 0x1
        or dex.string_ids[0].size == dex.string_ids[0].size
}
```

CalebFenton:
The appveyor failure looks like it's for cygwin and it's complaining about a flex command being missing. I don't think it has to do with my change.

PR #1076, opened 2019-06-06
Add crc32 to hash module.

wxsBSD:
Add crc32 functions to the hash module. They work in the normal "offset and
size" way and on arbitrary strings.

While here, add tests for it and also documentation.


wxsBSD:
Sigh, I forgot I switched laptops and had my @fb.com email address in my .gitconfig. Fixed it with a force push. Sorry for that.

wxsBSD:
Not sure if you intend for me to merge these or not but I don't have write access.

PR #1084, opened 2019-06-27
Add dependency on zlib when building with libmagic

metthal:
I ran into this when I was trying to compile yara with almost everything static so it can be portable among multiple Linux machines. Linking against static libmagic.a requires linking against zlib. Since zlib is dependency of libmagic, everyone who has installed libmagic will also have zlib.

I don't really know if this is the right way to solve it since I am not familiar with autotools whatsoever but I am open to any other (better) solutions.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -119,12 +119,14 @@ AC_ARG_ENABLE([magic],
   [AS_HELP_STRING([--enable-magic], [enable magic module])],
   [if test x$enableval = xyes; then
     build_magic_module=true
+    AC_CHECK_LIB(z, zlibVersion,,
+      AC_MSG_ERROR([please install zlib library]))
     AC_CHECK_HEADERS([magic.h],,
       AC_MSG_ERROR([please install libmagic library]))
     AC_CHECK_LIB(magic, magic_open,,
       AC_MSG_ERROR([please install libmagic library]))
     CFLAGS="$CFLAGS -DMAGIC_MODULE"
-    PC_LIBS_PRIVATE="$PC_LIBS_PRIVATE -lmagic"
+    PC_LIBS_PRIVATE="$PC_LIBS_PRIVATE -lmagic -lz"
   fi])
 
 AC_ARG_ENABLE([dotnet],


metthal:
I signed it!

plusvic:
Hi @metthal, I've reverted this change because it was causing trouble when building YARA in Alpine Linux. Can you please provide more information about what specific issue you were trying to fix so that we can come up with a solution that works for all environments?

Apparently your problem was while building a static version of YARA, can you please tell how did you hit that issue?

metthal:
Hi. Basically I was trying to build full portable version of yara for CentOS which wouldn't require installation of any additional packages.

You can use this Docker image to observe the bahvior

```dockerfile
FROM centos:7

RUN yum install -y epel-release && \
	yum install -y autotools cmake3 cmake3-data file-devel file-static \
		gcc libtool make openssl-devel openssl-static unzip wget && \
	yum clean all

RUN mkdir -p /tmp/build && cd /tmp/build && \
	wget https://github.com/akheron/jansson/archive/v2.12.zip && \
	unzip v2.12.zip && \
	cd jansson-2.12 && mkdir build && cd build && \
	cmake3 -DCMAKE_BUILD_TYPE=Release \
		-DJANSSON_EXAMPLES=OFF \
		-DJANSSON_BUILD_DOCS=OFF \
		-DJANSSON_WITHOUT_TESTS=ON \
		.. && \
	cmake3 --build . --target install -- -j && \
	cd / && rm -rf /tmp/build

# Force remove libmagic.so because we only want libmagic.a
# and I didn't find better way to force yara build system
# to use libmagic.a instead of libmagic.so
RUN rm /usr/lib64/libmagic.so

ARG yara_gitref

RUN mkdir /tmp/yara && cd /tmp/yara && \
	wget https://github.com/VirusTotal/yara/archive/${yara_gitref}.zip && \
	unzip ${yara_gitref}.zip && \
	cd yara-${yara_gitref} && \
	./bootstrap.sh && \
	./configure --enable-cuckoo --enable-magic --disable-shared && \
	make -j$(nproc) && \
	make install
```

Now if you run

```
docker build -t yara-centos --build-arg yara_gitref=2bafaf4ec354e57ffdbde8725fbcb6746bb94c63 .
```

you can see that it builds successfully (it is one commit before your revert). Now let's build the commit where you reverted the change

```
docker build -t yara-centos --build-arg yara_gitref=560e9af3b71c59832cdf6b5b8e71ebe0f5913b4a .
```

I've tried to debug the problem back when I first encountered it. The problem is that program which is used during configure time to test whether `libmagic` is installed or not fails to compile because it is missing `-lz` (undefined references to symbols from libz). I didn't know how else to force configure to use `libmagic.a`. Unfortunately libmagic doesn't come with `pkg-config` files so I can't read those to determine what is needed. What would you suggest?

PR #1085, opened 2019-07-01
URL Module

pilate:
Adds a URL module that enables parsing and component matching. Leverages the [curl URL parser](https://daniel.haxx.se/blog/2018/09/09/libcurl-gets-a-url-api/).


plusvic:
This module looks like a very specific use case, where you have to provide additional URLs for each file that you are scanning. Also, it creates additional dependencies from `curl` which makes the maintenance harder in platforms like Windows. So, the burden that it adds out-weights the benefits.

I'm not merging it into the master branch for these reasons.

PR #1090, opened 2019-07-14
Detection of regular expressions with xor option

regeciovad:
Detection of cases like `$re = /a.c/ xor` in which the regular expressions can not be converted into a string (like `/abc/`) and for which the `xor` option is not allowed.

### DIFF
--- a/libyara/grammar.c
+++ b/libyara/grammar.c
@@ -1,8 +1,8 @@
-/* A Bison parser, made by GNU Bison 3.0.5.  */
+/* A Bison parser, made by GNU Bison 3.0.4.  */
 
 /* Bison implementation for Yacc-like parsers in C
 
-   Copyright (C) 1984, 1989-1990, 2000-2015, 2018 Free Software Foundation, Inc.
+   Copyright (C) 1984, 1989-1990, 2000-2015 Free Software Foundation, Inc.
 
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -44,7 +44,7 @@
 #define YYBISON 1
 
 /* Bison version.  */
-#define YYBISON_VERSION "3.0.5"
+#define YYBISON_VERSION "3.0.4"
 
 /* Skeleton name.  */
 #define YYSKELETON_NAME "yacc.c"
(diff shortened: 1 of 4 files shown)


regeciovad:
I signed it!

regeciovad:
I agree and I changed the code to detect all regular expressions with xor.

regeciovad:
Thank you, I will move it to `yr_parser_reduce_string_declaration`.
Just to be sure, the xor with hex strings is already detected with error 11: "error: syntax error, unexpected <xor>, expecting <condition>". 
Should I return this error as well in the case of regular expressions? Or should I change it to error 59: `ERROR_INVALID_XOR_OPTION` with the message like "wrong use of xor option in \"%s\", xor is allowed only with a literal text string"?

plusvic:
Oh, I see. In the case of hex strings the grammar doesn't accept any modifiers. See:
https://github.com/VirusTotal/yara/blob/master/libyara/grammar.y#L578

That's why it already returns a syntax error. 

For coherency I'm leaning towards changing the grammar for preventing the "xor" modifier to be accepted after regular expressions. So instead of accepting the `strings_modifiers` rule in https://github.com/VirusTotal/yara/blob/master/libyara/grammar.y#L566, we could change it to something like `regexp_modifiers` which would work in the same way than `strings_modifiers` but excluding `xor`. With this solution you don't need a new error code and error message, it will return 
`syntax error, unexpected <xor>, expecting <condition>` when using `xor` with a regexp, as it does with hex strings.

metthal:
Regarding the string modifiers, is it even necessary for regular expressions to have modifiers like `fullword` or `nocase`? You can already express those using regular expressions directly (fullword with `\b`, nocase with `i` at the end of the regexp).

wxsBSD:
When I wrote the xor stuff my intention was to not allow it with a regex. I guess I missed that, sorry. As such, I’m in favor of changing the grammar to not allow it.

regeciovad:
> When I wrote the xor stuff my intention was to not allow it with a regex. I guess I missed that, sorry. As such, I’m in favor of changing the grammar to not allow it.

I removed the `xor` option with regexp from the grammar in the current version of the pull request based on a suggestion from @plusvic.

Issue #736, opened 2017-09-11
[PE] Debug information


Tigzy:
Hello, 
Is it possible to add a keyword in PE module to target PDB path?
With possibly regex?
like: 
```
rule my_program
{
    condition:
        pe.pdb.contains("\\path_to_the_program\\the_program.pdb")
}
```

HarsaroopDhillon:
yea its possible with CV_INFO_PDB70/20 :(, will someone do it? probs :D

PR #1092, opened 2019-07-21
Added parsing for PDB string in PE module

cherepanov74:
Implements this feature request: https://github.com/VirusTotal/yara/issues/736

Now it's possible to do following rules:
```
import "pe"

rule pdb_rule
{
   condition:
     pe.pdb contains "substring"
}
```

### DIFF
--- a/libyara/modules/pe.c
+++ b/libyara/modules/pe.c
@@ -266,7 +266,8 @@ static void pe_parse_debug_directory(
 {
   PIMAGE_DATA_DIRECTORY data_dir;
   PIMAGE_DEBUG_DIRECTORY debug_dir;
-  int64_t offset;
+  int64_t debug_dir_offset;
+  int64_t pcv_hdr_offset;
   int i, dcount;
   size_t pdb_str_len;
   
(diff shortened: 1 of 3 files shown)


cherepanov74:
I signed Google CLA

nyx0:
@plusvic Do you have an ETA for merging? thank you :)

plusvic:
> @plusvic Do you have an ETA for merging? thank you :)

Probably for next release, that is being baked right now. I have some comments about the latests changes in this PR though.

libyara/modules/pe.c
```diff
@@ -317,7 +321,7 @@ static void pe_parse_debug_directory(
       if (!struct_fits_in_pe(pe, pdb20, CV_INFO_PDB20))
         break;
       
-      pdb_str_len = strlen((char *)(pdb20->PdbFileName));
+      pdb_str_len = strnlen((char *)(pdb20->PdbFileName), MAX_PATH * sizeof(char));
```
plusvic: This is is an improvement over using `strlen` but it doesn't solve the original issue as you still can read out of the buffer's bounds. You are limiting the number of bytes being read to MAX_PATH, but the bytes remaining in the buffer can still be less than that. So, the argument you you pass to strnlen is the space available from the pdb20->PdbFileName pointer to the end of the buffer. You can use the `available_space` function for computing that.

libyara/modules/pe.c
```diff
@@ -261,6 +261,95 @@ static void pe_parse_rich_signature(
...
+      }
+    }
+    else if (yr_le32toh(cv_hdr->dwSignature) == CVINFO_PDB70_CVSIGNATURE)
+    {
+      PCV_INFO_PDB70 pdb70 = (PCV_INFO_PDB70)cv_hdr;
+      
+      if (!struct_fits_in_pe(pe, pdb70, CV_INFO_PDB70))
+        break;
+      
+      pdb_str_len = strnlen((char *)(pdb70->PdbFileName), MAX_PATH * sizeof(char));
```
plusvic: Same here.

cherepanov74:
Thank you for all your suggestions!

PR #1096, opened 2019-07-28
Private strings

wxsBSD:
Add a "private" modifier for strings which keeps that string out of the matched strings when calling the callback. This allows for rules to use strings which could potentially match a lot of times but are only useful as anchor points for other strings without polluting the output (or logs in our case). Essentially, you can now write rules like https://blog.korelogic.com/blog/2015/08/19/yara-ctf-2 without having A LOT of PK header matches in your output.

Without a private string you have to limit the number of string matches in the callback manually. With a private string it is completely hidden from the callback.

Another upside of this is the python bindings won't have to spend as much time converting the matches to python strings if they are marked private, so in theory there is a performance improvement to be had in some cases here (I haven't measured it but it sounds plausible =b).


wxsBSD:
I signed it!

Though it is only picking up my gmail address which I never use. :(

libyara/include/yara/types.h
```diff
@@ -272,6 +276,7 @@ struct YR_STRING
   int64_t fixed_offset;
 
   YR_MATCHES matches[YR_MAX_THREADS];
+  YR_MATCHES private_matches[YR_MAX_THREADS];
```
plusvic: Is this new field actually required? If the string is private the matches for that string are considered private, but they can be stored in the `matches` field, right?

I guess the intention here was keeping the `matches` field free from private matches so that any existing code doesn't need to change anything for respecting private strings. But in order to keep the code simple I would prefer keeping the matches as they are and simply adding the private flag to the strings. Clients like the command-line tool or yara-python then would decide what to do with private strings, they can hide them by default and show them on request. My only concern with this approach is that we would introduce some inconsistency with respect to private rules, which are completely hidden from libyara clients. Thoughts?

wxsBSD: The first implementation I did required clients to be aware of private strings, like you are suggesting. I decided I didn't like that approach because of the inconsistency. I felt it was better to have private strings behave the same as private rules, completely hidden to the caller. While a bit more work for YARA it's less work for callers and more consistency too. For me, this extra complexity was worth the consistency and less work for callers.

If you would prefer to switch to a simple flag on strings and let the callers filter them out I can easily do that but I think the lack of consistency and requiring all callers to change their existing code to support it is not the best approach. Ultimately, it's up to you - I'll implement which ever you want to see for your code. :)

wxsBSD: That's a good point about chained strings, and this introducing extra complexity there too. If we were to do as you suggest then 4.0 is certainly the right time to do it. I also agree that if we are going to expose private strings then consistency is important and we should also expose private rules too.

Personally, I'm a big fan of private rules and strings being handled completely internally and never exposed. It just makes more sense for them to only have meaning to libyara and the caller should have no idea they even exist. In my mind, the notion of private rules and strings is only relevant to the internal state of YARA, so introducing extra checks in the caller for something which has no relevance to them (and in fact, the rule author has explicitly said these things have no relevance to the callback) just seems wrong from a usability standpoint.

Ultimately, it's your call and I will happily implement either way (including fixing the chaining part if you think you want to stick with my current implementation).

plusvic: I've considered this carefully and I think we should proceed with the approach of keeping the strings hidden to users of libyara. The release date for YARA 4.0 is uncertain, given that I don't have too much time to spend on YARA lately, and I don't want this feature waiting forever in the pipeline.

wxsBSD: Ok, I’ll get this working with chains strings as soon as I can.

wxsBSD: OK, it's working now for chained strings (and I added a test case to confirm this).

PR #1098, opened 2019-07-30
Added -u/--unmatched command line arg

usualsuspect:
I've added a command line argument (-u/--unmatched) that only prints files that haven't been matched by any rule at all.

This is useful when trying to cover a file corpus with rules as you can quickly see what files aren't detected yet. This can be emulated using the count option (-c) and some command line fu but as it's a feature I frequently use I decided to add it to YARA itself.

### DIFF
--- a/yara.c
+++ b/yara.c
@@ -141,6 +141,7 @@ static bool ignore_warnings = false;
 static bool fast_scan = false;
 static bool negate = false;
 static bool print_count_only = false;
+static bool print_unmatched_only = false;
 static bool fail_on_warnings = false;
 static bool rules_are_compiled = false;
 static int total_count = 0;
@@ -174,6 +175,9 @@ args_option_t options[] =
 
   OPT_BOOLEAN('f', "fast-scan", &fast_scan,
       "fast matching mode"),
+
+  OPT_BOOLEAN('u',"unmatched", &print_unmatched_only,
+      "print unmatched files only"),
 
   OPT_BOOLEAN('h', "help", &show_help,
       "show this help and exit"),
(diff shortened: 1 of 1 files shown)


usualsuspect:
@googlebot I signed it!

plusvic:
Please add the new command-line option to the documentation and man page:

https://github.com/VirusTotal/yara/blob/master/docs/commandline.rst
https://github.com/VirusTotal/yara/blob/master/yara.man

plusvic:
I think this would be more powerful if we add two command-line options (--min-count and --max-count for example), that we can use in combination with `--count` in order to set a desired range for the number of matches. If the number of matches fall in the specified range the file is printed.

What you want would be --count --max-count=0

usualsuspect:
What's the use case for match count ranges? I haven't found a use for match counting at all so far, besides when looking for no matches at all, so I'm curious how people use them.

Personally I like the simplicity of just using "-u" over "--count --max-count=0" but I'm happy either way if I can locate unmatched files without additional scripts.

wxsBSD:
Personally, I'm comfortable with just grep'ing my way to an answer.

`./yara -c foo.yara /bin | grep ": 0$"`

This works great for the unmatched case, but requires a tiny bit more work to find things that have counts in a certain range. :(

plusvic:
Yes, that's exactly why I propose the --max-count and --min-count arguments. Checking for exactly 0 matches is trivial with a combination of commands. I'm not very fond of adding a new command line option when the same result can be obtained by combining other commands. That's the UNIX philosophy after all, keep individual commands simple, but allow to combine them in powerful ways. With the --max-count and --min-count arguments you are giving more flexibility to the user, someone may be interested in finding files that match exactly N rules, or more than N, or less than N. Those cases are more difficult to handle by combining commands. 

The idea is avoiding command line options that serves a single narrow purpose, if we are adding more options let's extract the greatest benefit from it.

PR #1101, opened 2019-07-31
fix error: yara try to open  wrong file

chuangshizhiqiang:
when yara file name is like    test:foo.yar    ,yara try to open foo.yar


chuangshizhiqiang:
@googlebot I signed it!

PR #1107, opened 2019-08-09
Fix chained strings

plusvic:
The function yr_re_ast_split_at_chaining_point was not splitting correctly regexps and hex strings that contained more than one large jump (a.k.a gap). For example, { 01 02 03 [0-1000] 04 05 06 [500-2000] 07 08 09 } was not splitted in three parts as expected, but only in two parts: { 01 02 03 } and { 04 05 06 [500-2000] 07 08 09 } . This didn't affected the correctness of the matches, but it caused ERROR_TOO_MANY_RE_FIBERS in some cases while trying to match { 04 05 06 [500-2000] 07 08 09 } because of the large jump in the pattern. The occurrence of the error depended on the data being scanned.

This PR fixes the aforementioned issue, improves the documentation of related functions and simplifies some portions of the code.

### DIFF
--- a/libyara/parser.c
+++ b/libyara/parser.c
@@ -461,17 +461,18 @@ int yr_parser_reduce_string_declaration(
     SIZED_STRING* str,
     YR_STRING** string)
 {
-  int min_atom_quality = YR_MIN_ATOM_QUALITY;
-  int min_atom_quality_aux = YR_MIN_ATOM_QUALITY;
+  int min_atom_quality = YR_MAX_ATOM_QUALITY;
+  int atom_quality;
 
-  int32_t min_gap;
-  int32_t max_gap;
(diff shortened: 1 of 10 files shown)


libyara/parser.c
```diff
@@ -461,17 +461,18 @@ int yr_parser_reduce_string_declaration(
     SIZED_STRING* str,
     YR_STRING** string)
 {
-  int min_atom_quality = YR_MIN_ATOM_QUALITY;
-  int min_atom_quality_aux = YR_MIN_ATOM_QUALITY;
+  int min_atom_quality = YR_MAX_ATOM_QUALITY;
```
wxsBSD: Shouldn't this be `YR_MIN_ATOM_QUALITY`?

PR #1108, opened 2019-08-09
Fix a heap overrun in the dotnet module.

plusvic:
A heap overrun can occur in call to set_sized_string when called with blob_offset and blob_result.length values that are not validated to be within the boundaries of the buffer containing the PE file. The validation was being done before adding blob_result.length to blob_offset.

Credit to OSS-Fuzz for finding this bug.

### DIFF
--- a/libyara/modules/dotnet.c
+++ b/libyara/modules/dotnet.c
@@ -1137,7 +1137,10 @@ void dotnet_parse_tilde_2(
         break;
 
       case BIT_ASSEMBLYREF:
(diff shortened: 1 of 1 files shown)


plusvic:
BTW, I miss some test cases for the dotnet module. At least a few tests covering the most important features would be nice. Making changes in this module without the safety net of test cases can lead to regression bugs.

wxsBSD:
I'll get some test cases for this module up as soon as I can (hopefully within a week).

PR #1112, opened 2019-08-17
Add field offsets.

wxsBSD:
Add field offsets. This is particularly interesting for certain obfuscators
where the fields contain highly interesting values. Knowing their offsets lets
us use these specific values in detections.


libyara/modules/dotnet.c
```diff
@@ -1661,6 +1685,9 @@ begin_declarations;
   declare_string_array("constants");
   declare_integer("number_of_constants");
 
+  declare_integer_array("field_offset");
```
plusvic: Array names are usual in plural form, this should be named `field_offsets` for consistency. Also this should be added to the documentation.

PR #1114, opened 2019-08-21
Allow include files parallel read access instead of denying all RW operations.

MatejKastak:
Hi,

while I was experimenting with parallel yara compilation of `.yar` files, I encountered an error where if a multiple files share similar includes it would not compile those files and throw an exception.

Currently `_yr_compiler_default_include_callback` opens files exclusively which forbids any writes and even _reads_ from those files. I think it makes more sense to restrict only writing of those files and let read only access succeed. This would allow parallel execution of yara.


MatejKastak:
@googlebot I signed it!

PR #1118, opened 2019-08-30
Update command line example

m417z:
Changed `yara -r /foo` to `yara /foo/bar/rules -r /foo`.
I understand that it's only an example, but I think that it's clearer when it's a valid one.


m417z:
@googlebot I signed it!

Issue #1120, opened 2019-08-31
get_string() and get_float() don't seem to work as documented


Northern-Lights:
I am trying to make use of the `get_string()` and `get_float()` macros in a module.  They are defined here:

https://github.com/VirusTotal/yara/blob/master/libyara/include/yara/modules.h#L318

For `get_string()`, the documentation states that the returned type is `char*`:

https://yara.readthedocs.io/en/latest/writingmodules.html#c.get_string

However, the code appears to instead return a `SIZED_STRING*`:

https://github.com/VirusTotal/yara/blob/master/libyara/object.c#L979
https://github.com/VirusTotal/yara/blob/master/libyara/include/yara/types.h#L726

Additionally, I cannot seem to get `get_float()` to return a non-zero `double`. This macro does not seem to be in the documentation; I am not sure of the reason for that. But when I use it, I always seem to get back the value `0.0000`.

Interestingly, if I put a `printf()` above this line (which returns the double):

https://github.com/VirusTotal/yara/blob/master/libyara/object.c#L952

then I seem to successfully get back the `double` value. But when I do something like:

`double d = get_float(module_object, "x.value")`

where I have definitions such as:

```
begin_declarations;
    begin_struct_array("x");
      declare_float("value");
...
```
then `d` always seems to be `0.0000`, despite my `printf()` line in `yr_object_get_float()` showing a non-zero value.

`get_integer()` seems to work as expected, so I was assuming that `get_float()` should work in essentially the same way. Is there a mistake in my usage somewhere?

Thanks in advance.

Northern-Lights:
Closed in #1122

PR #1122, opened 2019-08-31
fix get_float() for modules

Northern-Lights:
The prototype was missing from the headers. I don't know how it was still compiling, but it was causing the returned value to always be `0.0000`.


Northern-Lights:
@googlebot I signed it!

Northern-Lights:
Fixes #1120

PR #1130, opened 2019-09-10
Add vcpkg installation instructions

grdowns:
YARA is available as a port in vcpkg. Adding installation instructions will help users get started by providing a single command they can use to build YARA and include it into their projects

### DIFF
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -18,7 +18,7 @@ You can download and install YARA using the `vcpkg <https://github.com/Microsoft
     ./vcpkg integrate install
     vcpkg install yara
 
-YARA is kept up to date by Microsoft team members and community contributors. If the version of YARA in vcpkg is out of date, please `create an issue or pull request <https://github.com/Microsoft/vcpkg/>`_ on the vcpkg repository.
+The YARA port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please `create an issue or pull request <https://github.com/Microsoft/vcpkg/>`_ on the vcpkg repository.
 
 Download the source tarball and get prepared for compiling it::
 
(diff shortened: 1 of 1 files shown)


plusvic:
I'm ok with adding this to the documentation, but I think we should add a note specifying that this package is not maintained by YARA's authors and can be out-of-date. My concern is that this package may fall behind when new versions are released.

Out of curiosity, who maintains that package?

grdowns:
Thanks for the feedback @plusvic! I've added some clarification in the README concerning the port maintenance within vcpkg. As an example, I've updated our port of YARA in vcpkg to 3.10.0 https://github.com/microsoft/vcpkg/pull/8194 -- if there are any problems with it you can let me know. We maintain a CMake integration of YARA. As such it does not reflect any changes to the Makefiles in this repo, which can prevent us from automatically updating it. We'd be more than happy to submit the CMake integration upstream if you're interested in maintaining it

docs/gettingstarted.rst
```diff
@@ -10,6 +10,16 @@ find the latest release at https://github.com/VirusTotal/yara/releases.
...
 
+You can download and install YARA using the `vcpkg <https://github.com/Microsoft/vcpkg/>`_ dependency manager::
+
+    git clone https://github.com/microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    vcpkg install yara
+
+YARA is kept up to date by Microsoft team members and community contributors. If the version of YARA in vcpkg is out of date, please `create an issue or pull request <https://github.com/Microsoft/vcpkg/>`_ on the vcpkg repository.
```
wxsBSD: Shouldn't this say "The vcpkg is kept up to date by Microsoft team members and community contributors."

plusvic: Yes, it seems to say that YARA as a whole is maintained by Microsoft.

grdowns: Fixed! Sorry for the poor wording. Thanks for the feedback @wxsBSD and @plusvic. Let me know if there's anything else

PR #1132, opened 2019-09-14
Improved xor modifier.

wxsBSD:
Implement an improved xor modifier which allows you to do:

$a = "This program cannot" xor(1-255)

This will generate every single byte xor for the string from 1 to 255 (inclusive).

This commit also switches the default behavior of xor by itself back to the old behavior of xor'ing with 0 to 255 so it will search for the plaintext too.


libyara/grammar.y
```diff
@@ -601,40 +604,89 @@ string_declaration
...
-    | string_modifiers string_modifier    { $$ = $1 | $2; }
+    : /* empty */
+      {
+        $$.flags = 0;
+        $$.xor_min = 0;
+        $$.xor_max = 0;
+      }
+    | string_modifiers string_modifier
+      {
+        $$.flags = $1.flags | $2.flags;
```
plusvic: Here you are initializing `$$.xor_min` and `$$.xor_max` only when `$2.flags` has the `STRING_GFLAGS_XOR` flag, but what if `$1.flags` is the one that has `STRING_GFLAGS_XOR`? The values for `xor_min` and `xor_max` are not copied to the result in that case. The interesting thing is that there is a test case that should cover this and seems to be working fine, so I may be missing something.

Also, this leads me to another problem that we didn't had until now. What if the `xor` modifier is written twice with different arguments? Right now YARA does't prevents the duplication of a modifier, you can write:

```
  $a = "foo" wide wide 
```

That wasn't a problem until now because modifiers were just flags, but now that they can accept arguments the semantics are more complex. As this is a very unusual case, because it doesn't make sense to write the same modifier twice, I think we can keep the current semantics of accepting repeated modifiers and taking into account only the last occurrence. Just wanted to bring attention to this.

plusvic: I found why this is working. The answer is here:
https://github.com/wxsBSD/yara/blob/xor_fixes_2/libyara/grammar.c#L1689

The default action for a reduction rule when you don't provide one is `$$ = $1`, but interestingly enough bison does this in *all cases* before executing the user provided action. So, you are getting `$$` already initialized with the value of `$1` and this is why it works. However, stated in the bison's comment, this is not documented and we shouldn't rely on this behavior. We most assume that `$$` is always uninitialized .

libyara/lexer.l
```diff
@@ -201,6 +200,14 @@ octdigit      [0-7]
 "import"                { return _IMPORT_;      }
 
 
+"xor" {
+  yylval->modifier.flags = 0;
```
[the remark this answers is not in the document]
wxsBSD: I wasn't a big fan of this change, for the reason you mentioned, but I couldn't really get it to work without it. I'll take another shot at getting it to work without the lexer changes.

libyara/grammar.y
wxsBSD: > Here you are initializing $$.xor_min and $$.xor_max only when $2.flags has the STRING_GFLAGS_XOR flag, but what if $1.flags is the one that has STRING_GFLAGS_XOR? The values for xor_min and xor_max are not copied to the result in that case. The interesting thing is that there is a test case that should cover this and seems to be working fine, so I may be missing something.

My understanding was `$1` is the accumulation of all the other modifiers, so if it has `STRING_GFLAGS_XOR` flag we would have already processed it.

> Also, this leads me to another problem that we didn't had until now. What if the xor modifier is written twice with different arguments? 

I actually think it should be an error but wanted to leave it up to you. IMO, we should reduce ambiguity in the language. I'll update it to make this an error and add a test case for it too.

I'll take another pass through this code tonight and clean it up so that:

1. There are no changes to the lexer.
2. Duplicate modifiers are errors.
3. We don't rely on the undocumented automatic reduction of rules.

plusvic: > > Here you are initializing $$.xor_min and $$.xor_max only when $2.flags has the STRING_GFLAGS_XOR flag, but what if $1.flags is the one that has STRING_GFLAGS_XOR? The values for xor_min and xor_max are not copied to the result in that case. The interesting thing is that there is a test case that should cover this and seems to be working fine, so I may be missing something.
> 
> My understanding was `$1` is the accumulation of all the other modifiers, so if it has `STRING_GFLAGS_XOR` flag we would have already processed it.
> 

Yes, that's correct, `$1` is the accumulation of all the other modifiers. But `$1` could have the XOR flag set and some values in xor_min and xor_max, but if `$2` doesn't have the flag, the values already `$1` are not copied into `$$`. The values in `$$` are the result from this rule, and could be undefined if not explicitly set.

It works because the implementation detail in Bison that forces `$$ = $1` before calling our code. But this is not documented, in theory if you don't set the result value for your reduction rule in `$$` it is undefined.

> > Also, this leads me to another problem that we didn't had until now. What if the xor modifier is written twice with different arguments?
> 
> I actually think it should be an error but wanted to leave it up to you. IMO, we should reduce ambiguity in the language. I'll update it to make this an error and add a test case for it too.
> 
> I'll take another pass through this code tonight and clean it up so that:
> 
> 1. There are no changes to the lexer.
> 2. Duplicate modifiers are errors.
> 3. We don't rely on the undocumented automatic reduction of rules.

Ok, a I agree on making it an error. I hope there are no rules with duplicated modifiers out there.

libyara/grammar.y
```diff
@@ -601,40 +617,98 @@ string_declaration
...
-    | string_modifiers string_modifier    { $$ = $1 | $2; }
+    : /* empty */
+      {
+        $$.flags = 0;
+        $$.xor_min = 0;
+        $$.xor_max = 0;
+      }
+    | string_modifiers string_modifier
+      {
+        set_flag_or_error($$.flags, $2.flags);
```
plusvic: This has a problem similar to the one mentioned in my previous comments. The line `set_flag_or_error($$.flags, $2.flags)` does this:

```
$$.flags |= $2.flags;
```

But it turns out that `$$` may have any random value initially (at least in theory). In practice it doesn't have a random value because it gets initialized by Bison as `$$ = $1`, but Bison doesn't guarantee this, it's just an implementation detail that we shouldn't rely on.

Also, my previous comment still holds:

> Yes, that's correct, $1 is the accumulation of all the other modifiers. But $1 could have the XOR flag set and some values in xor_min and xor_max, but if $2 doesn't have the flag, the values already $1 are not copied into $$. The values in $$ are the result from this rule, and could be undefined if not explicitly set.
> 
> It works because the implementation detail in Bison that forces $$ = $1 before calling our code. But this is not documented, in theory if you don't set the result value for your reduction rule in $$ it is undefined.

I mean, this `if` statement is not enough:

```
        if ($2.flags & STRING_GFLAGS_XOR)
        {
          $$.xor_min = $2.xor_min;
          $$.xor_max = $2.xor_max;
        }
```

If the condition is not true `$$.xor_min` and `$$.xor_max` are not initialized, and we could lose any flag or xor limits stored in `$1`. This is not happening for the same reason explained above.

We should do something like:

```
        $$ = $1;   <-- This guarantees that $$ is initialized with whatever we have in $1, even if this already happens in the current implementation of Bison, we must do it explicitly. 

        set_flag_or_error($$.flags, $2.flags);

        if ($2.flags & STRING_GFLAGS_XOR)
        {
          $$.xor_min = $2.xor_min;
          $$.xor_max = $2.xor_max;
        } 
```

wxsBSD: Thanks for the detailed explanation. I see how it works now, and thank you for pointing it out!

wxsBSD:
I don't know if there is a good answer to this but this string:

`$a = "foobar" xor(1-255) wide`

will generate generate the xor values from 1 to 255 and the wide string, which is probably not what people will mean when they write this. When I think about those modifiers I think "generate me a wide string and then xor it with 1 to 255", not "generate a wide string and the xor version of that wide string from 1 to 255". A poor way to deal with this is to do:

`$a = "f\x00o\x00o\x00b\x00a\x00r\x00" xor(1-255)`

I'm not sure how to deal with this. Thoughts?

plusvic:
> I don't know if there is a good answer to this but this string:
> 
> `$a = "foobar" xor(1-255) wide`
> 
> will generate generate the xor values from 1 to 255 and the wide string, which is probably not what people will mean when they write this. When I think about those modifiers I think "generate me a wide string and then xor it with 1 to 255", not "generate a wide string and the xor version of that wide string from 1 to 255". A poor way to deal with this is to do:
> 
> `$a = "f\x00o\x00o\x00b\x00a\x00r\x00" xor(1-255)`
> 
> I'm not sure how to deal with this. Thoughts?

Right now `"foobar" xor(1-255) wide` extracts the atoms for the wide string (in this case it will be f\x00o\x00) and then apply the xor to those atoms, which results in the interleaved zeroes also xored. That's equivalent to `"f\x00o\x00o\x00b\x00a\x00r\x00" xor(1-255)`, and that's probably what I would expect, right? Would you expect something different?

plusvic:
What I would expect is:

`"foobar" xor(1-255) wide`   only matches the xored wide string
`"foobar" xor(1-255) ascii`   only matches the xored ascii string
`"foobar" xor(1-255) ascii wide`   only matches the xored ascii string and the xored wide string

So, once you use the xor modifier you are not matching the original non-xored string anymore, unless you write something like:

`"foobar" xor(0-255) wide`  it matches the original wide string because you are XORing with 0. Does it makes sense?

wxsBSD:
Yes, this makes sense. I'll get it fixed and update this PR.

wxsBSD:
Updated to address all concerns (I think).

One thing I've realized we can't express is a plaintext string and an xor range that doesn't start at zero. In the current iteration of this PR can you think of a way to do that?

I think the best way to express this would be `xor(0,10-20)` - that is, the grammar would accept a list of ranges. In this example the `0` is just syntactic sugar for `0-0`. Thoughts? Is it worth doing? I'm happy to write the code.

plusvic:
> Updated to address all concerns (I think).
> 
> One thing I've realized we can't express is a plaintext string and an xor range that doesn't start at zero. In the current iteration of this PR can you think of a way to do that?
> 
> I think the best way to express this would be `xor(0,10-20)` - that is, the grammar would accept a list of ranges. In this example the `0` is just syntactic sugar for `0-0`. Thoughts? Is it worth doing? I'm happy to write the code.

Your solution is exactly what I would do, but I would wait and see if this have enough demand and it's worth the effort. The good thing is that this would be backward compatible change. What I would certainly implement now is `xor(X)` as a syntactic sugar for `xor(X-X)` because this is easy to implement.

PR #1135, opened 2019-09-19
PE rich signature improvements

marc-etienne:
This pull request modifies the `pe.rich_signature.toolid` and `pe.rich_signature.version` functions to return the sum of "count values" in the Rich header for the given tool id and/or tool version.

Although it changes the behavior of the functions, it should be backward compatible with existing YARA rules. If there's a match the count should be more than 0, which will evaluates to `true` when converted to boolean. The only case where it wouldn't work is if the YARA rule compares the boolean return value with `1`.

Test was added and doc was updated to reflect the change.

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -792,25 +792,33 @@ Reference
 
         .. versionadded:: 3.5.0
 
-        Function returning true if the PE has the specified *version* in the PE's rich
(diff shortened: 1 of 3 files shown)


marc-etienne:
Note regarding CLA: I'm in the process of getting the corporate CLA signed, it should be done shortly.

marc-etienne:
@googlebot I signed it!

PR #1136, opened 2019-09-25
Fixed potential memory leak when error occurs in paranoid mode.

msuvajac:
When error would occur in paranoid mode there was no cleanup of previously allocated objects (e.g. stack).


plusvic:
Good catch!

PR #1139, opened 2019-09-25
Add Max File Size note

jramirezvt:


plusvic:
Actually this is not a YARA limitation, this is a limitation in how YARA is used in VirusTotal. So, this is not something that belongs to YARA's documentation, this should be clarified in VirusTotal's documentation instead.

PR #1140, opened 2019-09-26
Native Debian packaging

lyndalopez544:
Added Debian packaging info.


aurora-of-earth:
@googlebot I signed it!

aurora-of-earth:
@googlebot I fixed it.

aurora-of-earth:
@googlebot I fixed it.

hillu:
While I have no issue with the Debian package-building recipe being added, I'd like to point out that you are taking other people's work and marking it as your own. Since you edited `debian/copyright`, thereby removing the lines below, it does not seem that this is an accident. By signing the Google CLA, you are making false claims about clauses 2 (grant of Copyright License), 3 (grant of Patent License), and 4 (claim of being legally entitled to grant those licenses).
```
Files: debian/*
Copyright: 2013-2016 Joao Eriberto Mota Filho <eriberto@debian.org>,
                     Hilko Bengen <bengen@debian.org>
License: Apache-2.0
```

plusvic:
@lyndalopez544 I think @hillu is right, the original license was Apache 2.0 and it has been changed to BSD-3-Clause. Also, Hilko was already listed in the AUTHORS file, but Joao's name has disappeared from everywhere. In order to do this right I think that both Joao and Hilko must agree with this change in licensing and being added AUTHORS.

hillu:
@plusvic Since there hasn't been any reaction in almost 2 years,I think this PR should be closed.

PR #1141, opened 2019-10-10
Iterators

plusvic:
This introduces a change in the way in which "for-in" loops work by introducing the concept of iterators. These kind of loops now accept an iterator that will return a sequence of values. Besides the integer ranges (N..M) and integer enumerations (X,Y,Z), this kind of loops now accepts other kinds of iterables, like arrays. This allows conditions like the following one:

for any section in pe.sections: ( section.name == "foo" )

Until now the same expression was written as:

for any i in (0..pe.number_of_sections - 1): ( pe.sections[i].name == "foo" )

The new syntax is more legible and opens door for more powerful features in the future. Backward compatibility is maintained.


libyara/grammar.y
```diff
@@ -1722,14 +1612,63 @@ expression
...
+        $$.type = EXPRESSION_TYPE_INTEGER;
+        $$.value.integer = UNDEFINED;
+      }
+    ;
+
+
 integer_set
-    : '(' integer_enumeration ')'  { $$ = INTEGER_SET_ENUMERATION; }
-    | range                        { $$ = INTEGER_SET_RANGE; }
+    : '(' integer_enumeration ')'
```
wxsBSD: I assume you are intentionally not using string enumerations here? In theory they could be treated as an iterator, right?

libyara/include/yara/types.h
```diff
@@ -827,4 +845,82 @@ struct YR_DICTIONARY_ITEMS
...
+//  +------------+
+//  | true       |  <- true indicates that are no more items.
+//  +------------+
+//  |   . . .    |
+//
+// We can't use the UNDEFINED value in the stack as an indicator of the end
+// of the iterator, because it's legitimate for an iterator to return UNDEFINED
+// items in the middle of the iteration.
+//
+// The "next" function should return ERROR_SUCCESS if every went fine or an
```
wxsBSD: // The "next" function should return ERROR_SUCCESS if everything went fine or an

libyara/exec.c
```diff
@@ -194,21 +194,117 @@ static const uint8_t* jmp_if(
...
+    // Push the false value that indicates that the iterator is not exhausted.
+    stack->items[stack->sp++].i = 0;
+
+    obj = yr_object_array_get_item(
+        self->array_it.array, 0, self->array_it.index);
+
+    if (obj != NULL)
+      stack->items[stack->sp++].o = obj;
+    else
+      stack->items[stack->sp++].i = UNDEFINED;
```
wxsBSD: Rather than push UNDEFINED can't we just skip this item? I'm thinking of a scenario where someone has done something like this in a module:

`
set_integer(1, pe->object, "foo[%i]", 0);
set_integer(2, pe->object, "foo[%i]", 100);
`

In this case wouldn't we be pushing a lot of UNDEFINED values before we get to the next assigned value in the array? Wouldn't it be nicer to only return objects if they are there and skip over the unassigned array values?

libyara/grammar.y
plusvic: That's correct. I'm leaving that for another commit in order to keep this one as small as possible.

libyara/exec.c
plusvic: Actually that was my first idea, but I found out that it wouldn't work as expected. 

Suppose you have an array of integers with 3 items where the first two are undefined, like [undefined, undefined, 10]. If we simply skip the first two items that would be undistinguishable from an array with a single item, the loop would be executed only once, and an expression like this one would be true:

```
for all i in array : ( i == 10)
```

However this expression shouldn't be true because the equality is not true for all items in the array as there are two of them that are not defined.

libyara/grammar.y
wxsBSD: That makes sense. Thanks for clarifying!

libyara/exec.c
wxsBSD: That makes sense. The ways I see around this involve some rather drastic, and possibly backwards-compatability breaking, changes. In particular I'm thinking we can do something like `set_integer(1, pe->object, "foo")` and have it act more like an append function, where you don't get to specify the index. This would at least ensure the values that are set are contiguous, though it doesn't address the fact that you can still set undefined. As I think about it more, I think your approach is the best - it sucks that we can iterate over an arbitrary number of UNDEFINED values but it is the best option.

Thanks for the clarification here!

PR #1143, opened 2019-10-12
Fix .NET parser for big-endian architectures

hillu:
yara 3.11.0 fails the dotnet tests on s390x and other big-endian architectures, see <https://buildd.debian.org/status/logs.php?pkg=yara&ver=3.11.0-1>.

I sprinkled `yr_leXXtoh()` all over the code. Now the tests run successfully, but I am not quite sure whether I missed anything.

### DIFF
--- a/libyara/modules/dotnet.c
+++ b/libyara/modules/dotnet.c
@@ -92,17 +92,17 @@ void dotnet_parse_guid(
   char guid[37];
   int i = 0;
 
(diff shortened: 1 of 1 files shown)


wxsBSD:
To be fair, 3.10.0 (and likely earlier versions) also were wrong but I didn't have tests for these things. ;)

In any case, thanks for fixing these!

PR #1155, opened 2019-10-28
Add pe.number_of_imported_functions and pe.number_of_imported_libraries

marc-etienne:
Add two variables to the PE module to distinguish between library and function imports.

- `number_of_imported_functions`: the count of imported functions

- `number_of_imported_libraries`: same as `number_of_imports` but with a more explicit name to avoid confusion

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -855,7 +855,19 @@ Reference
 
     .. versionadded:: 3.6.0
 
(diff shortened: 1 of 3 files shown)


wxsBSD:
Pretty sure this can be closed as we have `pe.number_of_imports` and `pe.number_of_imported_functions` now.

Issue #1152, opened 2019-10-26
Possibly Incorrect Bitwise Not


utkonos:
Given this rule based on the unit tests: `rule test1 { condition: ~0xA1 & 0xFF == 0x5E }`

Then writing the rule this way: `rule test2 { condition: ~0xA1 == 0x5E }`

The second rule evaluates to false. However, `0xA1` is `10100001`. Performing bitwise not operation on this results in `01011110`. This is `0x5e` in hex.

Here is a sanity check I did using cryptii:
![image](https://user-images.githubusercontent.com/2775739/67613830-2202e380-f780-11e9-9f4e-a4209c8ea580.png)

And just to triple check:
![image](https://user-images.githubusercontent.com/2775739/67613842-3f37b200-f780-11e9-9d21-de1b1099d81c.png)

Finally, just to check the order of operations in the unit test (definitely correct):
![image](https://user-images.githubusercontent.com/2775739/67613859-79a14f00-f780-11e9-9ae5-3583b5515ec8.png)

To show one of my use cases: `~uint8(@op1 + !op1) & 0xFF == !op1` which does work, can be written: `~uint8(@op1 + !op1) == !op1`. I'm using this to calculate a jump that has some variable amount of code contained within a small yara hex string jump in the opcode.

wxsBSD:
This is because the bitwise not is promoting the value to a 64bit integer.

plusvic:
Exactly, when you do ~0xA1 the result is not 10100001, but 11111111111111111111111110100001.

edeca:
As a fix for this, simply AND the value with a mask, for example: 

    rule test2 { condition: (~0xA1 & 0xFF) == 0x5E }

PR #1159, opened 2019-11-06
Add documentation for bitwise not

utkonos:
As a clarification for:
https://github.com/VirusTotal/yara/issues/1152


wxsBSD:
The way this is worded is confusing and this is not specific to ~. Adding two 16 bit integers can overflow into something bigger than 16 bits. The crux is that internally everything is represented as a 64bit integer for bitwise and integer operations (ignoring doubles here). I’m not sure how much of this should be documented as it is an internal implementation detail that very well might change in the future.

If this is going to be documented I think it makes more sense to mention that integers are implicitly promoted and operations are done on the 64bit value. This way it covers more than just the ~ operator.

plusvic:
I agree with @wxsBSD in not documenting this as something specific to the ~ operator. I've added a brief note explaining this in 2aa868a.

PR #1164, opened 2019-11-14
Fix Rich header parsing.

### LINKED ISSUE #1149: Rich Header incorrectly parsed for driver/sys files
In module https://github.com/VirusTotal/yara/blob/master/libyara/modules/pe/pe.c there is an assumption that the Rich Header will start at offset 0x80, however it appears to start at offset 0x200 in several files identified, these all seem to be real mode device drivers.

E.g. a malicious and a bening example:
https://www.virustotal.com/gui/file/582ce3eea9c97d5e89f7d83953a6d518b16770e635a19a456c0225449c6967a4/content/hex
https://www.virustotal.com/gui/file/3593d3d08761d8ddc269dde945c0cb07e5cef5dd46ad9eefc22d17901f542093/content/hex

```
yara test.yara 3593d3d08761d8ddc269dde945c0cb07e5cef5dd46ad9eefc22d17901f542093 -D
pe
...
        rich_signature
                clear_data = UNDEFINED
                raw_data = UNDEFINED
                key = UNDEFINED
                length = UNDEFINED
                offset = UNDEFINED
...
```

wxsBSD:
The rich header doesn't always start at 0x80. The proper way to find the rich
header is to start at the NT header and work backwards one dword at a time
(technically the NT header is supposed to be 8 byte aligned, if someone has a
sample which isn't I'd love to see it) looking for Rich. Once you find it you
can get the XOR key which follows it and then continue to look backwards for
the DanS (after applying the XOR).

Decent documentation on this is at
http://bytepointer.com/articles/the_microsoft_rich_header.htm.

Also, fun fact which we may want to consider implementing is that the XOR key is
actually derived from the DOS header, so it may be an interesting way to tell if
the rich header has been tampered with if they do not properly recalculate the
key.

Fixes #1149.


libyara/modules/pe/pe.c
```diff
@@ -171,91 +172,99 @@ static void pe_parse_rich_signature(
...
 
-  headers_size = yr_le32toh(mz_header->e_lfanew) + \
-                 sizeof(pe_header->Signature) + \
-                 sizeof(IMAGE_FILE_HEADER);
+  // Most files have the Rich header at offset 0x80, but that is not always
+  // true. 582ce3eea9c97d5e89f7d83953a6d518b16770e635a19a456c0225449c6967a4 is
+  // one sample which has a Rich header starting at offset 0x200. To properly
+  // find the Rich header we need to start at the NT header and work backwards.
+  p = (DWORD*)(pe->data + nthdr_offset - 4);
+  while (p > (DWORD*)(pe->data + sizeof(IMAGE_DOS_HEADER)))
```
plusvic: I think that doing everything in the same loop makes the code more complicated than necessary. Why not searching for the Rich header in one loop, and once you find, continue in another loop that computes the signature starting at where the previous loop left? 

This way you avoid the `key == 0` part in `*p == RICH_RICH && key == 0` and the `if (key != 0)` statement and the code is easier to follow.

wxsBSD: I thought about doing that, will update it tonight.

PR #1166, opened 2019-11-14
Fix bug in while initializing the "guids" array in "dotnet" module.

### LINKED ISSUE #1165: Out of memory in dotnet module
OSS-Fuzz has found an out-of-memory condition in the "dotnet" module that is revealing another related bug. The problem is this loop:

https://github.com/VirusTotal/yara/blob/master/libyara/modules/dotnet/dotnet.c#L100

The stop condition is:
 `while (guid_size >= 16 && fits_in_pe(pe, guid_offset, 16))`

One problem is that `guid_size` is read from the file and not sanitized, so a corrupted file can have an arbitrarily large value. But the another problem is that `guid_offset` is not incremented in the loop, so the same guid is set in the `guids` array over and over again.

plusvic:
Closes #1165


plusvic:
@wxsBSD please check if my proposed fix is correct. I would also add some sanitation to make sure that `guid_size` is below a reasonable number, but I don't know what's a reasonable number for it.

A test case for making sure that the `guids` array is populated correctly would be also fine, but the test file in `tests/data` have a single GUID, so it doesn't cover this issue.

plusvic:
> This is fine. I've scanned a small collection of .NET samples I have laying around and 57 out of 1549 have more than one GUID in them (all 57 have exactly 2 GUIDs found). I'd say a reasonable number is under 10? We can always bump it if necessary.
> 
> Thanks for finding and fixing this!

Can you share one of those 57 files with me in order to add it as a test case? Preferable an small non-malware file.

wxsBSD:
> Can you share one of those 57 files with me in order to add it as a test case? Preferable an small non-malware file.

```
26aee1d981fd5cc00198c32bda5cc71d
8e577f61a57f3749848591405de0f5e4
192fde4b6df55eb9070ce81cdbb824c0
0e29e14eaa4e8813b8f1627e1226b6e4
5131060c2d1792cf8cfd28e3975c05db
db04cf22add685dcf8b5e0da9ed8e9d0
9e89345e332530ac546f4652efb3b39e
5ca871e98255190736f0bf4c903c2749
91919321ef2270b2aab50acf5175c9e1
d36a96aa97eef1c0077f49ab8a8f0a78
8389b2a922dca3169c133b57ea0416e2
40460d4d9ab0018067ce9b99448434cc
f885a63051c4825f6cfafd7b51b1c5d0
37760b5f0348c527130a9dca98b81abd
1cb615276b64a423c28a468a8a42f9ad
4520d0c7e743be4977bf6a2e535e7a4d
9982d5bb8c6a5c0dc8298a794306a23f
9cb822da9ebdff9346a36c68e79ccf2e
c6f3c5f9b01dfafbdffd032f7f358d16
654fe11bee844d0b6264a22c3e5e2ded
643ab2dd3d9cbe04552b3e321e322cb8
af7228975bb38794b2443988aad20e6e
44f720ba5203bece4ba732535b53e6cc
55a30f19da5f833c483a3abdd9d2e4d3
3a003859cf11445e7a3d0d9111b4fe3c
848e5f9a44e55fbb4e2fa82e501ab090
c5cd5622cec140ae6fa581707ffd3f3f
5524eeb3368da8a2a39005a5533ffb9a
5298d33e7161e6a7004e7b01dc5e3b35
05dc26c26ac7a010bcff3a2fea4f862a
f2dcab7bf739b88634c6bc3522b3dbc1
59e221fe31b2c193114dee07b9620cba
e1ea2322cfd20d00d896d673b6b822ff
ffc9f0c5f6d1c427bc0e8f2ede3b80b2
4deaea443a1bb42b4acd841b0bbd5fe3
b3e1011549206c0e21dd074b958a6346
c30691c17996dd1797307a651adf6ba6
24921f8f4700ea37712cabd77a1294cd
130cd7440c5cfee07269bdd259c95010
864aa6532468c56f79702c99de77484a
bd604c84067761b119bf2f6d5975c5ca
6cec0c9a8f5dfeda4c8b8b28778f2ca6
3dc8537103fa7c8dfae5fefc537cd01c
9f0d119ebcb28ff99902165ca697329a
6e6c9450d4af1851a5b0a67923fa202f
4f06df880ada88afb869b588b0216235
3e8419cd43374c74c3936a4b9e54adbb
20e66da0694253d9d5f0662e6f0bc11d
1a6beefd2676046c0f3a91dcd0f50d9b
09f2afd807291d0619a79381a73563b8
0e079fbfba3d07742ef126edf646fdd9
551ccced950aa0b916edecab3ad7a56d
667b3182896e05cf901af0bcfe83a096
7c7ca757370442d97dec6e784e5232a1
3c52098b019a45c90286437984aadf8e
fe5b70a9960523b43077180d376d693d
20f11459946f33405386da025395072f
```

I don't know if any of those are benign but they all have more than 1 GUID according to YARA. I spot checked one (9f0d119ebcb28ff99902165ca697329a) and while it does have 32 bytes in the #GUID table the second one is all nulls. It will make a good test case but it's arguable if it is meant to have two GUIDs or if it's an artifact of something else.

PR #1170, opened 2019-11-25
Follow symlinks in yara.c

### LINKED ISSUE #1168: Won't follow symbolic links/symlinks
yara 3.9.0 on Debian 10/x64

I want to scan recursively through a directory which collects multiple subdirectories referenced by a symlink:
```
/files/dir_a/...
/files/dir_b/...
/files/dir_c/...
[...]
/scanfiles/dir_a -> ../files/dir_a
/scanfiles/dir_c -> ../files_dir_c
```

Executing `yara -f -r /yrules/rules.yara /scanfiles/` does not match a single file and quits immedately.

I miss something like a `-F`-switch for `follow symbolic links`.

Is there a chance to see this in the future? Or any advice how I could follow symlinks (no, I won't do a `find ... -exec yara`, because this would be very slow because of process spawing)?

Thank you!

wxsBSD:
Fixes #1168

### DIFF
--- a/yara.c
+++ b/yara.c
@@ -419,11 +419,11 @@ static void scan_dir(
 
       snprintf(full_path, sizeof(full_path), "%s/%s", dir, de->d_name);
 
-      int err = lstat(full_path, &st);
+      int err = stat(full_path, &st);
 
       if (err == 0)
       {
-        if(S_ISREG(st.st_mode) || S_ISLNK(st.st_mode))
+        if(S_ISREG(st.st_mode))
         {
           file_queue_put(full_path);
         }
--- a/yara.c
+++ b/yara.c
@@ -423,13 +423,12 @@ static void scan_dir(
 
       if (err == 0)
       {
-        if(S_ISREG(st.st_mode))
+        if(S_ISREG(st.st_mode) || S_ISLNK(st.st_mode))
         {
           file_queue_put(full_path);
         }
         else if(recursive &&
                 S_ISDIR(st.st_mode) &&
-                !S_ISLNK(st.st_mode) &&
                 strcmp(de->d_name, ".") != 0 &&
                 strcmp(de->d_name, "..") != 0)
         {
--- a/yara.c
+++ b/yara.c
@@ -218,7 +218,7 @@ args_option_t options[] =
       "print tags"),
 
   OPT_BOOLEAN('r', "recursive", &recursive_search,
-      "recursively search directories"),
+      "recursively search directories (follows symlinks)"),
 
   OPT_INTEGER('k', "stack-size", &stack_size,
       "set maximum stack size (default=16384)", "SLOTS"),


yara.c
```diff
@@ -423,13 +423,12 @@ static void scan_dir(
 
       if (err == 0)
       {
-        if(S_ISREG(st.st_mode))
+        if(S_ISREG(st.st_mode) || S_ISLNK(st.st_mode))
```
plusvic: I don't get this, why the path is put into the file queue it's a symlink? The symlink could be pointing to a directory instead of a regular file, right?

Also, not following symlinks was a deliberate decision because, if I remember well, I found a symlink loop in my own machine while testing the code. I wouldn't assume that symlink loops are uncommon, instead I think it's better if we add a command-line flag for enabling this. This way the default behavior remains the same, if anyone wants to follow symlinks they can do, but if a loop is found at least they will figure out what's happening.

github-pba: I'm the requestor of https://github.com/VirusTotal/yara/issues/1168. 

You are right, symlinks can (and in my case, do) point to directories. My original request was for a new command line switch to follow symlinks when recursively parsing through directories.

And I confirm the risk of a loop, when a symlink in a directory points back to the directory the symlink is in:
```
/path/file_a
/path/file_b
/path/subdir/file_c
/path/subdir/file_d
/path/subdir/symlink -> /path
```
This means it's required to track which symlink has already been followed.

I could not make the code change of wxsBSD make to work, see comments on https://github.com/VirusTotal/yara/issues/1168

wxsBSD: By putting the path in the file queue even if it is a symlink we now match the behavior of scanning a single file on the command line, where we do follow symlinks. Rather than confuse people with the difference in behavior I thought making recursive scanning match single file scanning would reduce the surprise of this.

As for loops, see my comment on #1168 - I set up a loop and while it did spend time going around we eventually hit a limit and stop scanning.

github-pba: OK, checked it, and it works.

But I'm quite sure there will be complaints about issues with recursive scanning, which can take quite some time if the recursion is inside a larger directory, which leads to re-scan the same files over and over again, until ... something happens (my system is currently in this state - scanning the same 3 GB test data for the 25th time, finding the same files over and over again ... exactly as expected.

For me, this would be fine, as I know of this risk. But maybe @plusvic is not so wrong when he says following symlinks should propably be activated by a command line switch, passing the responsibility for this behaviour back to the user.

The alternative would be distincting between _file_ and _path to file_. Yara could keep track of the file scanned (for example, by keeping track of the file's inode ID [does every FS has inode IDs?]), and if a file was already scanned, the new scan is omitted.

Meeh. Recursion makes headaches.  :-)

github-pba: the number of times how often the files are scanned depends on the file name length:

22 Times for this structure:
```
root#:~/yara-linktest# ls -l subdir
total 76
-rwxr--r-- 1 root root 77027 Jan  8 14:25 digi.pdf
lrwxrwxrwx 1 root root     1 Jan  8 15:21 very-long-link-name-wasting-very-much-path-space -> .
```

42 times for this structure:
```
root#:~/yara-linktest# ls -l subdir/
total 76
-rwxr--r-- 1 root root 77027 Jan  8 14:25 digi.pdf
lrwxrwxrwx 1 root root     1 Jan  8 15:21 v -> .
```

Let me guess: it quits after the path is longer than 255 (no, must be 1000) characters or the path depth exceeds 40 levels ...

Hm .... unpredictablity hits computer science ...

ThomasGsp:
Hello,

>  I think it's better if we add a command-line flag for enabling this

Definitively.
In some cases, we don't want and don't need to follow symlink in recursive mode.

Eg: Multiples customers directory

- /home/customer_1
- /home/customer_1/...
- /home/customer_2
- /home/customer_2/<sub1>/<sub2>/<symlink_to_realfile>
- /home/customer_2/<sub1>/<sub3>/<realfile>

In any case, we will start the scan from the root ( /home/customer_X  ) and all file will be in the queue.

plusvic:
I've merged this, if people start finding loops we can add a command-line option to turn off symlink following.

PR #1173, opened 2019-11-29
Fix buffer overrun in "dotnet" module.

plusvic:
Credits to OSS-Fuzz.


wxsBSD:
Sorry, I have been working on something else and am just now cleaning out my github notifications. :)

Issue #1175, opened 2019-12-05
Enhancement: Additional Cuckoo module network support


mtonsmann:
See TL;DR

As an avid VT user, I was excited to learn about support for hunting with yara, including support for the cuckoo module. When I started to use the module, I noticed that it did not yet support many of the things that are available in many cuckoo JSON reports.

When looking through issues and PRs, I found #958 which wanted to update the module to support the latest cuckoo JSON format. Unfortunately it looks like this had some issues that were never fixed and it was not merged.

While updating the module to support the latest JSON as well as keeping it backwards compatible seems worthwhile, my request is to simply support 3 new functions to parse existing functionality from the cuckoo reports:

network.host(regexp)
	example: cuckoo.network.host(/192\.168\.1\.1/)
network.tcp(regexp)
	example: cuckoo.network.tcp(/192\.168\.1\.1:80/)
network.udp(regexp)
	example: cuckoo.network.tcp(/192\.168\.1\.1:53/)

I have read the source and think I can add these methods via a PR. I am also happy to write the documentation to support these methods. 

As far as I can tell there is no need to support new versions of cuckoo reports, but if there is interest then I could add support for those as well.

TL;DR before I start work on a PR, are the maintainers interested in seeing support for additional network methods in the cuckoo module?

mtonsmann:
added in #1177

PR #1177, opened 2019-12-06
Cuckoo module upgrades

mtonsmann:
See #1175. Added support for:
- network.host(regexp)
    -  example: cuckoo.network.host(/192.168.1.1/)
- network.tcp(regexp, int)
    - aguments: (destintion_ip, destination_port)
    - example: cuckoo.network.tcp(/192.168.1.1:80/)
- network.udp(regexp, int)
    - aguments: (destintion_ip, destination_port)
    - example: cuckoo.network.tcp(/192.168.1.1:53/)
- network.http_user_agent(regexp)
    - example: cuckoo.network.http_user_agent(/MSIE 6\.0/)

I tested with a JSON cuckoo report I grabbed from a VT CuckooFork run. I looked into creating/running tests, but as there are not any tests for the cuckoo module I decided it was out of scope of this PR to implement that. Happy to help with future work though. 

I think that the whole module could probably use a refactor, but I designed my additions around the existing code set. There is definitely some code-reuse and repetition.

### DIFF
--- a/docs/modules/cuckoo.rst
+++ b/docs/modules/cuckoo.rst
@@ -113,6 +113,13 @@ Reference
 
         Similar to :func:`http_request`, but only takes into account POST
         requests.
+       
+    .. function:: http_user_agent(regexp)
+
+        Function returning true if the program sent a HTTP request with a
+        user agent matching the provided regular expression.
+
+        *Example: cuckoo.network.http_user_agent(/MSIE 6\\.0/)*   
 
     .. function:: dns_lookup(regexp)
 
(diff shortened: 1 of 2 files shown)


docs/modules/cuckoo.rst
```diff
@@ -120,6 +120,36 @@ Reference
...
+
+      .. function:: udp(regexp, port)
+
+        Function returning true if the program contacted an IP address
+        matching the provided regular expression, over UDP on the provided
+        port number.
+
+        *Example: cuckoo.network.udp(/evil\\.com/, 53)*  
+
+      .. function:: user_agent(regexp)
```
plusvic: The name of the function is actually `http_user_agent`, right?

mtonsmann: That's correct. Nice catch!

mtonsmann: fixed in 7fce122

mtonsmann:
Not sure why 832fa82 is the only commit passing here, since all that is changing across the 3 commits is documentation. It appears to have to do with the "--enable-profiling" flag, but I am struggling to see how my changes could affect that, much less why one of the 3 commits passed but the others didn't. Any idea @plusvic?

PR #1186, opened 2019-12-23
Recheck atom quality after generating xor atoms.

### LINKED ISSUE #1172: XOR keyword bypasses atom quality checks
Although all three of the following rules match the same content and have comparable performance, the rule that XORs the letter **A** does not trigger the atom quality warning.
https://github.com/VirusTotal/yara/blob/master/libyara/parser.c#L736

```
rule exp_null
{
    strings:
        $null_20 = { 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 }
    condition:
        $null_20
}

rule exp_null_xor_space
{
    strings:
        $null_20 = { 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 } xor(0x20)
    condition:
        $null_20
}

rule exp_null_xor_A
{
    strings:
        $null_41 = { 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 } xor(0x41)
    condition:
        $null_41
}
```

```
$ time yara -r yara_00 test_dir 1>/dev/null
yara_00(4): warning: $null_20 in rule exp_null is slowing down scanning
real    0m0.322s
user    0m1.772s
sys     0m0.903s

$ time yara -r yara_20 test_dir 1>/dev/null
yara_20(4): warning: $null_20 in rule exp_null_xor_space is slowing down scanning
real    0m0.309s
user    0m1.846s
sys     0m0.793s

$ time yara -r yara_41 test_dir 1>/dev/null
real    0m0.306s
user    0m1.809s
sys     0m0.797s

```

wxsBSD:
When generating xor atoms we need to recheck the atom quality in case we have
just generated bad atoms.

Fixes #1172

### DIFF
--- a/libyara/atoms.c
+++ b/libyara/atoms.c
@@ -1579,14 +1579,15 @@ int yr_atoms_extract_from_string(
     yr_atoms_list_destroy(*atoms);
     *atoms = xor_atoms;
 
-    // Recheck the atom quality, in case we have just generated some poor atoms.
-    // https://github.com/VirusTotal/yara/issues/1172
-    for (item = *atoms; item != NULL; item = item->next)
-    {
-      quality = config->get_atom_quality(config, &item->atom);
-      if (quality < *min_atom_quality)
-        *min_atom_quality = quality;
-    }
+  }
+
+  // Recheck the atom quality, in case we have just generated some poor atoms.
+  // https://github.com/VirusTotal/yara/issues/1172
+  for (item = *atoms; item != NULL; item = item->next)
+  {
+    quality = config->get_atom_quality(config, &item->atom);
+    if (quality < *min_atom_quality)
+      *min_atom_quality = quality;
   }
 
   return ERROR_SUCCESS;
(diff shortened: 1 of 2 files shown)


libyara/atoms.c
```diff
@@ -1576,8 +1576,17 @@ int yr_atoms_extract_from_string(
...
       });
 
-      yr_atoms_list_destroy(*atoms);
-      *atoms = xor_atoms;
+    yr_atoms_list_destroy(*atoms);
+    *atoms = xor_atoms;
+
+    // Recheck the atom quality, in case we have just generated some poor atoms.
+    // https://github.com/VirusTotal/yara/issues/1172
+    for (item = *atoms; item != NULL; item = item->next)
```
plusvic: I think this should be done not only for the `xor` modifier, but for all of them. So that the minimum quality is always computed from the final atoms no matter the modifiers being used.

PR #1187, opened 2019-12-23
Added CounterCraft to Who is using Yara

tomac:


tomac:
@googlebot I signed it!

PR #1210, opened 2020-02-04
CIFuzz Dogfood

Leo-Neat:
OSS-Fuzz was wondering if yara would dogfood CIFuzz,
a new feature being rolled out in OSS-Fuzz. It allows for pull
requests to be fuzzed in CI using fuzz targets enabled in OSS-Fuzz. 
It will be active in dry_run mode meaning that it will never fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,20 @@
+name: CIFuzz
+
+on: [pull_request]
+
+jobs:
+  fuzz:
+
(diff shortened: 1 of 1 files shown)


plusvic:
Sure! Merging this branch is enough?

Leo-Neat:
> Sure! Merging this branch is enough?

Yes! It will just show up as a check that always passes on new pull requests.

PR #1215, opened 2020-02-11
CIFuzz renaming output to artifacts

Leo-Neat:
We had a last minute design decision to rename the output directory to 'artifacts'. Sorry for the inconvenience, and thanks for dog fooding CIFuzz!


plusvic:
Don't worry, that's what dog fooding is for!

PR #1219, opened 2020-02-19
Add check for reproducible build in PE module

cherepanov74:
Since there is a code that parses debug directory, I thought it's a good idea to add some value that can be used for checking if the PE binary build as "reproducible build". If this value is set, it means that timestamps in the PE can be replaced with some other values to achieve reproducibility.

Here is more information about reproducible builds:
https://devblogs.microsoft.com/oldnewthing/20180103-00/?p=97705
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format?redirectedfrom=MSDN#debug-type

### DIFF
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -291,6 +291,8 @@ static void pe_parse_debug_directory(
   size_t pdb_path_len;
   char* pdb_path = NULL;
   
+  set_integer(0, pe->object, "is_reproducible_build");
+  
   data_dir = pe_get_directory_entry(
       pe, IMAGE_DIRECTORY_ENTRY_DEBUG);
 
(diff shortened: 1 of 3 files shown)


libyara/modules/pe/pe.c
```diff
@@ -292,6 +292,8 @@ static void pe_parse_debug_directory(
   size_t pdb_path_len;
   char* pdb_path = NULL;
   
+  set_integer(0, pe->object, "is_reproducible_build");
```
wxsBSD, line 295: Does it make sense to set this to zero here? What happens if the debug directory entry (or something else we check later on) is corrupt. In this case a user can get an incorrect value (`pe.is_reproducible_build` would be zero) when it should be `UNDEFINED` at this point.

cherepanov74: Thank you for your suggestion. Now the value is set to 0 only when debug directory is empty or when code successfully iterated through all entries

PR #1226, opened 2020-03-07
Locale/language bug fix in pe module

vitaly-kamluk:
A change introduced in commit 75babbc0248f14b3ecb135c88d592268f1c1edce
  silently broke language check due to integer size and signedness.

  Tested using pe.language(0x04) on file 33659578e2f6e143c4e0533d598bf690

### DIFF
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -2448,7 +2448,7 @@ define_function(locale)
   for (i = 0; i < n; i++)
   {
     uint64_t rsrc_language = get_integer(
-        module, "resources[%" PRId64 "].language", i);
+        module, "resources[%" PRIi32 "].language",  (int32_t)i);
 
     if ((rsrc_language & 0xFFFF) == locale)
       return_integer(1);
@@ -2479,7 +2479,7 @@ define_function(language)
   for (i = 0; i < n; i++)
   {
     uint64_t rsrc_language = get_integer(
-        module, "resources[%" PRId64 "].language", i);
+        module, "resources[%" PRIi32 "].language", (int32_t)i);
 
     if ((rsrc_language & 0xFF) == language)
       return_integer(1);
--- a/tests/test-pe.c
+++ b/tests/test-pe.c
@@ -302,6 +302,14 @@ int main(int argc, char** argv)
       }",
       "tests/data/weird_rich");
 
+  assert_true_rule_file(
+      "import \"pe\" \
+      rule test { \
+        condition: \
+          pe.language(0x09) and pe.locale(0x0409) \
+      }",
+      "tests/data/mtxex.dll");
+
   yr_finalize();
   return 0;
 }


vitaly-kamluk:
@googlebot I signed it!

wxsBSD:
I’ve become a big fan of test cases in situations like this, to ensure there isn’t a regression in the future. Does any of the existing test PE files exhibit this bug? If so, it would make writing a test super easy.

vitaly-kamluk:
Hey Wes, I've been thinking about the same. Yes, there is a PE file to test this on, see MD5 33659578e2f6e143c4e0533d598bf690, available on VT.

Here is the yara rule that detects language correctly with Yara 3.10 but doesn't work in 3.11:
import "pe"
rule bugcheck_pe_language {
meta:
  description = "This is a test case for bug in Yara 3.11.0"
  bug_type = "regression"
  introduced_by = "75babbc0248f14b3ecb135c88d592268f1c1edce"
  fixed_in = "https://github.com/VirusTotal/yara/pull/1226"
  md5 = "33659578e2f6e143c4e0533d598bf690" //test sample
condition:
  pe.language(0x04) // LANG_CHINESE
}

craiu:
Following the discovery of this bug, we built a testing system that uses a few thousand test files and couple dozen Yara rules to test verify new builds of Yara before deploying them internally. This should catch other similar bugs in the future.

wxsBSD:
> Hey Wes, I've been thinking about the same. Yes, there is a PE file to test this on, see MD5 33659578e2f6e143c4e0533d598bf690, available on VT.

Thanks for the pointer! I didn't want to add another test file to the repository if I could trigger this with an existing one, and I discovered I could trigger it with tests/data/mtxex.dll which already exists in the repository. I wrote up a test case for this at https://github.com/wxsBSD/yara/commit/c04f0e92b142c721f485ea577ac46886cc3a7c67. It would be awesome if you could bring this commit into your branch so they are merged together.

> Following the discovery of this bug, we built a testing system that uses a few thousand test files and couple dozen Yara rules to test verify new builds of Yara before deploying them internally. This should catch other similar bugs in the future.

We set up something similar a while back to continuously check our rules for regressions. It is difficult to get full regression testing since it entirely depends upon what you're scanning and with what rules, which is why when things like this pop up I try to include test cases so a "make check" will catch these regressions if we ever introduce them again.

Thank you Vitaly and Costin for catching this bug and fixing it!

vitaly-kamluk:
> Thanks for the pointer! I didn't want to add another test file to the repository if I could trigger this with an existing one, and I discovered I could trigger it with tests/data/mtxex.dll which already exists in the repository. I wrote up a test case for this at [wxsBSD@c04f0e9](https://github.com/wxsBSD/yara/commit/c04f0e92b142c721f485ea577ac46886cc3a7c67). It would be awesome if you could bring this commit into your branch so they are merged together.

Hey, sorry, I misunderstood you about the test files. 
I pulled your commit, so it's part of the pull request now. Please see googlebot request above, it requires your consent too.

vitaly-kamluk:
@googlebot I consent

PR #1236, opened 2020-03-22
Fix for cpp compilation.

rjusto:


rjusto:
@googlebot I signed it!

PR #1237, opened 2020-03-22
Implement rol modifier

jkj728:
My classmates and I have implemented an rol modifier, which allows searches for every one byte rol modification to a string. This rol modifier also works for ror because they are inverse operations ( ROR(x) = ROL(8-x)). It is currently only implemented for strings but we plan on extending it to hex strings. For example
```
/*This will match on \xa8\xca\xe6\xe8\xca\xc8\x40\x86\xde\xc8\xca\x40\x90\xca\xe4\xca which is the given string shifted left 1 bit*/

rule test {
      strings:
        $a = "Tested Code Here" rol
      condition:
        $a
    }
```


jkj728:
@googlebot I signed it!

acloward:
@googlebot I signed it!

sskousen:
@googlebot I fixed it.

sskousen:
@googlebot I consent.

acloward:
@googlebot I consent.

wxsBSD:
The other modifiers that I added made sense because 1) they were annoying to write otherwise (xor eliminated the need for 256 strings in a single rule) or 2) buggy to get correct (I saw buggy implementations of the base64 string generation code out there). In your case you're saving at most 8 strings in a rule and it isn't that hard to get right.

What are the implications of this combined with other modifiers? How does it work when combined with xor or base64 or wide, for example?

I am starting to think that if we are going to continue to support these modifiers we need a good way to express the composition of them in the language.

jkj728:
@wxsBSD That is a good point. A group of seniors and I at our university have been working on a couple of keyword in different branches for a capstone project. Our project sponsor uses Yara extensively and mentioned to use that he would have great use of a few keywords: a 0-255 bit add, a 0-7 bit ROL/ROR and also a base64 keyword (which you  happened to write and merge in at the same time that we were working on one). 

I see where you are coming from with the fact that the rol keyword only saves the user from writing 7 other strings. As far as integrating it with the other keywords, we hoped to attempt to do that after seeing how the initial keywords were accepted before going too far down a potential rabbit hole. Do you think there is a specific keyword that we could try to integrate the rol with to make it more useful for the framework? or perhaps a way we could improve it to give it a better place within Yara?

I also am going to submit another PR today with the add keyword which definitely saves the user more time in brute forcing 0-255 add keys to eliminate the need for 255 extra strings, we would love your feedback on that potential feature as well. 

Thanks for your help & feedback!

wxsBSD:
I've got some notes jotted down on this and will post an issue for discussion in the next day or so. I'll link this pull request to it when I post it. Generally speaking, I think getting the language in a better spot to support more modifiers without having to rely on implicit interactions and documentation is the next step, and adding new modifiers should probably wait until that discussion has happened and any implementation is done (if any, this is not my project to say what does or does not make it into the language). :)

wxsBSD:
I've hopefully started discussion around ways to make modifiers better in #1249

plusvic:
As I commented on #1249, I agree with @wxsBSD that we either rethink the way in which modifiers are expressed (his proposal in #1249 is good and addresses the issues regarding to how to apply multiple modifiers at the same time), or we stop adding new modifiers. My impression so far is that both "rol" and "add" modifiers can be useful for some people, but not that many. Also, writing YARA rules that do the same thing without these modifiers is a bit harder but bearable.

The backward compatibility issues that such a change will bring are very important, and the benefits of adding these modifiers are out-weighted by the drawbacks in my opinion.

PR #1248, opened 2020-04-01
Implement add modifier

jkj728:
My classmates and I have implemented an add modifier which brute forces add encryption keys up to one byte in length (0-255). It is currently not integrated with the other keywords; however, we hope to add the functionality in the future if this feature is accepted.

Example:
```
/* If encrypted with a key of 0x1f, this string will match on \x73\x84\x92\x93\x84\x83\x3f\x62\x8e\x83\x84\x3f\x67\x84\x91\x84 */

rule test {
      strings:
        $a = "Tested Code Here" add
      condition:
        $a
    }
```


plusvic:
As discussed in #1249 we are not adding new string modifiers unless we solve the problem caused by the increasing number of modifier combinations. Without a mechanism that allows specifying the order in which the modifiers are applied, is hard to  define what should be the behaviour of combinations like `add xor` for example.

PR #1253, opened 2020-04-04
Fix signature parsing on 32-bit big-endian

hillu:
(Fixes test-pe on Debian/powerpc and Debian/hppa.)

### DIFF
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -249,7 +249,7 @@ static void pe_parse_rich_signature(
 
   set_integer(rich_len, pe->object, "rich_signature.length");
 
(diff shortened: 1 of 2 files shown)


libyara/modules/pe/pe.c
```diff
@@ -249,7 +249,7 @@ static void pe_parse_rich_signature(
 
   set_integer(rich_len, pe->object, "rich_signature.length");
 
-  set_integer(rich_signature->key1, pe->object, "rich_signature.key");
+  set_integer(le32toh(rich_signature->key1), pe->object, "rich_signature.key");
```
wxsBSD, line 252: Why not `yr_le32toh` here?

hillu: no reason. thanks for spotting.

Updating.

PR #1256, opened 2020-04-08
Added elf.symtab_symbol function into ELF modul

xbabka01:
Added `elf.symtab_symbol(name)` function into `elf` modul. Function return number of symbols with specified name.


xbabka01:
@googlebot I signed it!

PR #1257, opened 2020-04-09
Update Documentation for Strings and Base64

malvidin:
This has a small change, so that `base64` and `fullword` cause an error if used together. Based on how variable characters and padding are found with Base64 encoded data, this combination will cause more issues. In the rare cases where someone wants to use a specific Base64 encoded string with the `fullword` modifier, I would expect it would be easier to create one Base64 specific encoded string and add the `fullword` modifier to it.  

The other option would be to explicitly state in the documentation the quirky behavior of using both 
modifiers.
 
I also added a summary table for string modifiers and their restrictions.

This also updates the version to 4.0.0 for post 3.11.0 changes.
(edited to correct nocase to fullword)


malvidin:
@googlebot I signed it!

docs/writingrules.rst
```diff
@@ -441,15 +442,15 @@ If you want more control over the range of bytes used with the xor modifier use:
 The above example will apply the bytes from 0x01 to 0xff, inclusively, to the
 string when searching. The general syntax is ``xor(minimum-maximum)``.
 
-base64 strings
+Base64 strings
```
wxsBSD, line 445: Not sure why you're capitalizing `base64` here and in the next few changes.

wxsBSD:
I'd like to see tests that ensure this is a compiler error, so we don't regress on it in the future.

malvidin:
My understanding was that Base64 was capitalized, unless referring directly to the modifier or command, similar to how XOR/xor and ASCII/ascii are treated in the docs.

But looking back through the RFCs, it looks like base64 is also used, but capitalized when used at the beginning of the sentence or in titles.

I plan to make some tests, make uses of `base64` and `xor` modifiers consistent, and use XOR and Base64/base64 when referring to the operation/encoding in general.

malvidin:
I noticed that the `base64` modifier doesn't work properly on single byte. For `a`, It probably generates `Y`, null, and `h`, but the null probably stops it from getting to the `h`. When wide is used, it generates two bytes to search on, so that does not cause the same error.

malvidin:
I first tried changing the order so that the problematic encoding was performed last, but I think that specifying the second iteration with single string length is easier to understand.

malvidin:
This entirely really fix the problem, as there is now one offset base64 encoding that would never match. It might be better to reject single byte ASCII base64.

For `a`
```
Offset 0: YQ==
Offset 1: AGE=
Offset 2: AABh
```
So we extract `Y` and `h`, but `GE` decodes to include `a`, as does `[GWm2][EFGH]`. Without generating these leading and trailing groups, there may be rare occasions where YARA matches on a string that doesn't decode to the original string.

PR #1263, opened 2020-04-27
Fix Mach-O parsing in Mach-O module

PeterMatula:
### The problem
In `MACHO_PARSE_FILE()` there is a command parsing loop. When segment commands are parsed, `seg_count` is incremented. At the end, `number_of_segments` integer is set. But also the `LC_UNIXTHREAD` and `LC_MAIN` commands are parsed in the loop. Their parsing functions use uninitialized `number_of_segments` value - in `macho_rva_to_offset()` and `macho_offset_to_rva()`. This is clearly not safe, but it does not cause problems most of the times. If the binaries are not malformed, rva <-> offset conversion succeeds even when `number_of_segments` is some (random) huge value, because it hits in the first few good segments. However, it the binary is not so correct, this can cause iteration over up to `uint64_t` max value.

### The example
Run YARA (compiled with `--enable-macho`) on the following file [test.zip](https://github.com/VirusTotal/yara/files/4538444/test.zip) with the following ruleset:
```
import "macho"
```
If you are not (un)lucky, this will take a very long time because there is some huge number in the uninitialized `number_of_segments`. E.g. `18445260673521474303` in my case.

### The solution
A 2-step commands processing:
* First only the segments are processed.
* Then other commands that use info on the number of segments.

Cons: two steps.
Pros: value is initialized, and is a general solution.

Less general solution would be to keep the single loop and set `number_of_segments` after every segment command is processed. This would however rely on specific ordering of commands - i.e. commands using segment N go only after command for segment N itself. Otherwise it would not work. I don't know if this can be expected, so its better to be safe than sorry and loop twice.

@mbandzi can you please take a look at it? @metthal said you are the original author.

### DIFF
--- a/libyara/modules/macho/macho.c
+++ b/libyara/modules/macho/macho.c
@@ -360,9 +360,9 @@ void macho_parse_file_##bits##_##bo(                                           \
     set_integer(yr_##bo##32toh(header_64->reserved), object, "reserved");      \
   }                                                                            \
                                                                                \
+  /* The first command parsing pass handles only segments. */                  \
   uint64_t seg_count = 0;                                                      \
   uint64_t parsed_size = sizeof(yr_mach_header_##bits##_t);                    \
-                                                                               \
   uint8_t *command = (uint8_t*)(header + 1);                                   \
   for (unsigned i = 0; i < yr_##bo##32toh(header->ncmds); i++)                 \
   {                                                                            \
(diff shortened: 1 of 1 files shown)


PeterMatula:
I signed the CLA, but I'm not sure if it is linked correctly because I did not commit under my primary address.

PeterMatula:
@googlebot I signed it!

mbandzi:
Looks good to me. As far as I remember, the command order is not guaranteed in Mach-O so I would keep two loops.

knightsc:
@PeterMatula Do you know if this problem also exists in my branch here?

https://github.com/VirusTotal/yara/pull/1038

I've re-organized the Mach-O code in that branch and I'm curious if the issue still exists.

PeterMatula:
Btw, I don't think that that `continuous-integration/appveyor` fail is caused by my changes. It is not even in a file I touched.

knightsc:
@PeterMatula I updated your bug fix to work on the updated mach-o parsing code that was just merged.

https://github.com/VirusTotal/yara/pull/1272

I also included a test case in that PR.

PeterMatula:
@knightsc does this mean #1272 has the changes from my (this) PR and the bug is fixed in there? If so, and I see #1272 was already merged, then I can close this and consider it solved.

knightsc:
@PeterMatula , yes, you are correct. Your bug fix should be committed now. Good find and fix!

PeterMatula:
Great, thanks for getting it to `master`.

PR #1264, opened 2020-04-27
strtok_s is not defined in Cygwin. Use strok_r instead.

plusvic:
Also do some minor re-styling.


lmzke:
One additional thought: should the Linux side scan list line parsing gracefully consume carriage returns?  It should probably adhere to any rule file accommodations.

PR #1266, opened 2020-04-30
Move variable declaration to start of function.

wxsBSD:
Fixes https://github.com/VirusTotal/yara-python/issues/143


patrickml:
@wxsBSD do you know if there were any other changes like this in the 4.0 release?

wxsBSD:
There might be, I haven't swept the tree for all of them. Best I can suggest is to apply the patch and test the build using your toolchain.

PR #1271, opened 2020-05-04
Classes

regeciovad:
The goal of these changes is enable to use of rules like folowing in the VirusTotal Hunting:
```
rule contains_btc_address 
{
    strings:
        $re1 = /1[a-km-zA-HJ-NP-Z1-9]{25,34}/ fullword ascii wide
        $re2 = /3[a-km-zA-HJ-NP-Z1-9]{25,34}/ fullword ascii wide
    condition: 
        $re1 or $re2
}
```
The implementation allows a use of multiple characters per byte in atoms and per state in Aho-Corasick automaton. This is done via encoding in bitmask, part that is already used in YARA.

Some additional actions are needed, the Aho-Corasick has to be deterministic even for calculation of failure links. 

However, simplification is shown in many areas. For example, the `nocase` option only involves the additional characters in the bitmask, new atoms are not created. The atoms contain more information about strings and the Aho-Corasick automaton can be effective when used for searching.

The resulting transition table can be also smaller and more compact than if the same strings were generating multiple atoms.


libyara/ahocorasick.c
```diff
@@ -152,30 +178,138 @@ static int _yr_ac_queue_is_empty(
...
-// Given an automaton state and an input symbol, returns the new state
-// after reading the input symbol.
+// Returns true if the bitmap1 and bitmap2 are the same.
+//
+
+static bool _yr_ac_compare_classes(
+    YR_BITMASK* bitmap1,
+    YR_BITMASK* bitmap2)
+{
+  for (int i = 0; i < YR_BITMAP_SIZE; i++)
```
plusvic, line 190: What about using memcmp? memcmp is usually highly optimized and can leverage SSE instructions.

libyara/ahocorasick.c
```diff
@@ -871,16 +1618,27 @@ int yr_ac_add_string(
 
     for (int i = 0; i < atom->atom.length; i++)
     {
-      YR_AC_STATE* next_state = _yr_ac_next_state(state, atom->atom.bytes[i]);
+      if (atom->atom.mask[i] == YR_ATOM_TYPE_ANY)
+      {
+        yr_bitmask_clear_all(atom->atom.bitmap[i], sizeof(atom->atom.bitmap[i]));
+        for (int j = 0; j < YR_BITMAP_SIZE; j++)
```
plusvic, line 1624: The intention here is setting all the bits in the bitmap to 1, so implementing `yr_bitmask_set_all` using `memset` would be clearer and more efficient.

libyara/include/yara/types.h
```diff
@@ -186,6 +186,15 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     ((x) != NULL ? (x)->type == EXTERNAL_VARIABLE_TYPE_NULL : true)
 
 
+// The types of states in AC automaton
+#define YR_ATOM_TYPE_ANY                     0x00
+#define YR_ATOM_TYPE_CLASS                   0xCC
+#define YR_ATOM_TYPE_MASKED_LITERAL_0F       0x0F
```
plusvic, line 192: It looks like `YR_ATOM_TYPE_MASKED_LITERAL_0F` and `YR_ATOM_TYPE_MASKED_LITERAL_F0` are not used.

libyara/ahocorasick.c
```diff
@@ -240,6 +381,473 @@ static int _yr_ac_state_destroy(
...
+
+
+// Creates deterministic AC for failure function
+static bool dfa_subtree(
+    YR_AC_STATE* state,
+    YR_AC_STATE* input_state,
+    YR_ARENA* arena)
+{
+  YR_AC_STATE* next_state;
+  YR_AC_STATE* dfa_state1 = (YR_AC_STATE*) yr_malloc(sizeof(YR_AC_STATE));
```
plusvic, line 806: When the atom is not `YR_ATOM_TYPE_CLASS` nor `YR_ATOM_TYPE_ANY` (most of the cases) these two calls to `yr_malloc` are executed unnecessarily. These two states can be allocated only when atom's type is  `YR_ATOM_TYPE_CLASS` or `YR_ATOM_TYPE_ANY`.

regeciovad:
Hi Victor, thanks a lot for your helpful recommendations. I incorporate them all in the last commit. I also tweaked the function for managing matches list for new states in Aho-Corasick automaton.

PR #1276, opened 2020-05-12
Update YARA Sandboxed API for 4.0

cblichmann:
This PR updates YARA's sandbox to use the latest Sandboxed API and makes it compatible to the API changes in YARA 4.0.

Other changes:
- Updated Bazel dependencies to use `maybe()`
- Update to latest Abseil and newer GoogleTest to match what Sandboxed API uses (to avoid duplication)
- Force C++ parts to be build with C++17, as that is required for the sandbox

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,2 @@
+# Build in C++17 mode without a custom CROSSTOOL
+build --cxxopt=-std=c++17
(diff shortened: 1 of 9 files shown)


cblichmann:
I believe the CIFuzz errors are unrelated. The root cause seems to be a missing `yacc` for generating YARA's lexer/parser.

PR #1280, opened 2020-05-18
added pe.is_pe() documentation

mattnotmax:
Added documentation with leading example.


mattnotmax:
@googlebot I signed it!

PR #1282, opened 2020-05-18
Solving linking error for Openssl > 1.1

soumy:


wxsBSD:
Thanks for catching this. I don't have a version of OpenSSL around that is new enough to have these functions deprecated out, but after checking https://github.com/openssl/openssl/blob/master/include/openssl/x509.h#L670-L675 I think this should be fine. It will work with the current versions of OpenSSL and should set us up to work with new versions.

PR #1313, opened 2020-06-30
Added -m option to select file extensions to scan

emmanubo:
- upgraded VS project to VS2019
- added -m option (all changes in yara.c)


emmanubo:
closing pull request - will resubmit it after fixing a few things

PR #1314, opened 2020-07-01
added --exts argument in order to specify which file extension(s) should be scanned

emmanubo:
added --exts argument in order to specify which file extension(s) should be scanned.
This change is fully implemented in yara.c.
Usage example : 

     yara64.exe -C webshells.bin c:\inetpub\wwwroot  --exts=.asp/.aspx/.jsp/.php

### DIFF
--- a/.vs/ProjectSettings.json
+++ b/.vs/ProjectSettings.json
@@ -1,0 +1,3 @@
+{
+  "CurrentProjectSetting": "No Configurations"
+}
--- a/.vs/VSWorkspaceState.json
+++ b/.vs/VSWorkspaceState.json
@@ -1,0 +1,6 @@
+{
+  "ExpandedNodes": [
+    ""
+  ],
+  "PreviewInSolutionExplorer": false
+}
--- a/cli/yara.c
+++ b/cli/yara.c
@@ -158,14 +158,11 @@ static int max_strings_per_rule = DEFAULT_MAX_STRINGS_PER_RULE;
 #define USAGE_STRING \
     "Usage: yara [OPTION]... [NAMESPACE:]RULES_FILE... FILE | DIR | PID"
 
-// yaramod - added :
-//----------------------------------------------------------
-
 static char* argsearchexts = NULL;
 char* emptyext = ".";
 
-// check if file matches extension specified using -m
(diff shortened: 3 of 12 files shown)


emmanubo:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2FVirusTotal%2Fyara%2Fpull%2F1314) for more info**.

@googlebot I fixed it.

emmanubo:
@googlebot I fixed it.

DanielRuf:
Can you remove the .vs files and folders from this PR?

DanielRuf:
Why is the separator for the option list `/` instead of `,`?

See also the current CLI options and how they are used, for example multiple `-t` options to create an array.

emmanubo:
Hello Daniel, I cleaned up unneeded files in the PR (I couldn't delete the sqllite file but it's not required). Regarding the use of "/" as a separator, let's say it's a personal choice but using "," is fine too. I hope this feature will be implemented. I understand that you need the opposite feature (to select files not having a specific ext), I think it would be nice to have it too.

DanielRuf:
> Hello Daniel, I cleaned up unneeded files in the PR (I couldn't delete the sqllite file but it's not required).

Thank you very much. Looks much better now.

> Regarding the use of "/" as a separator, let's say it's a personal choice but using "," is fine too

I think this is the least problem since the final implementation can still change, as long as it works.

> I hope this feature will be implemented.

I hope so too.

> I understand that you need the opposite feature (to select files not having a specific ext), I think it would be nice to have it too.

Exactly. I need the opposite (exclude specific file extensions from the scan).

PR #1339, opened 2020-08-18
Fix rsrc with bad RVA.

wxsBSD:
In some cases the resource data has an RVA which does not map to a file offset.
Prior to this change we would stop parsing that resource when this is found,
which means we are losing useful information like the length and type/name
strings (if they exist).

I checked how tools like pe-studio and CFF explorer handle this case and they
just display the RVA with an unknown offset.

I think it is useful to not stop parsing if the RVA does not map to a file
offset. This commit makes it so "invalid" RVAs are set as YR_UNDEFINED and we
continue parsing. I'm also adding support for setting the RVA, because it is
useful to know in some cases.

I was careful to verify that in the case of parsing version information
resources that function does check that the RVA maps to a legitimate file offset
before parsing, so this is safe to do.

### DIFF
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -707,13 +707,16 @@ static int pe_collect_resources(
 {
   DWORD length;
 
+  set_integer(
+        yr_le32toh(rsrc_data->OffsetToData),
+        pe->object,
+        "resources[%i].rva",
+        pe->resources);
+
   int64_t offset = pe_rva_to_offset(pe, yr_le32toh(rsrc_data->OffsetToData));
 
   if (offset < 0)
-    return RESOURCE_CALLBACK_CONTINUE;
-
-  if (!fits_in_pe(pe, pe->data + offset, yr_le32toh(rsrc_data->Size)))
-    return RESOURCE_CALLBACK_CONTINUE;
+    offset = YR_UNDEFINED;
 
   set_integer(
         offset,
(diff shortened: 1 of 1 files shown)


wxsBSD:
Here is an example file which triggers this edge case: https://www.virustotal.com/gui/file/540519c9b6cc4da671e61bc10c61fd67c3f10f872a665d6315a4cae7ab4ac37f/details - notice the two resources with a hash of `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`. Both of them have an RVA that does not map to a file offset so we would incorrectly skip over them when parsing. This is the output of pe-studio for one of them:

Screen Shot 2020-08-17 at 10 27 00 PM

wxsBSD:
Also, thanks to @tlansec for noticing this and bringing it to my attention.

plusvic:
A test covering this edge case would be nice.

wxsBSD:
I wanted to add a test case but didn't want to add a malicious binary (most of the ones I've seen do this are malicious) to the repo. Now that I think about it, I'll see if I can't find a small PE with resources and modify the RVA manually. I'll get a test case up for this in the next day or so.

plusvic:
I found an issue with this change. While scanning corrupted files the PE module can consume a lot of memory because no check is performed to make sure that the resource entry is valid. We should set a limit to the number or items that can be added to the `pe.resources` array.

PR #1348, opened 2020-09-03
Add number_of_imported_functions.

### LINKED ISSUE #1345: [Feature Request] - number_of_imported_functions
Hi,

I was wondering if it would be possible to include a count of the number of imported functions in addition to the number of imports, e.g. for the following file:

https://www.virustotal.com/gui/file/0f75fabe825f8a7a55714d13535848c46f9a4e72fa748907e269f9c5a96febab/details

It would be:

number_of_imports - 1 
number_of_imported_functions - 1 

Whereas if it imported more than just "_CorDllMain" the # of functions would be larger.

I appreciate that for large import tables this might not be fast, so it might be better to make this available as a function, in a similar fashion to pe.imphash().

Cheers,
Tom

wxsBSD:
Add the number of imported functions which were parsed. This is the TOTAL number
of functions across ALL imported libraries.

Fixes #1345.


wxsBSD:
@tlansec - this does what you want but I'm curious if you think it's a good idea to also provide something like:

`pe.imports_func_count("foo")`

This would return the number of imported functions for the "foo" library. Another option, which should be backwards compatible for most people, is to make `pe.imports("foo")` return the number of imported functions from "foo" instead of just 0 or 1.

Being able to express both the total count across all libraries (as my patch currently does) and a per-library count can provide some greater flexibility in crafting rules.

tlansec:
That sounds like a good addition too.

In this case, the reason i want it is so that i can look for PE files with specific imports and that also have a small # in total, to kinda heuristically looked for packed files - therefore the total count is more important to me, but I can certainly see that the count for a given lib may also be interesting for some rules.

wxsBSD:
Turns out the ability to get function counts by library was merged in #1156 - so there's nothing left for me to add.

plusvic (approved):

plusvic (approved):

PR #1349, opened 2020-09-04
Implement -N (--no-follow-symlinks).

wxsBSD:
Implement a -N flag which causes yara to not follow any symlinks when scanning,
regardless of the recursive option.

### DIFF
--- a/cli/args.c
+++ b/cli/args.c
@@ -101,7 +101,7 @@ args_error_type_t args_parse_option(
   switch (opt->type)
   {
     case ARGS_OPT_BOOLEAN:
-      *(bool*) opt->value = true;
+      *(bool*) opt->value = !(*(bool*) opt->value);
       break;
 
     case ARGS_OPT_INTEGER:
--- a/cli/yara.c
+++ b/cli/yara.c
@@ -118,6 +118,12 @@ typedef struct COMPILER_RESULTS
 
 } COMPILER_RESULTS;
 
+typedef struct SCAN_OPTIONS
+{
+  bool follow_symlinks;
+  bool recursive_search;
+} SCAN_OPTIONS;
+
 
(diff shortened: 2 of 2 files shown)


wxsBSD:
This fixes #1347 - but it does so in a manner which I think is better than the suggestion. With the `-N` flag you can turn off symlinks when scanning, regardless of the recursive option.

wxsBSD:
Here's an example of the behavior.

```
wxs@wxs-mbp yara % cat rules/a.yara
import "hash"

rule a {
  condition:
    // This is the sha256 of /bin/ls on my system.
    hash.sha256(0, filesize) == "43969ad2a7eaf877f6852423f6ad803d7d89183ac152983d8d71e34d91c78079"
}
wxs@wxs-mbp yara % ls -l ~/tmp
total 104
lrwxr-xr-x  1 wxs  wxs      4 Sep  3 15:47 foo@ -> /bin
lrwxr-xr-x  1 wxs  wxs      7 Sep  3 15:48 ls@ -> /bin/ls
-rwxr-xr-x  1 wxs  wxs  51888 Sep  3 22:28 ls.copy*
wxs@wxs-mbp yara % ./yara rules/a.yara ~/tmp
a /Users/wxs/tmp/ls.copy
a /Users/wxs/tmp/ls
wxs@wxs-mbp yara % ./yara -r rules/a.yara ~/tmp
a /Users/wxs/tmp/ls.copy
a /Users/wxs/tmp/foo/ls
a /Users/wxs/tmp/ls
wxs@wxs-mbp yara % ./yara -rN rules/a.yara ~/tmp
a /Users/wxs/tmp/ls.copy
wxs@wxs-mbp yara %
```

wxsBSD:
The CIFuzz check is failing to build due to missing flex. This looks like it has nothing to do with my changes.

PR #1350, opened 2020-09-11
Switch maximum string limit to warning.

### LINKED ISSUE #1346: Change handling on strings that cause too many matches.
Hello,

At the moment sometimes when scanning large files its possible for even relatively sensible rules to cause an error of the format:

"error scanning foo blah in rule bar causes too many matches" 

The current behaviour when this happens is to stop scanning the file (afaik, based on testing with @wxsBSD ), but in this issue i would like to request a change so that instead it continues scanning the file. The requirements to make this work are:

1) Create a "max count" for each string, where each string can only match that # of times (I am unsure on what a sensible value for this is, but it should be based on whatever number currently causes this error to occur)
2) When this max count is reached for a given string, YARA should stop recording offsets // matches and should print a warning message "Warning maximum string count for string X matched in rule Y - rule Y may no longer work as intended"

The reason for this request is that although it doesnt occur often for regular files, when scanning large files it's possible for a variety of reasons that strings that are sensible on smaller files will cause errors, meaning users have to edit their rules file and rescan (which depending on the volume of data being scanned may be undesirable).

I also think that its not obvious to users with the current error message that the error will cause no rules to match the file, so as a minimum the error message could be modified to be more explicit.

Thanks,
Tom

wxsBSD:
When we hit the maximum match limit on a string we currently abort all scanning,
which can be problematic since it isn't a good reason to stop all scanning, and
is better off as a warning to the user.

This commit makes it so that when the string match limit is hit a warning is
printed and scanning will continue. The warning message indicates that the
results may be invalid. It is hard to know for certain if they are invalid
because it depends entirely upon the condition. If you use the # or @ operator
then your condition _MAY_ be invalid since the total number of matches and
locations is not known anymore. Just checking that the string exists won't be an
invalid state since we have confirmed the existence of _AT LEAST_ the maximum
number of occurrences, which clearly satisfies the existence of a single
instance of your string.

I decided to implement this as a callback so the user of libyara has some choice
on the default behavior. If they want to consider this a "stop all scanning"
error they are free to return CALLBACK_ERROR in their callback.

Fixes #1346.


wxsBSD:
Here's an example of this in action. First, the setup:

```
wxs@wxs-mbp yara % cat rules/x.yara
rule x {
  strings:
    $x = "XXXXXXXXXXXXXXX"
  condition:
    all of them
}

rule a { condition: true }
wxs@wxs-mbp yara % python3 -c "print('X' * 10000000)" > ~/x
wxs@wxs-mbp yara %
```

Now in master it just stops scanning and never reports ANY matches:

```
wxs@wxs-mbp yara % ./yara rules/x.yara ~/x
error scanning /Users/wxs/x: string "$x" in rule "x" caused too many matches
wxs@wxs-mbp yara %
```

Here it is with my changes:

```
wxs@wxs-mbp yara % ./yara rules/x.yara ~/x
Warning: maximum matches for string $x. Results may be invalid.
x /Users/wxs/x
a /Users/wxs/x
wxs@wxs-mbp yara %
```

It reports that we hit the limit and any results using that string may be invalid which depends entirely upon how the string is used in the condition. It then finishes scanning (no longer recording matches for that string) and then evaluates the conditions.

I'm not super happy with the implementation and still have to make it report the rule identifier to which that string belongs, but otherwise this is ready for review. I liked the idea of exposing it as a warning in the callback and letting users of libyara decide what to do with it depending upon their environment. This approach also opens up the possibility for any other similar things to be handled as warnings too.

tlansec:
From an output perspective - this looks sensible.

secDre4mer:
I agree that this looks sensible, though I'd suggest that `in rule "x"` should still be a part of the message. When scanning with a lot of rules that use the same naming schema for their strings, identifying the one that caused the warning will be difficult otherwise.

libyara/scan.c
```diff
@@ -999,6 +1001,25 @@ int yr_scan_verify_match(
...
+  if (context->matches[string->idx].count == YR_MAX_STRING_MATCHES)
+  {
+    result = callback(
+        context,
+        CALLBACK_MSG_SCAN_WARNING_MATCH_LIMIT,
+        (void*) string,
+        context->user_data);
+
+    // Bump the counter so we don't call the callback again.
+    context->matches[string->idx].count++;
```
plusvic, line 1013: I think this logic is not correct. When the count reaches `YR_MAX_STRING_MATCHES` the callback is called with `CALLBACK_MSG_SCAN_WARNING_MATCH_LIMIT`, let's say that the callback returns `CALLBACK_CONTINUE`, in which case the scanning will continue. If some other match is found for the same string now the count is larger than `YR_MAX_STRING_MATCHES` and the callback won't be called again, so YARA will behave as always a continue tracking the matches.

It looks like with this logic YARA is ignoring a **single** match.

wxsBSD: You're right. I'll get things fixed up (and add tests that make me more confident I'm doing the right thing) in the next day or so.

wxsBSD: I think I've fixed this now, and I also added a test case which should keep me honest in the future.

If you think this is right I'll take a shot at adding the documentation for this.

plusvic:
Now it looks OK. You can add the documentation.

libyara/include/yara/error.h
```diff
@@ -104,6 +104,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define ERROR_INVALID_MODIFIER               59
 #define ERROR_DUPLICATED_MODIFIER            60
 
+#define WARN_TOO_MANY_MATCHES                -1
```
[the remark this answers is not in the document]
wxsBSD: Oops, leftovers from a previous attempt at this.

wxsBSD:
Documentation is up, and I'll spend some time this weekend getting something together that exposes this via yara-python.

cli/yara.c
```diff
@@ -1044,7 +1044,13 @@ static int callback(
       mutex_unlock(&output_mutex);
     }
 
-    return CALLBACK_CONTINUE;
+      return CALLBACK_CONTINUE;
+
+    case CALLBACK_MSG_SCAN_WARNING_MATCH_LIMIT:
+      printf("Warning: maximum matches for string %s. Results may be invalid.\n",
```
wxsBSD, line 1050: Does it make sense to send this to stderr instead of stdout?

plusvic: Yep, I think that stderr would be better.

Issue #1356, opened 2020-09-17
YARA newbie trying to understand use of `data_base` and `offset`


simonhf:
Hello, I'm a YARA newbie and trying to better understand the C internals. I found something a little confusing, and so I thought I'd ask here for further clarification:

Have been looking at the function `yr_scan_verify_match()` [1] which gets passed parameters `data`, `data_size`, `data_base`, and `offset`. Everything is pretty obvious except for `data_base` and `offset`. These appear to be not one, but two different 'offsets' into the bytes at pointer `data` of length `data_size`:

```
int yr_scan_verify_match(
    YR_SCAN_CONTEXT* context,
    YR_AC_MATCH* ac_match,
    const uint8_t* data,
    size_t data_size,
    uint64_t data_base,
    size_t offset)
{
...
```

The function also appears to use both offsets together in the following code [2]:

```
  if (STRING_IS_FIXED_OFFSET(string) &&
      string->fixed_offset != data_base + offset)
    return ERROR_SUCCESS;
...
}
```

And then later on `data_base` and `offset` are passed to e.g. `_yr_scan_verify_literal_match()` [3] which appears to ignore `data_base` and only use `offset`:

```
static int _yr_scan_verify_literal_match(
    YR_SCAN_CONTEXT* context,
    YR_AC_MATCH* ac_match,
    const uint8_t* data,
    size_t data_size,
    uint64_t data_base,
    size_t offset)
{
...
    if (STRING_IS_ASCII(string))
    {
      forward_matches = _yr_scan_icompare(
          data + offset,
          data_size - offset,
          string->string,
          string->length);
    }
...
}
```

At first glance this appears to be a bug, but the code seems to be working never-the-less because AFAICT `data_base` is always zero at run-time? Under which circumstances is `data_base` non-zero? If it's not a bug, then what's a good explanation for how `data_base` is intended to be used?

Thanks in advance!

[1] https://github.com/VirusTotal/yara/blob/master/libyara/scan.c#L883
[2] https://github.com/VirusTotal/yara/blob/master/libyara/scan.c#L906
[3] https://github.com/VirusTotal/yara/blob/master/libyara/scan.c#L778

simonhf:
It looks like the terms `base` and `offset` are mentioned here [1] together with the comment "While scanning a file [base] is usually zero":

```
  .. c:member:: int64_t base

    Base offset/address for the match. While scanning a file this field is
    usually zero, while scanning a process memory space this field is the
[...]

PR #1361, opened 2020-09-30
Fix multi block bugs & run test-rules with single & multiple blocks

simonhf:
See https://github.com/VirusTotal/yara/issues/1356

### DIFF
--- a/libyara/include/yara/scanner.h
+++ b/libyara/include/yara/scanner.h
@@ -92,7 +92,7 @@ YR_API int yr_scanner_scan_mem_blocks(
     YR_MEMORY_BLOCK_ITERATOR* iterator);
 
 
-YR_API int yr_scanner_scan_mem(
+YR_API int (*yr_scanner_scan_mem)(
     YR_SCANNER* scanner,
     const uint8_t* buffer,
     size_t buffer_size);
--- a/libyara/include/yara/scanner.h
+++ b/libyara/include/yara/scanner.h
@@ -92,7 +92,7 @@ YR_API int yr_scanner_scan_mem_blocks(
     YR_MEMORY_BLOCK_ITERATOR* iterator);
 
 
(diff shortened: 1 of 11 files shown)


simonhf:
@googlebot I signed it!

libyara/include/yara/scanner.h
```diff
@@ -92,7 +92,7 @@ YR_API int yr_scanner_scan_mem_blocks(
     YR_MEMORY_BLOCK_ITERATOR* iterator);
 
 
-YR_API int yr_scanner_scan_mem(
+YR_API int (*yr_scanner_scan_mem)(
```
plusvic, line 95: I wouldn't do this, it changes the type of `yr_scanner_scan_mem` which is an exported function part of the API. I think you don't need to completely replace `yr_scanner_scan_mem` in tests, you can use `yr_scanner_scan_mem_blocks` providing your own iterator with custom `first` and `next` functions.

plusvic:
Building in Mac OS is failing with:

```
duplicate symbol _yr_scanner_scan_mem in:
    .libs/exec.o
    .libs/rules.o
duplicate symbol _yr_scanner_scan_mem in:
    .libs/exec.o
    .libs/scanner.o
```

libyara/include/yara/scanner.h
simonhf: Okay. I'll find another way to override the production code biz logic with test code biz logic at test run time. And hopefully the new method will make the clang error on Mac OS go away too :-)

simonhf: @plusvic Okay, so I left yr_scanner_scan_mem() intact so as not to change the API, and implemented the indirect call one level up. Finally I figured out how to appease clang running on Mac OS, and now the build and tests pass :-)

plusvic:
@simonhf I've done some modifications to your additions in #1363, please take a look and tell me if you are ok with the changes.

PR #1363, opened 2020-10-01
Review all calls to YR_DEBUG_FPRINTF and make the following changes:

plusvic:
* Use PRIx64 and PRIi64 for printing 64-bits values, which is more portable.
* Use %zu for printing size_t types.
* Remove the empty {} at the end of printed messages, as they don't provide that much information and clutter the format string
* Don't use ' in format strings for thousand separators as this is not very portable (https://stackoverflow.com/questions/10234766/warning-using-icc-and-printf-with-thousands-grouping-format-apostrophe)

### DIFF
--- a/libyara/proc.c
+++ b/libyara/proc.c
@@ -39,7 +39,7 @@ YR_API int yr_process_open_iterator(
     int pid,
     YR_MEMORY_BLOCK_ITERATOR* iterator)
 {
-  YR_DEBUG_FPRINTF(2, stderr, "+ %s(pid=%d) {}\n", __FUNCTION__, pid);
+  YR_DEBUG_FPRINTF(2, stderr, "+ %s(pid=%d)\n", __FUNCTION__, pid);
 
   YR_PROC_ITERATOR_CTX* context = (YR_PROC_ITERATOR_CTX*) \
       yr_malloc(sizeof(YR_PROC_ITERATOR_CTX));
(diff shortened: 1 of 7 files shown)


simonhf:
@plusvic, thanks for the various changes.

The only comment I would make is that there is / was a kind of strategy with the empty `{}`, which is that the instrumentation is designed to mimic simplified C function declaration syntax for easier human reading. The idea with `{}` is that it signifies both function entry and exit instrumentation, as well as any instrumentation in between in the same function.

Although the instrumentation I added was mostly empty `{}`, have a look at `_yr_scan_match_callback()` [1] which is a function which has separate entry and exit instrumentation, causing output like this:

```
+ _yr_scan_match_callback(match_data=.. match_length=..) { // match_offset=.. args->data=..
...
} // _yr_scan_match_callback() = <result>
```

Here the `{` and `}` are on separate lines and surround any other function instrumentation called in between.

A future instrumentation enhancement might be to indent or un-indent the lines depending upon the entry or exit instrumentation in a particular function running in a particular thread.

Anyway, that's the idea with the `{}`... I'll leave it up to you to decide if it's tolerable or not.

[1] https://github.com/VirusTotal/yara/blob/master/libyara/scan.c#L600

plusvic:
Ok, it makes sense, you I'll bring the curly braces back.

PR #1379, opened 2020-10-16
Classes (updated)

regeciovad:
The goal is to enable the use of rules like the following in the VirusTotal Hunting:
```
rule contains_btc_address 
{
    strings:
        $btc = /[13][a-km-zA-HJ-NP-Z1-9]{25,34}/ fullword ascii wide
    condition: 
        $btc
}
```
The implementation allows the use of multiple characters per byte in atoms and per state in the Aho-Corasick automaton. This is done via encoding in the bitmask, a part that is already used in YARA.

Some additional actions are needed. The Aho-Corasick has to be deterministic even for the calculation of failure links. 

However, simplification is shown in many areas. For example, the `nocase` option only involves the additional characters in the bitmask; new atoms are not created. The atoms contain more information about strings, and the Aho-Corasick automaton can be effective when used for searching.

The resulting transition table can be also smaller and more compact than if the same strings were generating multiple atoms.
For more details, I created a blog post about my changes and some test results: https://engineering.avast.io/yara-in-search-of-regular-expressions/.


libyara/ahocorasick.c
```diff
@@ -791,11 +1557,27 @@ int yr_ac_add_string(
...
     for (int i = 0; i < atom->atom.length; i++)
     {
-      YR_AC_STATE* next_state = _yr_ac_next_state(state, atom->atom.bytes[i]);
+      if (atom->atom.mask[i] == YR_ATOM_TYPE_CLASS)
+      {
+        // for a case like [a] - the class is a literal
+        int literal = _yr_ac_class_is_literal(atom->atom.bitmap[i]);
+        if (literal >= 0)
+        {
+          atom->atom.bytes[i] = literal;
```
plusvic, line 1566: With this change function `yr_ac_add_string` have the side effect of modifying the atoms that are passed to it in the list of atoms. This makes the logic harder to follow, because `yr_ac_add_string` is expected to modify the Aho-Corasick automaton, but not its input arguments like the string passed as `YR_STRING*` or the atoms passed as `YR_ATOM_LIST_ITEM*`.

I would say that this transformation from atoms of type `YR_ATOM_TYPE_CLASS` to `YR_ATOM_TYPE_LITERAL` should be done outside this function, at the place where the atoms are generated.

libyara/ahocorasick.c
```diff
@@ -209,19 +363,524 @@ static int _yr_ac_state_destroy(YR_AC_STATE* state)
...
+{
+  bool pop = true;
+
+  if (input_state->type == YR_ATOM_TYPE_LITERAL)
+    return pop;
+
+  int literal = _yr_ac_class_is_literal(input_state->bitmap);
+  if (literal >= 0)
+  {
+    input_state->input = literal;
```
plusvic, line 824: `input_state` is modified here, that's a side effect of `dfa_subtree` that should avoided. Why is this required here? Is it possible to convert all the states of type `YR_ATOM_TYPE_CLASS` that can be converted to `YR_ATOM_TYPE_LITERAL` in a single place?

libyara/ahocorasick.c
```diff
@@ -209,19 +363,524 @@ static int _yr_ac_state_destroy(YR_AC_STATE* state)
...
+  assert(counter != 0);
+
+  return non_literal;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+//
+// Detect if two states are in a collision
+//
+static bool _yr_ac_find_conflict_states(
```
plusvic, line 616: The description for this function says that it detects if two states are in collision, but it receives four states, what's the purpose of the other two states? Which are the states that are checked for conflict?

regeciovad:
Thank you for your review. I will look into a memory usage increase. My version generally takes a little bit of memory, but there should not be such a gap. I noticed that one of my changes from the past that decreased the memory usage (https://github.com/VirusTotal/yara/pull/1518) was reverted (https://github.com/VirusTotal/yara/blob/master/libyara/compiler.c#L269). I will check if this influences it. 
I will also update the code based on your comments and recommendations as soon as possible. Thank you for them.

libyara/re_grammar.y
```diff
@@ -428,6 +428,7 @@ single
         fail_if($$ == NULL, ERROR_INSUFFICIENT_MEMORY);
 
         $$->re_class = $1;
+        $$->mask = 0xCC;
```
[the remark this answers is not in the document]
regeciovad: I am sorry, I am updating my PR after a while, I had some deadlines, and I was also solving issues with this PR.
I changed the code based on this comment. However, I re-discovered my original motivation for this. For some reason, the type is not always correctly set up to `RE_NODE_CLASS` but `RE_NODE_LITERAL`.
This is causing missing matches on `tests/test-bitmap.c:328: rule does not match`.
The PR was, for that reason, failing tests for a few days. On other machines I have access to, the tests are OK.
I honestly have no idea why this is happening, and because I cannot recreate the failed test locally, I have trouble fixing the cause.

libyara/ahocorasick.c
```diff
@@ -132,26 +132,125 @@ static int _yr_ac_queue_is_empty(QUEUE* queue)
   return queue->head == NULL;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+//
+// Returns true if the bitmap1 and bitmap2 are the same.
+//
+static bool _yr_ac_compare_classes(YR_BITMASK* bitmap1, YR_BITMASK* bitmap2)
```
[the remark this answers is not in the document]
regeciovad: Updated.

libyara/ahocorasick.c
regeciovad: Updated.

libyara/ahocorasick.c
regeciovad: The change to the literal state is needed because it influences the algorithm. I tried to revise the algorithm several times, but I am mainly facing problems with match structure. It is sensitive to changes and copies, and many changes in the algorithm influence the order of matches, for example.

libyara/ahocorasick.c
regeciovad: I added some comments to make the process a little bit clearer, hopefully. When I am changing states, I have to update the matches and children's states as well. This is, however, not easy, with the goal of not changing the results, as I wrote in the previous comment. For that reason, I had to split it into several steps.

libyara/ahocorasick.c
```diff
@@ -209,19 +363,524 @@ static int _yr_ac_state_destroy(YR_AC_STATE* state)
...
+  else
+    return result;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+//
+// If the class becomes literal, it returns a literal found.
+// Otherwise, it returns 0.
+//
+static int _yr_ac_class_is_literal(YR_BITMASK* bitmap)
```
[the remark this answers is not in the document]
regeciovad: Updated.

libyara/re_grammar.y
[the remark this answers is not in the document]
regeciovad: Last comment regarding memory usage. I updated the `compiler.c` file with values decreasing memory usage. Hopefully, this will improve the issue you reported the last time.

PR #1381, opened 2020-10-21
feat: Compatibility to Windows XP

secDre4mer:
Don't use strnlen from the standard library on Windows as it is
not available on old systems like Windows XP. Instead, use the
OpenSSL version, or, if that's not available, create your own.
Similarly don't use strtok_s or _mkgmtime.

### DIFF
--- a/libyara/modules/pe/pe_utils.c
+++ b/libyara/modules/pe/pe_utils.c
@@ -254,7 +254,7 @@ int64_t pe_rva_to_offset(PE* pe, uint64_t rva)
 
 
 #if !HAVE_TIMEGM
-#if HAVE__MKGMTIME
+#if HAVE__MKGMTIME && !defined(__MINGW32__)
 #define timegm _mkgmtime
 #else
 


secDre4mer:
@googlebot I signed it!

hillu:
strnlen has been a standard C API for so long that I seriously doubt that it's unavailable on Windows XP. Can you provide a compiler or runtime error that leads you to think that strnlen can't be used?

secDre4mer:
You're right, @hillu . The strnlen part was an issue with another library that I incorrectly attributed to YARA.

The other two errors are (from my understanding) caused because I build with MinGW and MinGW links statically. As a result, YARA tries to link the DLLs in XP which don't contain those functions. Example error message is: The procedure entry point strok_s could not be located in the dynamic link library msvcrt.dll." and is shown by the OS upon trying to run the executable.

I'll rework the pull request substantially (removing the strnlen part and restricting the other changes to MinGW).

hillu:
Is `strtok_s` the only missing symbol? This could be fixed by replacing `strtok_s` calls in `populate_scan_list` with `strtok`. This ought to be unproblematic since it is called from `main`, from non-threaded code.

secDre4mer:
_mkgmtime is also missing in MinGW (see libyara/modules/pe/pe_utils.c). This isn't really hard to fix either, as you can see in my old commit.

hillu:
`_mkgmtime` does ring a bell. Does defining Define `__CRT__NO_INLINE`  on Win32.solve your problem?

See also https://github.com/spyre-project/spyre/commit/046cd4fca1b92d04b8fabe58e5a3d8b25f687c0a. Unfortunately I don't remember how I found this solution, it might have been by reading MinGW header files.

secDre4mer:
Yes, that apparently helped. Thanks for your support, @hillu ! 

Since apparently this is more of a MinGW issue, I'll withdraw the pull request.

Issue #1375, opened 2020-10-08
YARA newbie seeking best strategy for implementing asynchronous / incremental memory block access ?


simonhf:
I'm interested in tweaking the libyara API so that it can scan n arbitrarily large files concurrently in a single thread where the memory blocks for each file are available in order over time, but not all at once.

I know YARA supports threads, but I'm imagining n being 1,000 or 10,000 or 100,000 or more, which seems prohibitively large for numbers of threads, or? And at the same time, I'd prefer not to queue all those files on disk, etc.

Perhaps naively, this functionality seems within grasp since libyara is already designed to scan files as a sequence of consecutive overlapping discreet memory blocks, and already has a callback mechanism to fetch the next memory block, etc. The main difference seems to be that in the current libyara implementation, all the memory blocks needed are assumed to be available at the time libyara is called to scan the file. However, I'd like to change the functionality so that each call to libyara it scans what it can with the memory blocks available so far, but then can keep state and continue later when more memory blocks become available.

So I'm asking / brain-storming here as to whether anybody has ideas about the best way to approach implementing this functionality, as well as the challenges associated?

Having already dipped my toes in the libyara development swimming pool with a few recent PRs, this is my initial thoughts / ideas for how to approach the implementation of the functionality (please feel free to criticize, or suggest a different approach!):

Most functions in libyara return `ERROR_SUCCESS` [1] upon success, or one of the different `ERROR_*` errors / states. I'm thinking about creating a new return code called e.g. `ERROR_BLOCK_NOT_READY` which would be the libyara equivalent of the common Linux return codes `EAGAIN` or `EWOULDBLOCK`.

The libyara scanner will eventually callback to get the next memory block and that function might return `ERROR_BLOCK_NOT_READY` if the block isn't available.

Upon receiving `ERROR_BLOCK_NOT_READY` then all the functions in between would also return `ERROR_BLOCK_NOT_READY` back to the original caller of libyara.

The libyara caller can then wait until the next memory block is available, then call libyara again, and then the libyara scanner would immediately callback to get the next memory block until the next `ERROR_BLOCK_NOT_READY` or until the last memory block.

In theory this would be backward compatible with existing libyara API callers because their code would never have a next memory block callback returning `ERROR_BLOCK_NOT_READY`? And therefore, they'd never have to worry about that business logic?

[...]

PR #1387, opened 2020-11-06
Add ERROR_BLOCK_NOT_READY async support; part 1

simonhf:
See https://github.com/VirusTotal/yara/issues/1375#issuecomment-722241366

### DIFF
--- a/libyara/include/yara/types.h
+++ b/libyara/include/yara/types.h
@@ -666,6 +666,9 @@ struct YR_SCAN_CONTEXT
   // Pointer to user-provided data passed to the callback function.
   void* user_data;
 
+  // Pointer to user-provided data for the user (*_yr_scanner_scan_mem) function.
+  void* user_data_iterator;
+
   // Pointer to the user-provided callback function that is called when an
   // event occurs during the scan (a rule matching, a module being loaded, etc)
   YR_CALLBACK_FUNC callback;
@@ -716,12 +719,6 @@ struct YR_SCAN_CONTEXT
   // profiling_info is a pointer to an array of YR_PROFILING_INFO structures,
   // one per rule. Entry N has the profiling information for rule with index N.
   YR_PROFILING_INFO* profiling_info;
-
-  // Allocated iterator memory for the duration of the scan.
-  // Note: Stored here instead of stack for ERROR_BLOCK_NOT_READY functionality.
-  YR_MEMORY_BLOCK* addr_block;
-  YR_MEMORY_BLOCK_ITERATOR* addr_iterator;
-  YR_PROC_ITERATOR_CTX* addr_context;
 };
 
 union YR_VALUE
--- a/libyara/include/yara/types.h
+++ b/libyara/include/yara/types.h
@@ -203,6 +203,8 @@ typedef struct YR_MODIFIER YR_MODIFIER;
 
 typedef struct YR_ITERATOR YR_ITERATOR;
 
+typedef struct _YR_PROC_ITERATOR_CTX YR_PROC_ITERATOR_CTX;
+
 typedef uint32_t YR_AC_TRANSITION;
 
 #pragma pack(push)
@@ -709,6 +711,12 @@ struct YR_SCAN_CONTEXT
   // profiling_info is a pointer to an array of YR_PROFILING_INFO structures,
   // one per rule. Entry N has the profiling information for rule with index N.
   YR_PROFILING_INFO* profiling_info;
+
+  // Allocated iterator memory for the duration of the scan.
+  // Note: Stored here instead of stack for ERROR_BLOCK_NOT_READY functionality.
+  YR_MEMORY_BLOCK* addr_block;
+  YR_MEMORY_BLOCK_ITERATOR* addr_iterator;
+  YR_PROC_ITERATOR_CTX* addr_context;
 };
 
 union YR_VALUE
--- a/libyara/include/yara/types.h
+++ b/libyara/include/yara/types.h
@@ -634,6 +634,9 @@ struct YR_MEMORY_BLOCK_ITERATOR
   // case ERROR_SUCCESS        : No more blocks.
   // case ERROR_BLOCK_NOT_READY: Block not ready.
   int last_error;
+
+  // If buffer_size not known in advance, upon last block, iterator populates this buffer_size.
+  size_t buffer_size;
 };
 
 typedef int (*YR_CALLBACK_FUNC)(
--- a/libyara/include/yara/types.h
+++ b/libyara/include/yara/types.h
@@ -629,6 +629,11 @@ struct YR_MEMORY_BLOCK_ITERATOR
 
   YR_MEMORY_BLOCK_ITERATOR_FUNC first;
   YR_MEMORY_BLOCK_ITERATOR_FUNC next;
+
+  // If (next == NULL) and switch(last_error):
+  // case ERROR_SUCCESS        : No more blocks.
+  // case ERROR_BLOCK_NOT_READY: Block not ready.
+  int last_error;
 };
 
 typedef int (*YR_CALLBACK_FUNC)(
--- a/libyara/scanner.c
+++ b/libyara/scanner.c
@@ -256,6 +256,11 @@ YR_API int yr_scanner_create(YR_RULES* rules, YR_SCANNER** scanner)
     external++;
   }
 
+  new_scanner->matches_notebook = NULL;
+  new_scanner->addr_iterator    = NULL;
+  new_scanner->addr_context     = NULL;
+  new_scanner->addr_block       = NULL;
+
   *scanner = new_scanner;
 
   return ERROR_SUCCESS;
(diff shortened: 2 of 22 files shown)


libyara/scanner.c
```diff
@@ -466,59 +479,79 @@ YR_API int yr_scanner_scan_mem_blocks(
 
     block = iterator->next(iterator);
   }
+  // Come here if finished last block or no block currently available.
 
-  YR_TRYCATCH(
-      !(scanner->flags & SCAN_FLAGS_NO_TRYCATCH),
-      { result = yr_execute_code(scanner); },
-      { result = ERROR_COULD_NOT_MAP_FILE; });
+  YR_MEMORY_BLOCK* block_not_ready = iterator->first(iterator);
```
plusvic, line 484: This looks strange. Why calling `iterator->first()` here? Are you using `iterator->first()` for determining whether the previous call to  `iterator->next()` returned NULL because its the final block or because it was not ready?

If that's the case I think we need a better mechanism for knowing if `iterator->first()` and `iterator->next()` have reached the end of the block list or are indicating that the block is not available yet. Changing the signature of those functions so that they return a status code instead of a pointer to the block, or alternatively, adding a new field to the block that indicates if the block is actually ready or not.

plusvic (changes requested):

libyara/scanner.c
simonhf: Yes, I used `iterator->first()` to determine if the last call to `iterator->next()` returning `NULL` means 'last block', or 'block not ready'. I agree with you that this can be made better.

As a more intention revealing mechanism, I submit part 2 which extends `YR_MEMORY_BLOCK_ITERATOR` with `last_error` which is set to `ERROR_SUCCESS` if `iterator->next()` returning `NULL` means 'last block', or is set to `ERROR_BLOCK_NOT_READY` if `iterator->next()` returning `NULL` means 'block not ready'.

libyara/include/yara/types.h
```diff
@@ -709,6 +716,12 @@ struct YR_SCAN_CONTEXT
   // profiling_info is a pointer to an array of YR_PROFILING_INFO structures,
   // one per rule. Entry N has the profiling information for rule with index N.
   YR_PROFILING_INFO* profiling_info;
+
+  // Allocated iterator memory for the duration of the scan.
+  // Note: Stored here instead of stack for ERROR_BLOCK_NOT_READY functionality.
+  YR_MEMORY_BLOCK* addr_block;
```
plusvic, line 722: These three fields seem to be used only for test cases. Test cases should use only the public API, they are not privileged citizens, they should use the same features that are available for everyone else. `YR_SCAN_CONTEXT` is an opaque structure for API users, they should not rely on any fields in this structure, and this applies to test cases as well.

plusvic (changes requested):

libyara/scanner.c
plusvic: The solution in part 2 is much better.

libyara/include/yara/types.h
simonhf: Yes, that's correct, and thanks for noticing this.

So the issue is that currently the iterator data structures are allocated on the stack by (for tests) `_yr_test_single_or_multi_block_scan_mem()` [1] in `util.c`, or by (default but not supporting `ERROR_BLOCK_NOT_READY`) `__yr_scanner_scan_mem()` in `scanner.c`.

However, when using `ERROR_BLOCK_NOT_READY` then the data structures need to move to somewhere else, because `_yr_test_single_or_multi_block_scan_mem()`, `__yr_scanner_scan_mem()`, or equivalent user supplied function can be called repeatedly, and we want the iterator data structures to survive all that.

And I agree with your comments about `struct YR_SCAN_CONTEXT` being 'opaque structure for API users'. So I put my thinking hat on as to where the iterator data structures could be stored...

One idea is to put them in thread local storage declared global variables... but we've been down this path before and it's also not very elegant, and we have avoid using TLS so far. So assuming you also don't like the sound of this, I scratched it off the list.

An alternative idea: I noticed that `struct YR_SCAN_CONTEXT` already has an element called `user_data` for the user supplied callback. So I added a further element called `user_data_iterator`, and so that `struct YR_SCAN_CONTEXT` can remain an 'opaque structure for API users', I also added `yr_scanner_set_user_data_iterator()` and `yr_scanner_get_user_data_iterator()` to access it. This seemed to work nicely, except that I had to extend the `yr_rules_scan_xxx()` to take both `user_data` and `user_data_iterator`. And then I had to update a few test files that use that API.

What do you think? Or do you have a better idea?

[1] https://github.com/VirusTotal/yara/blob/master/tests/util.c#L188
[2] https://github.com/VirusTotal/yara/blob/master/libyara/scanner.c#L557

simonhf: Looks like the fuzz tests failed due the extra argument:
```
 + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -std=c++11 -I. /src/yara/tests/oss-fuzz/dotnet_fuzzer.cc -o /github/workspace/out/dotnet_fuzzer ./libyara/.libs/libyara.a -fsanitize=fuzzer
/src/yara/tests/oss-fuzz/dotnet_fuzzer.cc:76:3: error: no matching function for call to 'yr_rules_scan_mem'
  yr_rules_scan_mem(
  ^~~~~~~~~~~~~~~~~
/usr/local/include/yara/rules.h:81:12: note: candidate function not viable: requires 8 arguments, but 7 were provided
YR_API int yr_rules_scan_mem(
           ^
1 error generated.
```
I'm not sure how to run the fuzz tests myself (any instructions on how to do this?), but I just added the extra `NULL` arguments to various `.cc` files and hopefully part 4 will cause the fuzz tests to pass... fingers crossed.

plusvic: Looking at the code I've noticed that some of the `yr_scanner_xxx` functions are not easily retry-able after an `ERROR_BLOCK_NOT_READY`, for example `yr_scanner_scan_proc` calls `yr_process_open_iterator` each time it's called, and `yr_scanner_scan_file` maps and unmaps the file. If `yr_scanner_scan_file` returns `ERROR_BLOCK_NOT_READY` and it's retried multiple time that would involve mapping and unmapping the file multiple time which is far from ideal.

At the end, the only function that can be easily retried is `yr_scanner_scan_mem_blocks`, and that function already receives the pointer to `YR_MEMORY_BLOCK_ITERATOR` as an argument. So, the simplest solution is making `yr_scanner_scan_mem_blocks` the only function that can return `ERROR_BLOCK_NOT_READY` and can be retried after that. This puts the burden in the user of the API, as anyone who wants to use this new feature is forced to use `yr_scanner_scan_mem_blocks` which is relatively low level. However, I guess that anyone using the async approach is going to need to control it's own iterator anyways, so it looks like a reasonable solution to me.

If we make `yr_scanner_scan_mem_blocks` the only function that return `ERROR_BLOCK_NOT_READY` I think we don't need to make the `yr_scanner_scan_mem` function overridable anymore. The tests for this new case would be based on calls to `yr_scanner_scan_mem_blocks` with an iterator that provides custom `first` and `next` functions.

Maybe I'm missing something, but I think it could work.

simonhf: @plusvic thanks very much for the detailed comment and ideas!

*"for example ... `yr_scanner_scan_file` maps and unmaps the file. If `yr_scanner_scan_file` returns `ERROR_BLOCK_NOT_READY` and it's retried multiple time that would involve mapping and unmapping the file multiple time which is far from ideal."*

That's right, and I agree it doesn't make sense for `yr_scanner_scan_file` to use an iterator which may return `ERROR_BLOCK_NOT_READY` because the file being memory mapped is 'all there on disk' and so why would a block be 'not ready' after being memory mapped?

*"`yr_scanner_scan_proc` calls `yr_process_open_iterator` each time it's called"*

Likewise, `yr_scanner_scan_proc` attaches to a process which is 'all there in memory', so why would a block be 'not ready' after the process has been attached?

So the implementation areas where I see the new `ERROR_BLOCK_NOT_READY` feature being used would be e.g. in networking scenarios where files can be scanned block by block as they pass through the device. In these scenarios then the files never exist on disk, and executable files are never run or exist in memory to be attached to.

An example of a networking scenario where libyara could be baked in to scan blocks (read: packets, or blocks of a file extracted from packets) is Suricata, or other such tools for scanning / monitoring network traffic. In this historical case [1], Suricata initially 'rejected' libyara as a candidate but maybe that might change with the `ERROR_BLOCK_NOT_READY` feature?

*"If we make `yr_scanner_scan_mem_blocks` the only function that return `ERROR_BLOCK_NOT_READY` I think we don't need to make the `yr_scanner_scan_mem` function overridable anymore. ... Maybe I'm missing something, but I think it could work."*

I think it would work too, but... the only issue I see with this is regarding the testing added for testing (and effectively enforcing) the multiple blocks design feature of YARA in `test-rules.c` [2]. With the override-able  `yr_scanner_scan_mem` function then we can leverage the existing tests by running them again with a different iterator; the first time with a test iterator which returns a single large memory block (which is how the tests were originally designed) and also counts the number of memory blocks returned, and the second time with the test iterator returning multiple memory blocks and also counting the number of memory blocks returned. The total memory blocks from each loop are compared to prove that the different iterator mode has been used. Without being able to override the iterator creation in the `yr_scanner_scan_mem` function then we lose this extra test loop... Therefore, I think it would be worth keeping the multi block tests because:

 - They have already proved themselves useful by uncovering subtle multi block bugs.
 - They stop a multi block bug regression from happening in the future.

This table shows the compatibility if we keep the `yr_scanner_scan_mem` function override-able (else, the `Yes*` items will turn into `No`):
```
|                          iterator returns: | single | multi  | ERROR_BLOCK_NOT_READY |
|                                            | block  | blocks |                       |
|--------------------------------------------|--------|--------|-----------------------|
|               default iterator compatible? | Yes    | No     | No                    |
| test-*.c (tests not listed)    compatible? | Yes    | No     | No                    |
| test-rules.c                   compatible? | Yes    | Yes*   | No                    |
| yr_rules_scan_*                compatible? | Yes    | Yes    | No                    |
| yr_scanner_scan_(fd|file|proc) compatible? | Yes    | Yes    | No                    |
|                  test iterator compatible? | Yes    | Yes    | Yes                   |
| test-async.c                   compatible? | Yes    | Yes    | Yes                   |
| yr_scanner_scan_mem            compatible? | Yes    | Yes    | Yes*                  |
| yr_scanner_scan_mem_blocks     compatible? | Yes    | Yes    | Yes                   |
```
If you are okay with the idea of nuking the larger number of multi block tests in `test-rules.c` then I can rewind the override for the `yr_scanner_scan_mem` function. Otherwise I suggest keeping the override... at least for the time being. What are your thoughts?

[1] https://redmine.openinfosecfoundation.org/issues/1828
[2] https://github.com/VirusTotal/yara/blob/master/tests/test-rules.c#L2857

simonhf: @plusvic I also added part 5 which contains code for the dynamic file_size; the iterator can pass the file_size to the scanner when the last block is reached.

However, a couple of the Travis builds failed. It's not obvious to me why looking at the logs. Any idea what failed? Is the CI system 'flappy'?

plusvic: I'm OK with overriding `yr_scanner_scan_mem` for the tests, but I think a better compromise is moving all the overriding logic to the tests themselves instead of including it in libyara. I mean, instead of calling `yr_scanner_scan_mem` from the tests we can use a pointer to the original `yr_scanner_scan_mem` or to some replacement function as you already do. Calls to `yr_scanner_scan_mem` in functions like `matches_blob` (tests/utils.c) would go through the pointer and therefore could be redirected to the replacement function.

The problem with this is that the replacement function will be called only in cases where the tests are calling `yr_scanner_scan_mem` directly, because at those points we can start using the pointer, but calls to any other function like `yr_rules_scan_xxx` will go through the original `yr_scanner_scan_mem`.

This shouldn't be a major issue, since all the tests in test-rules.c (except for `test_process_scan`) are using macros that ultimately call `matches_blob`. So, all tests in test-rules.c (and also tests for modules like pe, elf, dotnet, etc)  could be easily run with a multi-block setup.

With this approach we keep the overriding of `yr_scanner_scan_mem` confined to the tests.

Regarding the CI fails, they seem unrelated to your changes, I'll take a look.

simonhf: Thanks for the further guidance.

If I understand you correctly then I will refactor so that the following is expected:

 - Remove `(*_yr_scanner_scan_mem)` from libyara and restore `yr_scanner_scan_mem` function.
 - Default `yr_scanner_scan_mem` function uses default iterator which handles neither multiple blocks nor `ERROR_BLOCK_NOT_READY`.
 - `test-async.c` works around `(*_yr_scanner_scan_mem)` by initially calling neither `yr_rules_scan_mem` nor `yr_scanner_scan_mem`, instead calling it's own test-only incantation of `yr_scanner_scan_mem` which calls `yr_scanner_scan_mem_blocks` directly.
 - `test-rules.c` continues to work with multi block testing by modifying the test `matches_blob` function by initially calling neither `yr_rules_scan_mem` nor `yr_scanner_scan_mem`, instead calling it's own test-only incantation of `yr_scanner_scan_mem` which calls `yr_scanner_scan_mem_blocks` directly.
 - Other test files also using the test `matches_blob` function will also work the same way but will default to single block testing.

I've started work on this refactor, and please confirm in the meantime that this is what you are expecting :-)

plusvic: Yes, that's exactly what I meant.

simonhf: Because of the extensive refactor then I submitted the changes as a new PR: https://github.com/VirusTotal/yara/pull/1391

plusvic:
Work continued in the `non-blocking-api` branch.

Issue #1384, opened 2020-11-03
Scanning a large process on Windows uses large amounts of memory


secDre4mer:
On Windows, scanning a process takes an amount of memory that is (roughly) proportional to the amount of memory the process uses. E.g., if a process has allocated 1GB on its heap, a YARA scan of this process will take another GB.
In contrast to this, during file scanning, YARA's memory usage is independent of the scanned file's size.

From my understanding of the YARA implementation, this is because YARA iterates over all regions in the target process (using VirtualQueryEx to query the region information) and each region is used as a singular memory block. The implementation details for this can be found in [yr_process_get_next_memory_block](https://github.com/VirusTotal/yara/blob/master/libyara/proc/windows.c#L134).
The highest amount of memory used by this approach is the largest region in the target process. 

As a possible solution, large regions could be split into multiple memory blocks with a maximum block size. This does open up the question, though, what a good "maximal block size" would be for this purpose. Also, I'm not familiar enough with the scanner's internal architecture to know whether this has any unintended side effects.

secDre4mer:
I've implemented the suggestion here: https://github.com/secDre4mer/yara/commit/e6464c5d4db4b483a05651f0e2f2b39fdc5b41c4
It's limited to Windows so far, and only tested rudimentarily.

plusvic:
My only concern with this is that it can lead to false negatives if the pattern you are looking for with you rule spans a block boundary. The risk is small if you use a large enough block size, and that's the price to pay for not using too much memory, but we must document well how it works. Also, putting a minimum acceptable block size that is large enough would. I wouldn't got with blocks smaller than 50MB or so.

secDre4mer:
Yes, agreed. I noticed that as well in my tests, but as you said, the risk becomes sufficiently small with large block sizes. The PR sets the default value to 1 GB, and users with less resources can then reduce it (with the resulting higher risk of false negatives).

plusvic:
Do you plan to implement it for the remaining platforms?

secDre4mer:
The pull request already includes commits that add the changes for the remaining platforms. I haven't tested FreeBSD or OpenBSD yet (I'm currently setting up a FreeBSD VM for the former), the others worked in my tests.

secDre4mer:
I've tested FreeBSD and OpenBSD now as well (and fixed a bug from my changes for both).

PR #1393, opened 2020-11-19
Add maximum size for process memory chunk

secDre4mer:
Adds a new configuration parameter that limits the size of process memory chunks read by YARA.

This handles the problem described in https://github.com/VirusTotal/yara/issues/1384. Since YARA previously allocated a full "sector" with equal permissions / from the same mapping, YARA could use an unlimited amount of RAM (especially when scanning processes with memory mapped files) without the caller having the ability to control this.


secDre4mer:
Experimental change so far - especially the OpenBSD / FreeBSD parts are untested so far.

secDre4mer:
@plusvic The fuzzing build is failing because the build environment apparently is missing YACC. Do you have an idea what caused this PR to trigger YACC when it usually doesn't?

plusvic:
I think it has to do with the timestamp of the source files or something like that. Sometimes it thinks that the `.c` files must be regenerated from the `.y` and `.l` files, while at some other times it thinks that the `.c` is up-to-date. I didn't look at it in detail, but that's my educated guess.

secDre4mer:
I see. So the correct solution would probably be to add YACC and other required tools to the fuzzer workflow. I'm not sure how to do that, though - https://github.com/VirusTotal/yara/blob/master/.github/workflows/main.yml contains the configuration, but since we run in a docker container hosted in another repository, we can't easily adjust it. 
Which is a bit annoying since this seems to be a technical issue and not something related to the pull request.

secDre4mer:
@plusvic The build issue no longer occurs, so I'd be grateful for any review.

mat-gas:
any update on this feature? it could be quite useful indeed..

and to silently benefit from t, why not set the defalt chunk size to 50 or 100 MB now instead of 1 GB? (where most users in most cases wouldn't see any difference?)

Neo23x0:
I am also interested to see this merged. We currently apply @secDre4mer's solution as a patch in our build process.

Issue #1394, opened 2020-11-20
elf symtab_entries


lprat:
Hi,
I tried to use "symtab_entries" on elf, but not work on version 4.0.2.
Can you help me to find why my rule dont work?

My rules:
```
import "elf"
rule test1 { 
        condition: 
          elf.symtab_entries > 0
}
rule test2 { 
        condition: 
        for any i in (0..elf.symtab_entries):
        (
            elf.symtab[i].bind == elf.STB_GLOBAL
        )
}
```
Tried to use but dont work:
```
$ yara -v
4.0.2
$ readelf -Ws /bin/ls

La table de symboles « .dynsym » contient 128 entrées :
   Num:    Valeur         Tail Type    Lien   Vis      Ndx Nom
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __ctype_toupper_loc@GLIBC_2.3 (2)
     2: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND getenv@GLIBC_2.2.5 (3)
     3: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fgetfilecon@LIBSELINUX_1.0 (4)
     4: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND sigprocmask@GLIBC_2.2.5 (3)
     5: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __snprintf_chk@GLIBC_2.3.4 (5)
     6: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND raise@GLIBC_2.2.5 (3)
     7: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND abort@GLIBC_2.2.5 (3)
$yara rules.yar /bin/ls
$yara -D ../rules.yar /bin/ls
elf
	symtab
	symtab_entries = YR_UNDEFINED
	dynamic
		[0]
			type = 1
			val = 1359
		[1]
			type = 1
			val = 1375
		[2]
			type = 12
			val = 16384
		[3]
			type = 13
			val = 98720
		[4]
			type = 25
			val = 144208
		[5]
			type = 27
			val = 8
		[6]
			type = 26
			val = 144216
		[7]
			type = 28
			val = 8
		[8]
			type = 1879047925
			val = 776
		[9]
			type = 5
			val = 4024
		[10]
			type = 6
[...]

PR #1395, opened 2020-11-20
fix issue 1394

lprat:
Fix issue: https://github.com/VirusTotal/yara/issues/1394

### DIFF
--- a/libyara/modules/elf/elf.c
+++ b/libyara/modules/elf/elf.c
@@ -306,7 +306,8 @@ static const char* str_table_entry(
             set_string(section_name, elf_obj, "sections[%i].name", i);                    \
         }                                                                                 \
                                                                                           \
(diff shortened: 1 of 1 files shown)


virusdefender:
and it seems `symtab_entries` is missing on `elf. ET_DYN ` files as well

plusvic:
`symtab` and `dynsym` are two separate tables, should we have two separate tables in the ELF module as well? I'm not sure what's the better approach, but I'm wondering what happens if an ELF has both `symtab` and  `dynsym`, it that possible? If yes, do they have the same symbols? Wouldn't be confusing to put symbols from `dynsym` in a field named `symtab`?

plusvic:
Closing this PR as #1427 looks like a better alternative.

plusvic:
Implemented in #1427

PR #1398, opened 2020-11-26
Raise warning not only with .* and .+, but with .{x,} as well.

### LINKED ISSUE #1386: Inconsistent "reasonable value for N" warnings using regex
When using regex, warnings are displayed when using the `.*` or `.*?` syntax.  However, a similar warning is not displayed when using the` .{0,}` or `.{0,}?` syntax.  As these queries essentially mean the same thing, I would expect the warnings to be consistent as well.

plusvic:
Closes #1386.


libyara/re.c
```diff
@@ -365,9 +368,15 @@ int _yr_re_node_contains_dot_star(RE_NODE* re_node)
   return false;
 }
 
-int yr_re_ast_contains_dot_star(RE_AST* re_ast)
+////////////////////////////////////////////////////////////////////////////////
+// Detects the use of .*, .+ or .{x,} in a regexp. The use of wildcards with
+// quantifiers that don't have a reasonably small upper bound causes a
+// performance penalty. This function dectects such cases in orde to warn the
```
wxsBSD, line 374: Tiny typo here: s/orde/order/

plusvic: Fixed

PR #1400, opened 2020-11-26
Set a hard limit to the number of resources collected in the `pe.reso…

plusvic:
…urces` array.

Without this limit some corrupted files where causing the module to use more than 1GB of memory. One example is:
tests/oss-fuzz/pe_fuzzer_corpus/clusterfuzz-testcase-minimized-pe_fuzzer-5671228022718464


wxsBSD (approved):
That's a lot of resources. ;)

PR #1403, opened 2020-11-30
Implement non-blocking scan API.

plusvic:


plusvic:
Work continued in the `non-blocking-api` branch.

PR #1411, opened 2020-12-02
Don't use environment variables for controlling block size in tests.

plusvic:
If yr_test_mem_block_size is changed to something else test-async breaks, so being able to change this value with an environment variable is not very useful if the tests can't adapt to different sizes.


plusvic:
@simonhf please take a look.

PR #1421, opened 2020-12-20
pe: add signature[i].algorithm_oid

hillu:
Certificate signature algorithms are encoded using object identifiers
whose text representations depend on the cryptographic library. For
stable YARA rules, it is wiser to rely on the OID, so let's provide
it.

For example, while OpenSSL 1.1.1d resolves 1.2.840.113549.1.1.11 to
"sha256WithRSAEncryption" (or "RSA-SHA256" for the short name), the
only string representation I have gotten out of the Windows CryptoAPI
is "sha256RSA".

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -746,7 +746,26 @@ Reference
 
     .. c:member:: algorithm
 
-        Algorithm used for this signature. Usually "sha1WithRSAEncryption".
+        String representation of the algorithm used for this
+	signature. Usually "sha1WithRSAEncryption". It depends on the
+	X.509 and PKCS#7 implementationss and possibly their versions,
+	consider using algorithm_oid instead.
+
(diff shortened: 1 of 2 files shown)


plusvic:
Could you add it to the documentation as well? Explaining why this new field is necessary as you did in the commit description would be useful for people reading the documentation.

wxsBSD:
This has implications on any work to migrate away from OpenSSL as any rules using these OpenSSL specific names will be broken if the new library doesn't report the same thing.

hillu:
> This has implications on any work to migrate away from OpenSSL as any rules using these OpenSSL specific names will be broken if the new library doesn't report the same thing.

Usage of OpenSSL's `X509_NAME_oneline` to produce `subject` and `issuer` fields has related implications. Alternative implementations should mimic its behavior.

PR #1423, opened 2020-12-20
RfC: PE Authenticode parsing support using Windows Crypto API

hillu:
Here's some code for a not-yet-quite-complete implementation of the PKCS7-specific bits of `pe_parse_certificates`. Once this is done, it might allow for full-featured YARA builds on Windows without the OpenSSL dependency.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -279,6 +279,7 @@ AS_IF([test "x$have_crypto" = "xno"],
 	          AC_MSG_RESULT([The "hash" module functions will be provided through the Microsoft Crypto API])
             # FIXME: Add PC_LIBS_PRIVATE entries?
             build_hash_module=true
+	    LIBS="$LIBS -lcrypt32"
           ],
           [],
           [#include <windows.h>])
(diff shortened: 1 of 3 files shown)


wxsBSD:
I haven't looked at the code but do want to point out that this may result in a differential between a windows version and non-windows that can be difficult to debug. Having one consistent library for parsing these will ensure that rules will be applied consistently.

Just something to think about...

hillu:
> this may result in a differential between a windows version and non-windows that can be difficult to debug.

I agree. If you have a look at the FIXME comment and the changed test, you'll see that there already are some slight differences in the output. Even for string representations, we currently rely entirely on OpenSSL's behavior. (e.g. "sha256WithRSAEncryption" where wincrypt would just give us "sha256RSA").

Giving users the option to get rid of the OpenSSL dependency altogether is still a worthwhile goal, in my opinion.

hillu:
Since Avast's `authenticode-parser` was merged, my patch no longer applies. Closing the PR.

PR #1426, opened 2020-12-22
Add convention checker for clang format, fix remaining source files, …

simonhf:
…and fix newer gcc error

PR #1429, opened 2020-12-23
Add clang format convention check, fix source files, & fix gcc error

simonhf:
New PR version of https://github.com/VirusTotal/yara/pull/1426 due to updated non-blocking-api branch.


simonhf:
@plusvic I don't think the CI build is failing due to a change I made, or?

PR #1430, opened 2020-12-25
Implement "percent" keyword for string sets.

wxsBSD:
This commit implements the "percent" keyword and extends the grammar to
recognize it for string sets. I had initially planned on just using the '%'
symbol but that was causing some ambiguities in the grammar so I settled on
defining a new keyword.

The grammar now recognizes "50 percent of them" or variations on that. The
percentage number must be between 1 and 100 (inclusive) or the compiler will
return an error.

The implementation uses a new opcode (OP_OF_PERCENT) and this is because the
stack for these looks different from a typical "for_expression _OF_ string set"
rule. In the "for_expression" case (which uses OP_OF) the stack looks like this
(assuming two strings $a and $b):

$a
$b
undefined <-- end of list
2 <-- for_expression

The OP_OF operation checks each of the strings until it hits the end of list and
then compares the number found to whatever is left on the stack (2 in this
case).

The new grammar causes a stack at runtime which looks like this:

50 <-- percent number
$b
$a
undefined <-- end of list

The OP_OF operator won't work in this case because the percent number is sitting
on the top of the stack compared to the for_expression sitting at the bottom of
the stack.

I also cleaned up a minor oversight in compiler.c when printing the
ERROR_INVALID_MODIFIER message. It is the same as the other compiler errors so I
just moved it into the same set of fallthrough case statements.

Lastly, add some test cases to make sure this is working as intended.


wxsBSD:
I'll update documentation to reflect the new keyword and usage of it if you are happy with this approach.

wxsBSD:
Closing this as the work in #1434 is better.

PR #1434, opened 2021-01-04
Replace the 'percent' keyword with '%'

plusvic:
Also makes the percentage a primary expression instead of number, which solves the grammar conflict and has the additional benefit that it opens the door for using variables as percentages. For example:

```
rule test {
strings:
   ...
condition:
  x% of them.  // x is a variable with an externally passed value.
}
```

There's a pending change, though. In order to allow external variables we must remove the check that ensures that the percentage is between 1 and 100, as we don't know the value of the variable at compilation time.


wxsBSD:
This is a much better approach and was one I was looking at when I did the initial work in my PR. I decided against it because I managed to convince myself it would open up weird situations like "pe.number_of_sections% of them" which makes no sense. Turns out I was wrong and this isn't possible.

Overall, I think this is fine. I'm a little unsure what removing the check for a value between 1 and 100 would mean though. In theory I could do something like `125% of them` or `-1% of them`. Personally, I think that allowing people to express unrealistic percentages is worth the ability to specify an external variable.

plusvic:
With the last commit you can use a variable in `x% of ...`. If `x` is a variable we don't know its value at compiling time, but if `x` is a constant we make sure that it is in the range [1,100].

wxsBSD (approved):

libyara/exec.c
```diff
@@ -1304,10 +1304,28 @@ int yr_execute_code(YR_SCAN_CONTEXT* context)
...
+      }
+      else  // OP_OF_PERCENT
+      {
+        YR_DEBUG_FPRINTF(
+            2, stderr, "- case OP_OF_PERCENT: // %s()\n", __FUNCTION__);
+
+        // If, by some weird reason, we manage to get an undefined string
+        // reference as the first thing on the stack then count would be zero.
+        // I don't know how this could ever happen but better to check for it.
+        if (count == 0)
```
wxsBSD, line 1324: Probably want to also check for "r2.i == YR_UNDEFINED" here too. If we do something like `pe.number_of_sections% of them" and `pe.number_of_sections` is undefined then this will evaluate to true:

```
wxs@wxs-mbp yara % cat rules/test.yara
import "pe"

rule a {
  strings:
    $b = "foo"
  condition:
    pe.number_of_sections% of them
}
wxs@wxs-mbp yara % ./yara -s rules/test.yara /bin/ls
a /bin/ls
wxs@wxs-mbp yara %
```

wxsBSD (changes requested):

wxsBSD (approved):
Looks good to me!

PR #1441, opened 2021-02-01
Clarify that pe.imphash() requires a lowercase string

runasand:
Updated to reflect that pe.imphash() requires a lowercase string, just like the Hash module.


runasand:
@googlebot I signed it!

plusvic (approved):

Issue #1443, opened 2021-02-04
Incorrect WIN_CERTIFICATE alignment in 'pe_parse_certificates' function


Aleh-Kazakevich:
When parsing PE certificate table, the pointer to second and other WIN_CERTIFICATE structure
may not be 8-byte aligned, as required by specification:

`win_cert = (PWIN_CERTIFICATE)(end + (end % 8));`

Example:
end = 0x1234000A, win_cert = 0x1234000c (not 8-byte aligned!)

See the 'pe_parse_certificates' function in the 'pe.c' file
https://github.com/VirusTotal/yara/blob/master/libyara/modules/pe/pe.c#L1507
https://github.com/VirusTotal/yara/blob/master/libyara/modules/pe/pe.c#L1526

plusvic:
If I understand correctly you mean that there are certain PE file that don't follow the specification and don't have the certificate aligned as it should be. Is that what you mean?

Can you sent some of those files that we can use as a reference or for including it as a test case?

Aleh-Kazakevich:
No, the problem is that the 'win_cert' variable itself is not 8-byte-aligned when moving to the next 'WIN_CERTIFICATE' structure.

plusvic:
Oh, I see what you mean. The in way in which the computation is done is incorrect. The correct one would be:

```
win_cert = (PWIN_CERTIFICATE)((end + 7) & -8);
```

plusvic:
Fixed in e4819f6b98ebdd97eff9e32badb6edd519e76dc3

PR #1445, opened 2021-02-04
Fix issue #1443.

plusvic:
As pointed out in the bug report the correct way of aligning the win_cert pointer to the next 8-bytes boundary is not correct. If the previous certificates ends at 0x1234000A, the next one should start at 0x12340010, not at 0x1234000c which is not aligned to 8 bytes.


plusvic:
@wxsBSD do you have some file with multiple certificates that could be used for a test case?

wxsBSD:
Let me do some digging tonight and see what I can find.

wxsBSD:
7c3f863cbff4db18689d6f0dde31e0f1c320763a30c802490709e166c859d39b has two WIN_CERTIFICATES in it. The first one starts at 0x150ae0 and the second one starts at 0x152ce0.

wxsBSD (approved):

wxsBSD:
This doesn't seem to affect that one test case I mentioned, but I think it should be right.

PR #1450, opened 2021-02-16
add math.tonumber function to convert bool to number

virusdefender:
more data types can be supported if needed, for example
string to number, etc.


virusdefender:
@googlebot I signed it!

PR #1453, opened 2021-02-21
Update yara.c (skip symlinks to . and show 64 instead of 32 hex chars)

2d4d:
1. skip all symlinks pointing to . so you can now do
```bash
yara -r test.yar /usr/bin
```
on debian without figuring out why it doesn't end, figuring out the problem and using -N (/usr/bin/X11 points to . )

2. Show 64 instead of 32 hex chars (regex are shown in full so there's enough clutter anyway)

### DIFF
--- a/cli/yara.c
+++ b/cli/yara.c
@@ -566,6 +566,21 @@ static void scan_dir(
         de = readdir(dp);
         continue;
       }
+      else if (S_ISLNK(st.st_mode))
+      {
+        // always skip symlinks that point to . like /usr/bin/X11 -> . on debian
+        char buf[2]="";
+        int len;
+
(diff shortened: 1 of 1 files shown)


2d4d:
something broken with fuzzing?

Path does not exist /home/runner/work/yara/yara/out/artifacts | Fuzzing:                     .github#L1
-- | --

Exit code 1 returned from process: file name  '/home/runner/runners/2.277.1/bin/Runner.PluginHost', arguments 'action  "GitHub.Runner.Plugins.Artifact.PublishArtifact, Runner.Plugins"'. | Fuzzing:                     .github#L1
-- | --

PR #1470, opened 2021-03-22
Reduce memory pressure caused by Linux process scanning

hillu:
Reading memory pages from /proc/$PID/mem has the unpleasant side                                                                            
effect of forcing otherwise unused process memory pages into the                                                                            
target process VM, considerably increasing the target process'                                                                              
resident set size.                                                                                                                          
                                                                                                                                            
As a typical example, many programs only make use of a subset of code                                                                       
from shared libraries that they are linked against. The solution                                                                            
consists of replicating memory mappings mappings from files if they                                                                         
are direcly available through the filesystem. All information needed                                                                        
to find the file (and to determine if it is indeed identical to the                                                                         
file mapped by the target process) is available from /proc/$PID/maps.                                                                       
Only pages that may have been changed according to /proc/$PID/pagemap                                                                       
are then read via /proc/$PID/mem to overwrite the local mapping.                                                                            
                                                                                                                                            
Mappings whose underlying filesystem object has changed read from                                                                           
/proc/$PID/mem as before.                                                                                                                   
                                                                                                                                            
Mappings that are not backed by a regular file (block devices, stack                                                                        
or heap sections, etc.) are assumed to behave like zeroed-out files                                                                         
and otherwise treated the same.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -54,6 +54,9 @@ yarac_SOURCES = \
 yarac_LDADD = -Llibyara/.libs -lyara
 yarac_DEPENDENCIES = libyara/.libs/libyara.la
 
+tests_mapper_SOURCES = tests/mapper.c
+tests_mapper_CFLAGS = -O0
+
 test_alignment_SOURCES = tests/test-alignment.c tests/util.c
 test_alignment_LDADD = libyara/.libs/libyara.a
 test_arena_SOURCES = tests/test-arena.c tests/util.c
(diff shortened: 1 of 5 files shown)


hillu:
Oh well, this seems to bomb gloriously in Travis with Address Sanitizer enabled. This crashed using the original implementation, too, though.

hillu:
And now the Bazel build fails for some reason I don't understand.

hillu:
Bazel build fixed. :-)

ozanh:
Is there any progress about this PR? This PR seems to be the only current solution to scan Linux processes without a problem or a crash or a memory exhaustion.

plusvic (approved):

plusvic (approved):

plusvic:
Version v4.2.0-rc1 is coming with this change included, I hope it will be released in a few weeks.

PR #1472, opened 2021-03-25
Linux proc scanning: Read entire lines from /proc/$PID/maps

hillu:
Entries in /proc/$PID/maps may contain long filenames. Skip over those
if they haven't been fully read.


plusvic:
How long are those entries? Increasing the buffer's size would help?

hillu:
The last component is a filename. `_POSIX_PATH_MAX` is 4096. Adding ~100 bytes for the data before the filename (and a "(deleted)" entry should be enough for **almost** all cases. Quoting of special characters in the filename could increase the lenght past 4095 bytes.

plusvic (approved):

PR #1478, opened 2021-04-08
fix: undeclared variable c

phantinuss:
I was trying to compile the current master on linux and it failed because of an undeclared variable c. I fixed the issue in the loop, without further understanding what the loop actually does. yara now builds again without errors.


phantinuss:
@googlebot I signed it!

phantinuss:
unneed double assignment

PR #1483, opened 2021-04-26
Add additional functions to math module

secDre4mer:
Add further basic statistics to math:
- Mode of file or area
- String / byte counting
- Calculating the percentage of a byte in a given file or area


Neo23x0:
@googlebot I fixed it.

c3rb3ru5d3d53c:
Looking forward to using this! :smile:

docs/modules/math.rst
```diff
@@ -118,3 +118,44 @@ file and create signatures based on those results.
...
+
+    .. versionadded:: 4.2.0
+
+    Returns how often a specific byte or substring occurs, starting at *offset*
+    and looking at the next *size* bytes. When scanning a
+    running process the *offset* argument should be a virtual address within
+    the process address space.
+    *offset* and *size* are optional; if left empty, the complete file is searched.
+
+    *Example: math.count("$[]", 0, 100) >= 5*
```
plusvic, line 132: The `count` variant receiving a byte is ok to me, but I have concerns about the variant that accepts a string. The concern is that this is very inefficient way of counting the number of occurrences for a string. I'm aware that YARA doesn't offer a way for counting the number of occurrences for a string in a certain offset range, but if that functionality is required it could be implemented in some other way.

Also `math.count("foo")` is redundant, as this can be already done by other means in a more efficient way. In general I think that this variant encourages the use of this function with whole files, and that's not something I would recommend. Even for functions like `hash.sha256()` we require explicit offset and size.

secDre4mer: > Also math.count("foo") is redundant, as this can be already done by other means in a more efficient way.

Done, I've deleted it.

> The concern is that this is very inefficient way of counting the number of occurrences for a string. I'm aware that YARA doesn't offer a way for counting the number of occurrences for a string in a certain offset range, but if that functionality is required it could be implemented in some other way.

The main use case we had in mind for this function was counting a short and relatively common string that is only required in specific circumstances, e.g. to check for obfuscation. Since the string is only required in comparatively rare cases, adding it as a normal YARA string would slow down performance globally, whereas with this function it could be evaluated only in specific cases.
Quick and dirty example:
`uint16(0) == 0x253c and filesize < 10KB and math.count("$[", 0, filesize) > 500`

I agree that the current implementation is inefficient. I've reworked it somewhat and I'll gladly take hints or references for further optimization.

plusvic: I think I'm fine with this as long as it doesn't encourage users to use the new feature without thinking twice about the implications. The latest changes are probably enough.

plusvic: @secDre4mer, after @wxsBSD's changes the functions that receive strings can be removed and leave only those that receive a single byte.

secDre4mer: Alright, I'm not 100% convinced that there's no use case for the string count methods (see my example above), but I can understand that it's better to provide a single interface for string counting rather than adding a functional duplicate that works slightly different and is faster only in very specific cases.

Neo23x0: I think, with only a single byte, we would rewrite this condition 
```
uint16(0) == 0x253c and filesize < 10KB and math.count("$[", 0, filesize) > 500
```
to
```
uint16(0) == 0x253c and filesize < 10KB and math.count("$", 0, filesize) > 500 and math.count("[", 0, filesize) > 500
```
which is more fuzzy but could still work on obfuscated payloads.

plusvic:
It looks like the test cases for the `math` module are not passing.

secDre4mer:
Sorry for that, @plusvic, I simply forgot removing the test cases for the string version of count(). Done.

secDre4mer:
@wxsBSD something about the `of ... in` syntax has been bugging me: What would be the equivalent of `math.count("AA", 0, 100) > 2`?

The obvious solution, `2 of $s1 in (0..100)` will never be true because only a single occurrence of $s1 is counted at https://github.com/VirusTotal/yara/blob/83acb204ed9a87c4f5804d1d7f4f7d79b975dba0/libyara/exec.c#L1386

Is there something I am missing here?

plusvic:
Certainly the semantics are different.  The new syntax is actually  `N of <string_set> in <range>`, which means that at least N strings from `<string_set>` are found in `<range>`. Which is different from "at least N occurrences of a single string S in <range>".

`2 of $s1 in (0..100)` is not even valid, the correct syntax would be `2 of ($s1) in (0..100)`. But this is not covering your use case. This would require yet another kind of statement, like:

```
N occurrences of $s1 in (0..100)
```

secDre4mer:
Alright, I think I understand now. Thanks for the explanation!

plusvic:
I realized it just now. I didn't noticed the differences between the new operator and what you really wanted until now.

secDre4mer:
True, there's some difference, but as @Neo23x0 mentioned above, the byte count function is sufficient for most cases. 

I'd prefer to focus on merging that for now and discuss the idea of `N occurrences of $s1 in (0..100)` somewhere else.

wxsBSD:
> @wxsBSD something about the `of ... in` syntax has been bugging me: What would be the equivalent of `math.count("AA", 0, 100) >= 2`?

As @plusvic mentioned, you are completely correct that the new operator does not cover your use case (but it does cover a good use case, so I'm still happy it was done). As I understand your use case you are looking for the total number of occurrences of a string in a range.

I'm proposing `#s1 in (0..100) >= 2` as a way to express what you want. We would walk all the matches for `$s1` and total up the number of matches in the given range. I believe this syntax would cover your use case well, while being concise and fit in nicely with similar constructs (I'm specifically thinking of `$a in (0..100)` here).

If there is an interest in using this syntax, and @plusvic agrees, I will definitely add it.

One thing this explicitly does not cover is the idea of `N occurrences of <string set> in (0..100)` because there are some edge cases there to consider. For example, consider this possible syntax:

`2 of (#s*) in (0..100)`

I believe this is a good syntax for expressing this logic. The edge case that needs to be considered is what happens when there are `$s0` and `$s1`, where each of them match once. Would you expect the logic to be:

1. The total number of occurrences of strings from this set to be at least 2.
2. The number of occurrences of any of the strings from this set to be at least 2.

In the situation I'm describing, option 1 would match while option 2 would not.

I'm happy to leave the problem applying this logic to string sets for a future diff, but it is something I want to throw out there for consideration as I think it is the next logical step down this path.

plusvic:
`#s1 in (0...100) >= 2` sounds good, much better than `2  occurrences of $s1 in (0..100)`, and very coherent with other constructs. 

I would keep `N occurrences of <string set> in (0..100)` out of the discussion for now. However, something like this can could work:

`#s*` -> number of aggregated matches for all strings that start with `s`
`#s* in (0..100)` -> number of aggregated matches in range (0.100) for all strings that start with `s`

This doesn't cover all kinds of string_sets, but provides some flexibility.

wxsBSD:
> `#s1 in (0...100) >= 2` sounds good, much better than `2 occurrences of $s1 in (0..100)`, and very coherent with other constructs.

Sounds good. I will work on it tonight and hopefully have a PR up soon.

> I would keep `N occurrences of <string set> in (0..100)` out of the discussion for now. However, something like this can could work:
> 
> `#s*` -> number of aggregated matches for all strings that start with `s`
> `#s* in (0..100)` -> number of aggregated matches in range (0.100) for all strings that start with `s`
> 
> This doesn't cover all kinds of string_sets, but provides some flexibility.

These are good ideas. I'll not implement them right now, but will revisit in the future.

PR #1486, opened 2021-05-01
Fix OpenBSD include order to correct unknown type errors and broken compilation

pwrdwnsys:
After 4.0.5, the proc/openbsd.c has had the list of includes sorted in alphabetical order, which breaks compilation with numerous `unknown type name` errors. This fix reverts back the ordering and allows compilation to complete successfully. Tested OK on OpenBSD 6.9.


pwrdwnsys:
Having looked at the fuzzer output, it appears that the fuzzer crashed owing to a missing file rather than as a result of my proposed change. Any way to force the fuzzer to retry the job?

plusvic (approved):

PR #1508, opened 2021-05-28
Check if iterator is YR_UNDEFINED before accessing

meme:
In certain programs, the iterator will be an undefined value which will cause a SIGSEGV, and result in an `ERROR_COULD_NOT_MAP_FILE`.

I used `SCAN_FLAGS_NO_TRYCATCH` to find this issue.

### DIFF
--- a/libyara/exec.c
+++ b/libyara/exec.c
@@ -187,7 +187,7 @@ static int iter_array_next(YR_ITERATOR* self, YR_VALUE_STACK* stack)
   if (stack->sp + 1 >= stack->capacity)
     return ERROR_EXEC_STACK_OVERFLOW;
 
-  if (self->array_it.index < yr_object_array_length(self->array_it.array))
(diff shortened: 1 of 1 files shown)


plusvic:
Can you provide the rule that you used to trigger this issue? It would be great to add it as a test case. I'm also thinking about whether the result from iterating an undefined array should be undefined or behave as an empty array as you propose.

meme:
Here is the reduced rule:
```yara
import "macho"

private rule Bad
{
    condition:
        uint32(0) == 0xbebafeca and
            (for any section in macho.segments[0].sections: (section.sectname == "__bad") or
                for any section in macho.segments[1].sections: (section.sectname == "__bad"))
}
```

I cannot share the executable.

I do not feel strongly either way about how this should work semantically, but I felt that an empty array was appropriate.

plusvic (approved):

PR #1509, opened 2021-06-02
Added `pe.version_info_list` with the list of all version info records

xbabka01:
`pe.version_info` is stored in YARA just as a dictionary (hash table). However it keeps track only of the very last key, value pair it sees so if there are more keys A with values B and C, only C is going to be available in YARA.

Because of mentioned reason these were added:
* `pe.number_of_version_infos` : Number of entries in `pe.version_info_list`
* `pe.version_info_list` : Array of pairs (key, value) containing version infos

Sample: `00bb7a440abacc9d9350f1641fe4a9705246642838b927614cfc5cbe20c89751`

Mentioned sample contain multiple `LanguageId` version info records:
* `LanguageId`: `userdefault`
* `LanguageId`: `ar`
* `LanguageId`: `bg`
* ...

For a complete list of version information records, see this [gist](https://gist.github.com/xbabka01/80746b289c3a36b3de3f2fe6a3b304dc).


metthal:
We'd really welcome this PR to reach upstream since we have some troubles scanning such samples with the current approach to `pe.version_info`. So +1.

plusvic (approved):

plusvic:
@xbabka01 Would you mind making the corresponding change to the documentation?

xbabka01:
> @xbabka01 Would you mind making the corresponding change to the documentation?

Documentation for `pe.version_info_list` added

PR #1523, opened 2021-07-01
Add `imports_details` and `delayed_imports_details`

xbabka01:
Same as we have `pe.export_details` in which we can iterate over the list of exports, we should also have `pe.import_details` so that we don't have to access every single import through `pe.imports` function.

Added structure:
* `pe.imports_details`
   * `library_name`
   * `number_of_functions`
   * `functions`
     * `name`
     *  `ordinal`

Also Yara did not parse delay imports. So also added added integer values:
* `pe.number_of_delay_imports` :  Number of delay imported DLLs in the PE. (Number of IMAGE_DELAYLOAD_DESCRIPTOR parsed from file)
* `pe.number_of_delay_imported_functions`:  Number of delay imported functions in the PE.

and added structure:
* `pe.delay_imports_details`
   * `library_name`
   * `number_of_functions`
   * `functions`
     * `name`
     *  `ordinal`

Added overloads for `pe.imports`


docs/modules/pe.rst
```diff
@@ -1017,6 +1017,32 @@ Reference
 
     *Example:  pe.imports(/kernel32\.dll/i, /(Read|Write)ProcessMemory/) == 2*
 
+.. c:type:: import_details
+
+    Array of structures containing information about the PE's imports libraries.
+
+    .. c:member:: library_name
```
Christamac1, line 1024:
```suggestion
    .. c:member:: library_name
```
:Cዘr፤$ተአ::

docs/modules/pe.rst
```diff
@@ -1017,6 +1017,32 @@ Reference
...
+
+        Library name.
+
+    .. c:member:: number_of_functions
```
Christamac1, line 1028: C:Christa

Christamac1 (approved):

xbabka01:
pe_imports source code https://gist.github.com/xbabka01/cdccf5aaf80a88c11426b7e92856c910

docs/modules/pe.rst
```diff
@@ -1017,6 +1025,120 @@ Reference
 
     *Example:  pe.imports(/kernel32\.dll/i, /(Read|Write)ProcessMemory/) == 2*
 
+.. c:function:: delayed_imports(dll_name, function_name)
```
plusvic, line 1028: What if instead of adding `delayed_imports` and `any_imports` we simply implement another version of `imports` that receives an extra argument indicating if you are interested in delayed imports only, standard imports only, or both. The extra argument could be an integer where one bit, means "include delayed imports" and another bit means "include standard imports", by xoring both values you can indicate that you want both.

The current `imports` function can be implemented by calling this new function.

xbabka01: I will implement it your way. At first, I implemented it this way, because it  seemed "cleaner" for me. But now after I read your comment, I agree that it will be better to use binary flags parameter instead of functions. This way there will be one function instead of three.

Examples:

`pe.imports(lib, fun, pe.STANDARD_IMPORT )`
`pe.imports(lib, fun, pe.DELAY_IMPORT)`
`pe.imports(lib, fun, pe.STANDARD_IMPORT | pe.DELAY_IMPORT)`

plusvic: Exactly, that's the idea.

xbabka01: Refactored imports functions

Same as example except order of parameters. Flag specifying which import to search is first. If flag was last it will create multiple overloads with same parameters (`lib, ordinal` vs `lib, flag`) 

Example:

- `pe.imports(, pe.STANDARD_IMPORT, lib, fun)`

libyara/modules/pe/pe.c
```diff
@@ -3010,6 +3459,24 @@ begin_declarations
...
+  begin_struct_array("import_details");
+    declare_string("library_name");
+    declare_integer("number_of_functions");
+    begin_struct_array("functions");
+      declare_string("name");
+      declare_integer("ordinal");
+    end_struct_array("functions");
+  end_struct_array("import_details");
+
+  begin_struct_array("delay_import_details");
```
plusvic, line 3471: `delayed_import_details` is declared twice.

libyara/modules/pe/pe.c
```diff
@@ -1067,9 +1099,278 @@ static IMPORTED_DLL* pe_parse_imports(PE* pe)
...
+    return NULL;
+
+
+  offset = pe_rva_to_offset(pe, yr_le32toh(directory->VirtualAddress));
+
+  if (offset < 0)
+    return NULL;
+
+  import_descriptor = (PIMAGE_DELAYLOAD_DESCRIPTOR)(pe->data + offset);
+  for (; struct_fits_in_pe(pe, import_descriptor, IMAGE_DELAYLOAD_DESCRIPTOR); import_descriptor++)
```
plusvic, line 1226: It looks like this line is past the 80 columns limit used in all the code base. I encourage using `clang-format` for auto-formatting the code.

xbabka01: Formatted using `clang-format`.

libyara/modules/pe/pe.c
xbabka01: Redundancy removed.

plusvic (approved):

PR #1539, opened 2021-08-14
Load version info records with empty value

tomaskender:
Load version info records that have empty values.

Let's say that a version info key `Comments` has a ValueLength = 0.
Right now such values are skipped and thus this key is not displayed in a dump.
These items do not include a null terminator in their values, that's why they have to be added manually.

This PR aims to display such key-value as
```
version_info_list
  [0]
    key = "Comments"
    value = ""
  ...
version_info
  Comments = ""
  ...

```


tomaskender:
@googlebot I fixed it.

plusvic (approved):

plusvic:
@tomaskender I test case for this edge case would be nice, do you have some small PE file with this characteristic that can be added to the tests?

tomaskender:
> @tomaskender I test case for this edge case would be nice, do you have some small PE file with this characteristic that can be added to the tests?

Addressed in the pr above

Issue #1171, opened 2019-11-27
Build: syntax error near unexpected token PROTOBUF_C


github-pba:
Yara fails to build on Debian 10. Seems to be related to Issue  #1163

```
root@server:~# rm -r yara
root@server:~# git clone https://github.com/VirusTotal/yara.git
Cloning into 'yara'...
remote: Enumerating objects: 56, done.
remote: Counting objects: 100% (56/56), done.
remote: Compressing objects: 100% (33/33), done.
remote: Total 15236 (delta 25), reused 39 (delta 17), pack-reused 15180
Receiving objects: 100% (15236/15236), 18.54 MiB | 1.22 MiB/s, done.
Resolving deltas: 100% (10942/10942), done.

root@server:~# cd yara/
root@server:~/yara# ./bootstrap.sh
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.
libtoolize: copying file 'build-aux/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
configure.ac:20: installing 'build-aux/ar-lib'
configure.ac:20: installing 'build-aux/compile'
configure.ac:39: installing 'build-aux/config.guess'
configure.ac:39: installing 'build-aux/config.sub'
configure.ac:8: installing 'build-aux/install-sh'
configure.ac:8: installing 'build-aux/missing'
Makefile.am: installing 'build-aux/depcomp'
parallel-tests: installing 'build-aux/test-driver'
configure.ac: installing 'build-aux/ylwrap'


root@server:~/yara# ./configure
checking whether make supports nested variables... yes
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... no
checking for mawk... mawk
checking whether make sets $(MAKE)... yes
checking whether make supports the include directive... yes (GNU style)
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
[...]

PR #1540, opened 2021-08-14
Escape `PKG_CHECK_MODULES`

catap:
This small changes allows to fix an issues https://github.com/avast/retdec/issues/990 and probably https://github.com/VirusTotal/yara/issues/1171

The root cause of this issues is different shell which is used. Some simple shell like dsh I belive can't survive on this line without escaping, some modern like bash easly can.

Thus at § 3.5 at https://autotools.io/pkgconfig/pkg_check_modules.html it is cleare suggested to escape by `[...]`.


catap:
@googlebot I signed it!

plusvic (approved):

PR #1541, opened 2021-08-18
Fix edge case with pid check.

wxsBSD:
A common use case (I've run into this many times, as have some friends) is to
run yara against a file named after the hash of the content. If the hash happens
to start with a number yara.c treats it as a PID, and will fail. The workaround
is to prefix the file with "./" (or whatever the equivalent is on Windows) so
that the strol() check fails.

This diff switches the behavior so it treats the argument as a file first, and
only if that can't be opened is it then checked for a PID. Further, to ensure the
PID check is accurate I extended it so that the entire argument must be
converted. If there are any characters left after the conversion then it is not
a valid PID and we will not attempt to scan it as it is (by definition) not a
valid file nor a valid PID.

### DIFF
--- a/cli/yara.c
+++ b/cli/yara.c
@@ -1551,12 +1551,17 @@ int _tmain(int argc, const char_t** argv)
     yr_scanner_set_flags(scanner, flags);
     yr_scanner_set_timeout(scanner, timeout);
 
-    long pid = _tcstol(argv[argc - 1], NULL, 10);
+    // Assume the last argument is a file first. This assures we try to process
+    // files that start with numbers first.
+    result = scan_file(scanner, argv[argc - 1]);
+    if (result == ERROR_COULD_NOT_OPEN_FILE) {
+      // Is it a PID? To be a PID it must be made up entirely of digits.
+      char *endptr = NULL;
+      long pid = _tcstol(argv[argc - 1], &endptr, 10);
 
-    if (pid != 0)
-      result = yr_scanner_scan_proc(scanner, (int) pid);
-    else
-      result = scan_file(scanner, argv[argc - 1]);
+      if (pid != 0 && argv[argc - 1] != NULL && *endptr == '\x00')
+        result = yr_scanner_scan_proc(scanner, (int) pid);
+    }
 
     if (result != ERROR_SUCCESS)
     {


wxsBSD:
Here's an example of this in action. There are 3 files I'm using to test, named `1`, `1a` and `bar`. The goal is that any file that starts with a number is treated as a file first, and only when it can't be opened is it then checked for a PID (with a stricter check than is currently done). If it is possibly a PID it is then passed to `yr_scanner_scan_proc()`.

```
wxs@wxs-mbp tmp % cat 1
foo
wxs@wxs-mbp tmp % cat 1a
foobar
wxs@wxs-mbp tmp % cat bar
bar
wxs@wxs-mbp tmp % cat ~/src/yara/rules/test.yara
rule a {
  strings:
    $a = "foo"
    $b = "bar"
  condition:
    any of them
}
wxs@wxs-mbp tmp % ~/src/yara/yara -s ~/src/yara/rules/test.yara 1
a 1
0x0:$a: foo
wxs@wxs-mbp tmp % ~/src/yara/yara -s ~/src/yara/rules/test.yara 1a
a 1a
0x0:$a: foo
0x3:$b: bar
wxs@wxs-mbp tmp % ~/src/yara/yara -s ~/src/yara/rules/test.yara bar
a bar
0x0:$b: bar
wxs@wxs-mbp tmp % ~/src/yara/yara -s ~/src/yara/rules/test.yara 2
error scanning 2: can not attach to process (try running as root)
wxs@wxs-mbp tmp % ~/src/yara/yara -s ~/src/yara/rules/test.yara 2a
error scanning 2a: could not open file
wxs@wxs-mbp tmp %
```

This shows that arguments which start with numbers are treated as files first and PIDs second (with a stricter check, see `2a` failure is treated as a file, not a PID). This should make it a little nicer to scan files named after hashes (and that happen to start with a number).

wxsBSD:
Also, using the `--scan-list` option is not affected by this. Everything is treated as a file, as per usual:

```
wxs@wxs-mbp yara % cat ~/x
/Users/wxs/tmp/1a
/Users/wxs/tmp/1
/Users/wxs/tmp/bar
/Users/wxs/tmp/2
/Users/wxs/tmp/2a
4
wxs@wxs-mbp yara % ./yara rules/test.yara -s --scan-list ~/x
error scanning /Users/wxs/tmp/2: could not open file
error scanning 4: could not open file
error scanning /Users/wxs/tmp/2a: could not open file
a /Users/wxs/tmp/1a
0x0:$a: foo
0x3:$b: bar
a /Users/wxs/tmp/bar
0x0:$b: bar
a /Users/wxs/tmp/1
0x0:$a: foo
wxs@wxs-mbp yara %
```

The three files which do not exist (`/Users/wxs/tmp/2`, `/Users/wxs/tmp/2a` and `4`) are all treated as files and not PIDs.

plusvic (approved):

cli/yara.c
```diff
@@ -1551,12 +1551,17 @@ int _tmain(int argc, const char_t** argv)
...
-      result = scan_file(scanner, argv[argc - 1]);
+    // Assume the last argument is a file first. This assures we try to process
+    // files that start with numbers first.
+    result = scan_file(scanner, argv[argc - 1]);
+    if (result == ERROR_COULD_NOT_OPEN_FILE) {
+      // Is it a PID? To be a PID it must be made up entirely of digits.
+      char *endptr = NULL;
+      long pid = _tcstol(argv[argc - 1], &endptr, 10);
+
+      if (pid != 0 && argv[argc - 1] != NULL && *endptr == '\x00')
```
JohnLaTwC, line 1562: since `pid` is declared to be a `long`, the call to `_tcstol` could return a negative number?  But there are no negative PIDs, so how about:

```
      if (pid > 0 && argv[argc - 1] != NULL && *endptr == '\x00')
```

PR #1544, opened 2021-08-23
Fullword fix

wxsBSD:
Fix https://github.com/VirusTotal/yara-python/issues/184, which turned out to be because isalnum() is locale aware. When running yara-python on MacOS it considers certain byte values to be alphanumeric when I think the intention is to only consider [0-9a-zA-z]. See the details of the issue for a simple script which reproduces it.

Since this only appears under MacOS I didn't have a good way to write a test case but I did test it. The test is in https://gist.github.com/wxsBSD/2e9b270552122cd865cdafe01989fdca - this shows that both "fullword" on a text string and "\w and \W" both work as expected under MacOS now.

### DIFF
--- a/libyara/include/yara/strutils.h
+++ b/libyara/include/yara/strutils.h
@@ -99,3 +99,5 @@ int strcmp_w(const char* w_str, const char* str);
 size_t strlcpy_w(char* dst, const char* w_src, size_t n);
 
 #endif
+
+int yr_isalnum(const uint8_t* s);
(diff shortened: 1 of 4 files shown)


wxsBSD:
I had some people test this and it appears NOT to trigger under Ubuntu and possibly FreeBSD (though the person testing may have set a non-default locale so I don't know how accurate the test was). In any case, I can definitely trigger this on MacOS, which is good enough for me. ;)

wxsBSD:
I'm a little concerned this might break some existing rules but I think it's better to be consistent across OS' and break some rules in the process.

plusvic (approved):

Issue #1454, opened 2021-02-24
match multiple conditions in offset range


csyzhang:
Hi, I am learning Yara and have a question about how to create a rule that matches multiple conditions in an offset range
For example, I check two strings str1 and str2. Both str1 and str2 can appear multiple times in the file. My hit condition is that I see both strings and there exists a matching pair of str1&str2 whose offset is less than 100 bytes. The order doesn't matter. Does Yara support this? if it supports, how the rule looks like? Thanks!

plusvic:
This does what you want:

```yara
import "math"

rule exaple {
  strings:
   $str1 = "aaaaa"
   $str2 = "bbbbb"

  condition:
   for any i in (1..#str1) : (
     for any j in (1..#str2) : (
       math.max(@str1[i],@str2[j]) - math.min(@str1[i],@str2[j]) < 100
     )
   )
}
```

The condition would be clearers with:

```
math.abs(@str1[i] - @str2[j]) < 100
```
instead of:

```
math.max(@str1[i],@str2[j]) - math.min(@str1[i],@str2[j]) < 100
```

But the "math" module doesn't have a "abs" function for getting the absolute value.

PR #1545, opened 2021-08-23
Add a math.abs()

wxsBSD:
As mentioned in #1454, this makes checking the distance between two matches
regardless of the order of matches easier.


wxsBSD:
Saw this mentioned in the issue and realized it is a good idea, so I implemented it. We have needed something like this in the past, not sure why I didn't think about implementing it!

plusvic (approved):

PR #1552, opened 2021-08-25
a heuristic for atoms with repeating bytes

regeciovad:
At Avast, we noticed that Yara is sometimes choosing atoms with repeating bytes such as `90 90 90 90`. We found, that it can lead to a noticeable scanning slow down, such as with a file `66b8f108fd41d9f03c8b286cf0851de0403768c2bde9f39516cf97f4ae54a560`. 

I proposed to penalize all literal atoms that contain only one unique byte.

```
rule1.yar - around 7 KB, 5 rules
$ time ./yara rule1.yar 66b8f108fd41d9f03c8b286cf0851de0403768c2bde9f39516cf97f4ae54a560
real    0m1.239s
user    0m1.231s
sys     0m0.013s
$ time ./updated_yara rule1.yar 66b8f108fd41d9f03c8b286cf0851de0403768c2bde9f39516cf97f4ae54a560
real    0m0.203s
user    0m0.194s
sys     0m0.013s

rule2.yar - around 356 KB, over 900 rules
$ time ./yara rule2.yar 66b8f108fd41d9f03c8b286cf0851de0403768c2bde9f39516cf97f4ae54a560
real    0m18.497s
user    0m18.492s
sys     0m0.009s
$ time ./updated_yara rule2.yar 66b8f108fd41d9f03c8b286cf0851de0403768c2bde9f39516cf97f4ae54a560
real    0m0.631s
user    0m0.627s
sys     0m0.010s
```


libyara/atoms.c
```diff
@@ -154,15 +155,13 @@ int yr_atoms_heuristic_quality(YR_ATOMS_CONFIG* config, YR_ATOM* atom)
...
     }
   }
 
   // If all the bytes in the atom are equal and very common, let's penalize
   // it heavily.
-  if (unique_bytes == 1 && (yr_bitmask_is_set(seen_bytes, 0x00) ||
-                            yr_bitmask_is_set(seen_bytes, 0x20) ||
-                            yr_bitmask_is_set(seen_bytes, 0xCC) ||
-                            yr_bitmask_is_set(seen_bytes, 0xFF)))
+  if (unique_bytes == 1 && literal_bytes > 1)
```
plusvic, line 164: With this change it looks like an atom like 00 FF FF is considered better than 14 14 14 14, but I would say it's quite the contrary. I agree on taking into account repetitions of 0x90 in the heuristics (I've faced this issue myself), but treating all repetitions of any byte equally doesn't seem like the right approach. Not all repetitions are equally bad. 

What about penalizing the repetition of 0x90 as it's already done for 0x20, 0xCC and 0xFF?

PR #1559, opened 2021-08-31
none keyword

wxsBSD:
Add the "none" keyword to the grammar which is just 0 (similar to "all" and "any").


plusvic (approved):

lupiotte:
Hello,

From my understanding the "none" statement doesn't work as I expect ;). Indeed even if a string appear in a scanned file the expression using "none" will report a true condition.
I have leveraged different version yara-v4.2.1-1934-win64 / yara-master-1798-win64 / yara-4.2.0 with the same issue.
Test done via official precompiled binaries (only x64 version tested) and executed under Windows environement (W10).

BTW you haven't included in your testing rules tests (tests/test-rules.c) which should have highlighted this issue.
image

Something like:

assert_**false**_rule(
      "rule test { strings: $a = \"dummy1\" $b = \"dummy2\" $c = \"ssi\" "
      "condition: none of them }",
      TEXT_1024_BYTES "mississippi");

assert_**false**_rule(
      "rule test { strings: $a1 = \"dummy1\" $b1 = \"dummy2\" $b2 = \"ssi\""
      "condition: none of ($a*, $b*) }",
      TEXT_1024_BYTES "mississippi");
 
Hope it's help.

PR #1560, opened 2021-09-01
Add parsing of '/N' section names

xbabka01:
Even though it's not a standard, MinGW and Cygwin compilers use this feature to store section names which are longer than 8 characters. Basically whenever a section name is /N where N is positive integer, then this is a index into COFF string table where the section name is stored.

Yara will expose it via full_name. If the name in the section table contains a slash (/) followed by a representation of the decimal number in ASCII format, then this field  contains a string from the specified offset in the string table. Otherwise, this field contains the same value as a name field.


docs/modules/pe.rst
```diff
@@ -560,6 +560,13 @@ Reference
 
         Section name.
 
+    .. c:member:: full_name
+
+        If the name in the section table contains a slash (/) followed by
```
plusvic, line 565: Explaining that this is a technique used by some compilers, as you already did in the commit message would be useful in the documentation too. Without this context is hard to figure out why this feature is useful.

libyara/modules/pe/pe.c
```diff
@@ -1895,6 +1895,71 @@ static void pe_parse_certificates(PE* pe)
...
+  }
+
+  // Calculate offset of string table (String table is immediately after coff
+  // table)
+  uint64_t string_offset = coff_offset + coff_number * sizeof(IMAGE_SYMBOL);
+  uint64_t string_index = 0;
+
+  // Calculate string index/offset in string table
+  int i = 1;
+  for (; i < IMAGE_SIZEOF_SHORT_NAME && isdigit(section_name[i]); i++)
```
plusvic, line 1938: It looks like `i` can be declared inside the `for` statement, it's used only inside the loop.

tests/test-pe.c
```diff
@@ -587,6 +587,16 @@ int main(int argc, char** argv)
       "tests/data/ca21e1c32065352d352be6cde97f89c141d7737ea92434831f998080783d5386");
 
 
+  assert_true_rule_file(
```
plusvic, line 590: A test case with a file that actually has a section like this is important for making sure that this feature is working as intended.

xbabka01: Added `pe_mingw` (program that print `Hello`) which was compiled by mingw into `tests/data`. This is then used in added test to test `full_name`

libyara/modules/pe/pe.c
xbabka01: Moved

docs/modules/pe.rst
xbabka01: Changed doc

PR #1565, opened 2021-09-16
Implement OP_COUNT_IN.

wxsBSD:
Add a new instruction (OP_COUNT_IN) which is used to check the number of
occurrences of a string in a range. You can use this with the following syntax:

 #a in (0..100) == 2

This will check that the $a string matches exactly 2 times in the range
provided.

### DIFF
--- a/docs/writingrules.rst
+++ b/docs/writingrules.rst
@@ -899,6 +899,16 @@ For example:
 This rule matches any file or process containing the string $a exactly six times,
 and more than ten occurrences of string $b.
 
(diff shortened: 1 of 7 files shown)


wxsBSD:
This is as discussed in https://github.com/VirusTotal/yara/pull/1483#issuecomment-920111367.

wxsBSD:
If you're happy with it, I will add docs and we can get it merged.

plusvic:
Looks good to me.

plusvic (approved):

PR #1568, opened 2021-09-20
Add function pe.is_dotnet() for .NET file detection

HoundThe:
This PR adds `is_dotnet()` to the `pe` module to check if the file is indeed a .NET file. I've created a single internal function for this in the `dotnet` module so the detection in `pe` and `dotnet` are identical (To dodge the situation when `pe` says the file is not .NET, but `dotnet` says otherwise and vice versa). The rules for the detection are based on advice from @ladislav-zezula and are as follows:

For 32bit:

- DataDirectory[14].VirtualAddress != 0
- EntryPoint bytes 0xFF 0x25
- DataDirectory[14].VirtualAddress address contains 32bit value 0x48 (Size of CLI header)
- DataDirectory[14].VirtualAddress + 8 contains RVA on which exists value 0x424a5342 (.NET Metadata magic as specified in ECMA-335)

For 64bit:

- DataDirectory[14].VirtualAddress != 0
- NumberOfRvaAndSizes >= 14
- DataDirectory[14].VirtualAddress address contains 32bit value 0x48 (Size of CLI header)
- DataDirectory[14].VirtualAddress + 8 contains RVA on which exists value 0x424a5342 (.NET Metadata magic as specified in ECMA-335)

I've had to remove one of the .NET tests because the file had an invalid entry-point, but I've reused it in test-pe.c


HoundThe:
@googlebot I signed it!

HoundThe:
In the 'dotnet' module this check is pretty much default, before parsing the .NET file. I thought this kind of function could make more sense in a PE module, as you're checking 'if this PE file is indeed .NET' without actually parsing/analyzing additional .NET features at the time.

plusvic:
Sure, I agree that this can be useful for anyone that doesn't want to use the "dotnet" module, and only wants a quick way for checking is a file is a .NET binary or not. But giving that we already have a dedicated "dotnet" module why not taking advantage of it for anything that is .NET-related? 

My guess is that anyone using the `is_dotnet()` function sooner or later is going to be interested in other .NET-specific features.

I think the best argument towards adding this to the "pe" module instead of "dotnet" is that "dotnet" is not built into YARA by default. But I think that "dotnet" is mature enough to be included by default.

HoundThe:
Okay, but in the case of moving the `is_dotnet()` into the `dotnet` module, then maybe it would be better as a module member and not a function? Because the same code would be executed twice, once during `dotnet` module processing and one extra time during the function call. What do you think?

plusvic:
Sure, it makes more sense as a module attribute instead of a function.

plusvic:
I've reviewed the`pe` module and realized that a few inconsistencies have been introduced over time, `is_pe` is an attribute while `is_dll`, `is_64bit` and `is_32bit` are functions. I think we should homogenize this, and probably a function makes more sense because it's more flexible. The function could simply return a precomputed value, though.

It's too late for removing the `is_pe` attribute, but a new function `is_pe` could be added.

HoundThe:
So you want to change, `is_dotnet` to a function as well? And you want to make these `pe` module changes within this pull request?

plusvic:
Yes, revert `is_dotnet` back to a function. Adding a `is_pe` function is better done in a separate PR.

HoundThe:
I've reverted it to the function and I've cached the result in a similar way that the PE module caches.

plusvic (approved):

PR #1591, opened 2021-11-13
few small bugfixes

camptatopenmars:
fix UINT16_MIN macro fence
avoid redefinition of 'result' in a macro
fix assert since apparently 'i' was intended to be re-used in exec.c

### DIFF
--- a/cli/common.c
+++ b/cli/common.c
@@ -71,6 +71,7 @@ char* unicode_to_ansi(const char_t* str)
 }
 #endif
 
+#ifndef MINLIBYARA
 bool compile_files(YR_COMPILER* compiler, int argc, const char_t** argv)
 {
   for (int i = 0; i < argc - 1; i++)
@@ -204,6 +205,7 @@ int define_external_variables(
 
   return result;
 }
+#endif
 
 bool is_integer(const char* str)
 {
--- a/cli/common.h
+++ b/cli/common.h
@@ -40,6 +40,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 char* unicode_to_ansi(const char_t* str);
 #endif
 
+#ifndef MINLIBYARA
 bool compile_files(
 	YR_COMPILER* compiler,
 	int argc,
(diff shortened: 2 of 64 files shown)


camptatopenmars:
did the CLA thing

plusvic:
What's the purpose of the `MINLIBYARA` stuff? The `cli/minyara.c` and `cli/minruntime.c` contain a lot of code that is duplicated elsewhere and makes maintenance harder.

camptatopenmars:
> What's the purpose of the `MINLIBYARA` stuff? The `cli/minyara.c` and `cli/minruntime.c` contain a lot of code that is duplicated elsewhere and makes maintenance harder.

My apologies, forgot git would continue to stack up changes if I used the same branch for my continued doodling.  I only wanted to call your attention to the _first_ commit fb3ef7f since I think these are minor, but real bugs in the source tree.

w.r.t. the MINLIBYARA changes, those are to enable me to build a libyara that is minimal - it drops out floating point and only evaluates rules compiled by a full libyara.  Point being that it makes it a lot easier to embed libyara into a runtime environment that is constrained and lacks a full crt.  The actual issue with this is that the generated flex files have direct file i/o, malloc/realloc etc. calls in them.  So I end up still having to do some ugly things there, but it got me most of the way.

plusvic:
I've handpicked some of your changes in e2f3e544ce2dc07cd81ddc267e3af8c32897a38f and 61d0b5812e522b130971178b24c1c7f98956f098. I'm not sure why it's necessary to rename the `result` variable used in macros, though.

camptatopenmars:
Result was used in a outer scope in a few places so the macro ended up
reusing the name.

T.

plusvic:
> Result was used in a outer scope in a few places so the macro ended up reusing the name. T.

Yes, but notice that `result` is declared within its own scope, it hides the outer `result` but, don't mess with it

camptatopenmars:
No real interest from me if you take that part or not, just explaining
why.  I'm just cleaning up compiler warnings - even innocent ones since the
various signed/unsigned casts require a bit more study to see how innocent
they really are.

t.

plusvic:
Thank you for the additional context. I've renamed the "result" variable in 7251a51a084dc7571db6f602244d182ea2ceffcc. I'm closing the issue as the main three issues has been addressed.

Issue #1585, opened 2021-10-29
Default module idea "console"


tlansec:
This is an idea for a new default YARA module whose sole functionality is to print the result of the function within it. This would be used during signature development to debug matched data which is otherwise invisible to the user. For example:

```
import "console"
import "pe"

rule example_rule
{
condition:
    console.log(pe.imphash())
}
```

This would print out the pe.imphash() value which cannot be seen (by YARA) in any other way.

Obviously there are lots of ways to see an imphash for a file (although none via YARA) - but for calculated objects the console output could be helpful: e.g.

`console.log(uint16(some_calculated_offset)+32)`

Issue #1221, opened 2020-02-25
Dump data from yara functions


dzzie:
I am hoisting this out of a comments on another thread.

It is pretty handy to have a debug function which can dump calculated values from yara so you can quickly see what a script is thinking or to output the results of a quick query for testing. the -D dump module info is helpful, but not all things like imphash() are included.

I ended up adding a dbg function to the pe module because it was the easiest route. I think something like this would be a worth including in the main branch.

Examples I have used it for:

```
//invalid import table = d41d8cd98f00b204e9800998ecf8427e which is md5("") 
pe.dbg(pe.imphash()) 

//dbg always returns 1, so 'for any' we invert to loop through all
for any i in (0 .. 32):(not pe.dbg(i,pe.sections[i].name))

//see how resource names are held any path?
for any i in (0 .. pe.number_of_resources-1):(
         not pe.dbgw(pe.resources[i].name_string)
 ) 

//quick and dirty calc/dump
for all i in (0 .. pe.number_of_sections -1):(
            pe.dbg( 
                pe.sections[i].name,
                math.entropy(
                    pe.sections[i].raw_data_offset, 
                    pe.sections[i].raw_data_size
                )
            )
        )
```
The current prototypes I am supporting include:
```
  declare_function("dbgw", "s", "i", dbgw); //wide string version for resource names
  declare_function("dbg", "s", "i", dbg);   //string  
  declare_function("dbg", "i", "i", dbgi);      //int 
  declare_function("dbg", "f", "i", dbgf);    //float
  declare_function("dbg", "sf", "i", dbgsf);   //string , float
  declare_function("dbg", "si", "i", dbgsi);    // string , int
  declare_function("dbg", "is", "i", dbgis);   // int , string
  declare_function("dbg", "ii", "i", dbgii);     // int , int 
  declare_function("dbg", "ss", "i", dbgss);  //string , string
```
a few sample implementations:
http://sandsprite.com/blogs/index.php?uid=7&pid=478#dbg

PR #1594, opened 2021-11-23
Add console module.

wxsBSD:
Add a console module which can be used to print things during condition
execution. This is especially useful for debugging complex conditions where
you're following multiple offsets. For example, people sometimes want to express
a PE by doing this:

uint32(uint32(0x3c)) == 0x50450000

What they really want is:

uint32be(uint32(0x3c)) == 0x50450000

Or they want:

uint32(uint32(0x3c)) == 0x00004550

In either case it would be helpful if they could do:

console.hex(uint32(uint32(0x3c)))

Another example of where this could be useful is being able to know what YARA
thinks an imphash is just by doing:

console.log(pe.imphash())


wxsBSD:
This is for #1585.

I decided to make it a compile-time flag for now but am willing to change it if necessary. Also, I'm open to suggestions on how to make this the most usable for people if it is desired.

wxsBSD:
Forgot to mention, this probably needs to be updated to work with Win32. I'm happy to try to do that...

xorhex:
Can the functions be overloaded so a label/text value can be specified as well (to help with which write to console is which - something like  `console.hex("Address to jump table",  @value[i])`

plusvic:
I this ready to merge? Or do you prefer to implement the callback first?

wxsBSD:
I've got a couple of changes coming hopefully today, including the callback.

wxsBSD:
I've updated the PR so that:

- The module is on by default. This will make it more useful to people out of the box. I'm happy to reverse this if desired.
- There is now a `CALLBACK_MSG_CONSOLE_LOG` which is used by this module to call the main callback. Each function generates a single string which is passed to the callback, which means the callback logic is very simple but there is a bit of work that needs to be done in the module to generate the single string.

cli/yara.c
```diff
@@ -1197,6 +1197,10 @@ static int callback(
       return CALLBACK_ERROR;
 
     return CALLBACK_CONTINUE;
+
+  case CALLBACK_MSG_CONSOLE_LOG:
+    printf("%s", (char*) message_data);
```
hillu, line 1202: The library user code (i.e. yara(1)) should be responsible for inserting newlines or whatever separator it wants to add.

It might be a good idea to add a prefix here.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
+      *p++ = s->c_string[i];
     else
-      printf("\\x%02x", (unsigned char) s->c_string[i]);
+    {
+      sprintf(p, "\\x%02x", (unsigned char) s->c_string[i]);
+      p += 4;
+    }
   }
-  printf("\n");
+  *p = '\n';
```
hillu, line 64: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
```
hillu, line 104: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
 {
-  SIZED_STRING* s = sized_string_argument(1);
+  char* msg = NULL;
+  char* s = string_argument(1);
   int64_t i = integer_argument(2);
-  printf("%s%lli\n", s->c_string, i);
+  YR_SCAN_CONTEXT* ctx = __context;
+  YR_CALLBACK_FUNC callback = __context->callback;
+
+  asprintf(&msg, "%s%lli\n", s, i);
```
hillu, line 141: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
 
 define_function(log_float)
 {
+  char* msg = NULL;
   double f = float_argument(1);
-  printf("%f\n", f);
+  YR_SCAN_CONTEXT* ctx = __context;
+  YR_CALLBACK_FUNC callback = __context->callback;
+
+  asprintf(&msg, "%f\n", f);
```
hillu, line 160: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
 {
-  SIZED_STRING* s = sized_string_argument(1);
+  char* msg = NULL;
+  char* s = string_argument(1);
   double f = float_argument(2);
-  printf("%s%f\n", s->c_string, f);
+  YR_SCAN_CONTEXT* ctx = __context;
+  YR_CALLBACK_FUNC callback = __context->callback;
+
+  asprintf(&msg, "%s%f\n", s, f);
```
hillu, line 180: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
 
 define_function(hex_integer)
 {
+  char* msg = NULL;
   int64_t i = integer_argument(1);
-  printf("0x%llx\n", i);
+  YR_SCAN_CONTEXT* ctx = __context;
+  YR_CALLBACK_FUNC callback = __context->callback;
+
+  asprintf(&msg, "0x%llx\n", i);
```
hillu, line 199: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
 {
-  SIZED_STRING* s = sized_string_argument(1);
+  char* msg = NULL;
+  char* s = string_argument(1);
   int64_t i = integer_argument(2);
-  printf("%s0x%llx\n", s->c_string, i);
+  YR_SCAN_CONTEXT* ctx = __context;
+  YR_CALLBACK_FUNC callback = __context->callback;
+
+  asprintf(&msg, "%s0x%llx\n", s, i);
```
hillu, line 219: Get rid of this newline.

libyara/modules/console/console.c
```diff
@@ -36,80 +39,192 @@ define_function(log_string)
...
 
 define_function(log_integer)
 {
+  char* msg = NULL;
   int64_t i = integer_argument(1);
-  printf("%lli\n", i);
+  YR_SCAN_CONTEXT* ctx = __context;
+  YR_CALLBACK_FUNC callback = __context->callback;
+
+  asprintf(&msg, "%lli\n", i);
```
hillu, line 121: Get rid of this newline.

hillu (changes requested):

cli/yara.c
wxsBSD: Good point! Thanks. :)

libyara/modules/console/console.c
```diff
@@ -0,0 +1,265 @@
+/*
+Copyright (c) 2021. Wesley Shields <wxs@atarininja.org>. All Rights Reserved.
```
plusvic, line 2: Let's use the unified copyright header `Copyright (c) 2021. The YARA Authors. All Rights Reserved.` You are already listed in the `AUTHORS` file.

libyara/modules/console/console.c
```diff
@@ -0,0 +1,265 @@
...
+  // Add an extra byte for the NULL terminator.
+  char* msg = (char*) yr_calloc((s->length * 4) + 1, sizeof(char));
+  if (msg == NULL)
+    return_integer(YR_UNDEFINED);
+
+  char* p = msg;
+  for (size_t i = 0; i < s->length; i++)
+  {
+    if (isprint((unsigned char) s->c_string[i]))
+      *p++ = s->c_string[i];
```
plusvic, line 57: If one of the branches uses `{}` let's do it in both.

libyara/modules/console/console.c
```diff
@@ -0,0 +1,265 @@
...
+  if (msg == NULL)
+    return_integer(YR_UNDEFINED);
+
+  char* p = msg;
+  strlcpy(msg, m, msg_len);
+  p += strlen(m);
+  for (size_t i = 0; i < s->length; i++)
+  {
+    if (isprint((unsigned char) s->c_string[i]))
+      *p++ = s->c_string[i];
```
plusvic, line 96: If one of the branches uses {} let's do it in both.

libyara/modules/console/console.c
```diff
@@ -0,0 +1,265 @@
...
+
+  yr_free(msg);
+  return_integer(1);
+}
+
+define_function(log_integer)
+{
+  char* msg = NULL;
+  int64_t i = integer_argument(1);
+  YR_SCAN_CONTEXT* ctx = __context;
```
plusvic, line 116: Use `scan_context()` instead of `__context`.

libyara/modules/console/console.c
```diff
@@ -0,0 +1,265 @@
...
+  yr_free(msg);
+  return_integer(1);
+}
+
+define_function(log_integer_msg)
+{
+  char* msg = NULL;
+  char* s = string_argument(1);
+  int64_t i = integer_argument(2);
+  YR_SCAN_CONTEXT* ctx = __context;
```
plusvic, line 136: Use scan_context() instead of __context.

libyara/modules/console/console.c
```diff
@@ -0,0 +1,265 @@
...
+
+  yr_free(msg);
+  return_integer(1);
+}
+
+define_function(log_float)
+{
+  char* msg = NULL;
+  double f = float_argument(1);
+  YR_SCAN_CONTEXT* ctx = __context;
```
plusvic, line 155: Use scan_context() instead of __context.

plusvic (approved):

Neo23x0:
Will this be compiled into YARA by default or do I have to set ` --enable-console`? 

If it is compiled into YARA by default, how can I disable it?

hillu:
It's implemented as a callback, so if your application that uses libyara should choose to ignore the new callback code, there's no risk exposure.

EWickens:
Hey all, 

Loving the console module - thank you! 

Just wanted to ask about potential support for a format string such as %s within console.log. A potential rule would be as follows, where %s could print out a null terminated string at the given offset (and maybe an option to print out a string of a defined length too?):

`import "console"
rule console_test{
	strings:
    	$https = "https://" wide
	condition:
		$https and console.log("HTTPS URL: %s", @https)
}`

wxsBSD:
The contents at `@https` may or may not be null terminated, it could just be an arbitrary sequence of bytes if someone does `$x = { aa bb cc dd ee ff }. Assuming it is going to be null terminated is not possible, so the question then becomes do we print only the matched bytes since YARA knows their contents and length? At that point we are duplicating the `-s` flag in YARA.

datorr2:
> Will this be compiled into YARA by default or do I have to set ` --enable-console`?
> 
> If it is compiled into YARA by default, how can I disable it? It's a risk exposure that I'd like to avoid.

> It's implemented as a callback, so if your application that uses libyara should choose to ignore the new callback code, there's no risk exposure.

So just wanted to double-check, in the Windows binaries on the release page for v4.2.0-rc1, is it compiled with console support?

I tried to use the feature as detailed in the docs and also from this page:
https://isc.sans.edu/forums/diary/YARAs+Console+Module/28288/

... And it doesn't appear to work.

image

hillu:
@datorr2 I just built the version tagged v4.2.0-rc1 on Linux and I do get output from your sample rule.

wxsBSD:
The fact that you don't get an error during the import statement indicates the console module is there. My guess is the buffer is not flushed on Windows. I don't have a Windows machine to test on though so I can't be sure. But this certainly does look like a Windows specific bug: CC @plusvic

wxsBSD:
@datorr2 Any chance you can try the 32bit version? Upon closer inspection I think it may have only been added to the 64bit build on Windows.

hillu:
An x86_64-w64-mingw32 build on Windows does not cause this problem either. Seems to happen only on MSVC builds, then.

dzzie:
this resolves 1221 which can now be closed as well: https://github.com/VirusTotal/yara/issues/1221

PR #1600, opened 2021-12-06
Add MD5 of decrypted rich data to PE module

HoundThe:
A simple addition to the PE module. An MD5 hash of the clear (XOR decrypted) rich header data.

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -924,6 +924,10 @@ Reference
 
         Data after being decrypted by XORing it with the key.
 
+    .. c:member:: hash
+
+        MD5 hash of the clear rich header data.
+
     .. c:function:: version(version, [toolid])
 
         .. versionadded:: 3.5.0
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -924,7 +924,7 @@ Reference
 
         Data after being decrypted by XORing it with the key.
 
-    .. c:member:: hash
+    .. c:member:: md5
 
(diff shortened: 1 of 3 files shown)


docs/modules/pe.rst
```diff
@@ -924,6 +924,10 @@ Reference
 
         Data after being decrypted by XORing it with the key.
 
+    .. c:member:: hash
```
plusvic, line 927: I would name this `md5` instead of `hash`, as it conveys more information about the kind of hash without having to go to the documentation.

wxsBSD: Is there a reason we don't want to use "hash.md5()" here?

plusvic: Fair enough, as we already have the offset and size for the rich header using `hash.md5()` should be enough. And as bonus you can use the hashing algorithm you want.

HoundThe: Yes, md5 sounds good. Regarding using a hash.md5(), if I understand the function correctly, it operates above a specified chunk of the file, but the rich header hash is done above the decrypted rich data with the rolling xor key, in the file, it would still be encrypted.

HoundThe: Oh, I see I can also pass the clear data directly to the hash module. 
`hash.md5(pe.rich_signature.clear_data) == "..."`
Cool, I guess hash.md5() is enough then.

plusvic:
Closing this issue, as `hash.md5(pe.rich_signature.clear_data) == ...` provides the desired functionality.

PR #1605, opened 2021-12-06
Add parsing of .NET user types from .NET metadata stream

HoundThe:
This PR introduces parsing of the user-defined types from the .NET metadata tilde stream. It extracts class names, base types, modifiers, methods with their return types. arguments, etc. (more in the docs). I've been testing the modification with IL/DnSpy and RetDec.

For better code readability and maintainability, as it includes many string concatenations, I've created a simple string-like type in the libyara, that allows me to work with variable strings.

I've also noticed that Bazel tests do not include `test-dotnet.c`. I wasn't sure why, so I've included it.

I've also included some binary with .NET user type metadata to have something to write tests with, I am not sure what the policy is with including another binary to the testing sample.


libyara/modules/dotnet/dotnet.c
```diff
@@ -16,66 +16,1639 @@ limitations under the License.
...
+  uint32_t biggest;
+  uint32_t x;
+
+  if (count == 0)
+    return 0;
+
+  va_start(ap, count);
+  biggest = va_arg(ap, uint32_t);
+
+  for (i = 1; i < count; i++)
```
plusvic, line 46: Define `i` in the loop, like `for (int i = 1; ...)`. There are many places in the code were we already define these variables outside the `for` statement, this was because certain old Microsoft compilers didn't support it, but we are not using those compilers anymore.

libyara/modules/dotnet/dotnet.c
```diff
@@ -16,66 +16,1639 @@ limitations under the License.
...
+  return false;
+}
+
+// ECMA-335 II.23.2 blob heap uses variable length encoding of integers
+static uint32_t read_blob_unsigned(const uint8_t** data, uint32_t* len)
+{
+  if (*len < 1)
+    return 0;
+
+  /* first byte is enough to decode the length
```
plusvic, line 511: Use `//` for all comments. When you want to comment whole functions or large sections of code it becomes harder if you have `/**/` in the code, as this types of comments can't be nested.

libyara/simple_str.c
```diff
@@ -0,0 +1,103 @@
+#include <stdarg.h>
```
plusvic, line 1: Add comment with copyright information.

HoundThe:
Thanks for the feedback! Sorry for the delay, I was finishing the last few exams past week.

HoundThe:
I've noticed the test failing on misaligned load and changed the reading functions to work byte-by-byte.

plusvic:
I sent a PR to your branch with a proposal for fixing the same issues: https://github.com/HoundThe/yara/pull/1

I prefer using the `yr_unaligned_uXX` functions, which make the intention clearer.

plusvic (approved):

PR #1607, opened 2021-12-07
Add interface tests for math module

LearnToGetBetter:
Add interface tests, including string_entropy, data_entropy, string_deviation, data_deviation, string_mean, data_mean, string_serial_correlation, data_serial_correlation, string_monte_carlo_pi, data_monte_carlo_pi and in_range.

### DIFF
--- a/tests/test-math.c
+++ b/tests/test-math.c
@@ -111,6 +111,118 @@ int main(int argc, char** argv)
       }",
       "CCABACC");
 
+  assert_true_rule_blob(
+      "import \"math\" \
+      rule test { \
+        condition: \
+          math.entropy(\"AAAAA\") == 0.0 \
+      }",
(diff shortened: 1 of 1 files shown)


tests/test-math.c
```diff
@@ -111,6 +111,118 @@ int main(int argc, char** argv)
       }",
       "CCABACC");
 
+  assert_true_rule_blob(
```
plusvic, line 114: In cases where the rule doesn't uses the data for anything, you can use `assert_true_rule` instead of `assert_true_rule_blob`

plusvic (approved):

tests/test-math.c
LearnToGetBetter: Modified as recommended. Please review it again.

PR #1611, opened 2021-12-07
Fix spelling errors

LearnToGetBetter:


plusvic (approved):

libyara/include/yara/pe.h
```diff
@@ -150,7 +150,7 @@ typedef struct _IMAGE_FILE_HEADER
 // Local symbols stripped from file.
 #define IMAGE_FILE_LOCAL_SYMS_STRIPPED     0x0008
 // Aggressively trim working set
-#define IMAGE_FILE_AGGRESIVE_WS_TRIM       0x0010
+#define IMAGE_FILE_AGGRESSIVE_WS_TRIM      0x0010
```
plusvic, line 153: It looks like `IMAGE_FILE_AGGRESIVE_WS_TRIM` is the way in which this is declared in Window's headers, see: https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_file_header

LearnToGetBetter: The modification has been rolled back. Please review it again.

PR #1613, opened 2021-12-08
Solving memory leakage in function get_cache in magic.c (Synchronous modification to 4.0.x)

LearnToGetBetter:
Synchronous modification to 4.0.x


plusvic:
Actually the 4.0.x is already frozen and no other 4.0.x version will be published. It could be merged in the 4.1.x branch, but due lo the low impact of this fix, I would release it in the upcoming 4.2.0 version.

PR #1615, opened 2021-12-14
Fix yr_re_fast_exec on larger than 2GiB files

vlaci:
There was a silent integer overflow on the `max_bytes_matched`
calculation as `input_forwards_size` can be way more than what can fit
in a signed int. `yr_re_exec` doesn't suffer from the same issue as
a match there is bounded to `YR_RE_SCAN_LIMIT` which already fits into
an int just fine.

### DIFF
--- a/libyara/re.c
+++ b/libyara/re.c
@@ -2085,10 +2085,13 @@ int yr_re_fast_exec(
   RE_FAST_EXEC_POSITION* last;
 
   int input_incr = flags & RE_FLAGS_BACKWARDS ? -1 : 1;
-  int max_bytes_matched = flags & RE_FLAGS_BACKWARDS
-                              ? (int) yr_min(input_backwards_size, INT_MAX)
-                              : (int) yr_min(input_forwards_size, INT_MAX);
   int bytes_matched;
+  int max_bytes_matched;
+
+  if (flags & RE_FLAGS_BACKWARDS)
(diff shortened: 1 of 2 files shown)


libyara/re.c
```diff
@@ -2084,8 +2086,8 @@ int yr_re_fast_exec(
 
   int input_incr = flags & RE_FLAGS_BACKWARDS ? -1 : 1;
   int max_bytes_matched = flags & RE_FLAGS_BACKWARDS
-                              ? (int) input_backwards_size
-                              : (int) input_forwards_size;
+                              ? (int) yr_min(input_backwards_size, INT_MAX)
+                              : (int) yr_min(input_forwards_size, INT_MAX);
   int bytes_matched;
```
plusvic, line 2091:
```sugestion
  int bytes_matched;
  int max_bytes_matched;

  if (flags & RE_FLAGS_BACKWARDS)
    max_bytes_matched = (int) yr_min(input_backwards_size, YR_RE_SCAN_LIMIT);
  else
    max_bytes_matched = (int) yr_min(input_forwards_size, YR_RE_SCAN_LIMIT);
```

plusvic (approved):

QKaiser:
@plusvic when can we expect this to land in official releases of Yara ?

plusvic:
Yes, this has been fixed in v4.2.0-rc1, the release candidate before the official 4.2.0.

PR #1622, opened 2021-12-21
Implement our own version of asprintf.

plusvic:
This fixes two issues:

1. asprintf is a GNU extension and it's not available in all compilers, specifically Microsoft Visual Studio.
2. asprintf allocates memory using malloc, but the console module frees the memory using yr_free, this works fine in those platforms in which yr_malloc/yr_free is a direct call to malloc/free, but it won't work in Windows where yr_malloc/yr_free are calls to HeapAlloc/HeapFree with a dedicated heap.


wxsBSD (approved):
These look good to me. Thank you for taking care of this as I don't have a Windows machine. :(

PR #1623, opened 2021-12-22
Process Authenticode signatures using authenticode-parser

HoundThe:
This PR removes the current Authenticode parsing to use the https://github.com/avast/authenticode-parser library. This offers additional information, including a verification of the signatures.

I have kept the existing members and only added new ones, so there isn't an issue with backward compatibility.

I am unsure about all the build systems, as the authenticode-parser depends on OpenSSL, it should not get built if OpenSSL is not found and I am not completely sure if/how to reflect that for example in the VS projects and Bazel.


plusvic:
This is a large contribution, you should add yourself or Avast to the `AUTHORS` and `CONTRIBUTORS` files. The `AUTHORS` file contain the copyright holders, if you are working on this on behalf of Avast, the company should appear in `AUTHORS` while your own name can be added to `CONTRIBUTORS`.

HoundThe:
Thanks, I've added Avast to AUTHORS and myself to CONTRIBUTORS then

plusvic (approved):

plusvic (approved):

PR #1624, opened 2021-12-22
Integrate tlshc library, calculate telfhash and import hash in ELF module

HoundThe:
This PR integrates the tlshc library https://github.com/avast/tlshc that provides a C implementation of the TLSH hashing function https://github.com/trendmicro/tlsh, calculates Telfhash using the TLSH hashing function, and also creates a simple MD5 hash of imported symbols for ELF files.

### DIFF
--- a/libyara/modules/elf/elf.c
+++ b/libyara/modules/elf/elf.c
@@ -177,6 +177,9 @@ cleanup:
   return_string(YR_UNDEFINED);
 }
 
+#if defined(HAVE_LIBCRYPTO) || defined(HAVE_WINCRYPT_H) || \
+    defined(HAVE_COMMONCRYPTO_COMMONCRYPTO_H)
+
 define_function(import_md5)
 {
   YR_OBJECT* obj = module();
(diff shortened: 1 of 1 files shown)


HoundThe:
Some tests were failing due to the usage of `md5_ctx` in `import_md5` not being hidden in case there is no crypto lib to support it.

plusvic (approved):

dmknght:
The current tlsh library of Avast is missing important feature: calculate 2 hashes and give the score.
The original code has this method `int totalDiff(const Tlsh *, bool len_diff=true) const;`. However, there's no function like that in https://github.com/avast/tlshc/blob/main/include/tlshc/tlsh.h. Therefore, current hash comparison in Yara module is based on exact hash matching. It's not good IMO.
To use this method / function with Yara, the tlsh module needs to convert a hash from text to a struct's data. Therefore the method `int fromTlshStr(const char* str);` is required. It's also missing in Avast's module.
I created 2 issues on Avast's repository about 2 missing functions. Hope they update this soon.
https://github.com/avast/tlshc/issues/1
https://github.com/avast/tlshc/issues/2

PR #1625, opened 2021-12-25
Add support for PDB path in Apple EFI files (MTOC signature)

cherepanov74:
EFI modules for the Apple platform have something similar to PDB path, which is a path generated by mtoc utility.


plusvic:
Merged in #1627, with tests fixed.

PR #1630, opened 2022-01-12
Add Apple to list of users

mferrier:
usage noted here https://support.apple.com/en-ca/guide/security/sec469d47bd8/web


plusvic:
XProtect has been using YARA for a long time, but we are adding companies to the "who's using YARA" list when the request comes from someone representing the company. Unfortunately nobody from Apple has reached out asking for this, in fact I'm very surprised that they mentioned YARA explicitly in this article. This is the first time I see that they say publicly that XProtect uses YARA under the hood.

mferrier:
@plusvic okay, thanks for the explanation!

Issue #1631, opened 2022-01-14
Regression bug while parsing rules.


plusvic:
After 7fabd955ab5c7d81bcb04caa22ffdb6625d6a841 the following rule produces a syntax error:

```
rule test
{
   condition:
      for all a in (0..3): (a*0 == 0)
}
```

It looks related to the `a` identifier followed by the `*` operator.

plusvic:
The lexer produces the `_IDENTIFIER_WITH_WILDCARD_` token every time it sees some identifier followed by `*` (without any spaces). The text `a*` produces the  `_IDENTIFIER_WITH_WILDCARD_` token, instead of producing `_IDENTIFIER_` followed by `*`. In fact, just adding a space, like in `a *0` fixes the syntax error.

PR #1632, opened 2022-01-14
Fixes issue #1631.

plusvic:
The lexer doesn't produce an _IDENTIFIER_WITH_WILDCARD_ token anymore, instead it produces two tokens _IDENTIFIER_ and `*`. In the context of a rule_enumeration_item, the grammar handles this sequence of tokens as an identifier with a wildcard instead of a multiplication. The drawback with this solution is that spaces are now valid in between the identifier prefix and the `*`. For example, this is valid:

`all of (some_rule_prefix *) `


wxsBSD (approved):
This seems right to me and sorry for the bug! It's not great that we now accept spaces like you mention but it's only in rule enumerations, and I doubt anyone will even realize it works this way. ;)

Issue #1637, opened 2022-01-19
The mapped file for each segment could not be found during process memory scan


driverxdw:
Hello, when I used the `v4.2.0-rc1` version of yara to do a single process `memory scan` test, I found that the effect was not very good (the size of vmrss increased a lot, almost equal to vmsize), so I used `gdb` to dynamically debug it and found that in the `proc_info->map_path` variable obtained in `libyara/proc/linux.c:366` has `'\n'` at the end, resulting in `libyara/proc/linux.c:164` unable to find the corresponding file, so yara will access all virtual memory of the target process, without going to the optimization logic after `libyara/proc/linux.c:245`, which will eventually lead to an increase in vmrss. Sorry for my poor english, i hope someone can help me correct the grammar in this issue.

PR #1638, opened 2022-01-20
remove the '\n' in the map_path variable

driverxdw:
remove the '\n' in the map_path variable

### DIFF
--- a/libyara/proc/linux.c
+++ b/libyara/proc/linux.c
@@ -324,17 +324,25 @@ YR_API YR_MEMORY_BLOCK* yr_process_get_next_memory_block(
   if (proc_info->next_block_end <= current_begin)
   {
     int n, path_start;
+    char *p;
 
     while (fgets(buffer, sizeof(buffer), proc_info->maps) != NULL)
     {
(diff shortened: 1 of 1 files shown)


driverxdw:
#1637

zyazhb (approved):

libyara/proc/linux.c
```diff
@@ -337,6 +337,13 @@ YR_API YR_MEMORY_BLOCK* yr_process_get_next_memory_block(
         } while (c >= 0 && c != '\n');
       }
 
+      // Remove '\n' at the end of the line
+      len = strlen(buffer);
```
plusvic, line 341: I think we can reuse the result returned by `strrchr` instead of calling `strlen`, which would iterate over the buffer again. If `buffer` contains the `\n` character `strrchr` already returns a pointer to it, if not, you don't need to remove it.

driverxdw: I have re-edited and pushed it, please review it again, thank you very much~ XD

plusvic (approved):

PR #1644, opened 2022-02-06
Fix mistake in console docs.

wxsBSD:


wxsBSD:
This is a copy/paste mistake when I originally wrote the documentation for the console module. The only two supported hex functions are `hex(string, integer)` and `hex(integer)`.

plusvic (approved):

Issue #1646, opened 2022-02-07
Slow scanning


zagge-cgeo:
When using yara with following rule 

```
rule lotsofdots  {
   strings:
     $ = /foobar\.{128,}foobar/
   condition:
     all of them
}
```

to scan the sample e8fa826ce434a9b4a9c52198579f286c available on virustotal  it takes over 30 minutes on my i7 to finish.

The prefix/suffix foobar is not in the sample. There is no warning about slow scanning when using the above rule.

wxsBSD:
I'm a bit out of my league on this but it is matching on all the `.` bytes at the end of the file. From my brief reading of the way atoms are extracted out it is searching for both `foob` and `....` simultaneously. Because there are so many `.` bytes at the end of the file the performance becomes terrible. I would have thought the atom selection process would have thrown out `....` as an option, but it doesn't appear so.

The atoms that end up being selected look like `OR(2E2E2E2E,666F6F62)` (that is from `yr_atoms_tree_node_print()`) and `_yr_atoms_choose()` is selecting the `....` atom when it should arguably be selecting the `foob` one. I'm still digging a bit (mostly because I'm out of my league and want to learn a bit here) to see if I can adjust the selection algorithm to avoid this situation.

wxsBSD:
I put up a PR which attempts to address this issue, but I have no idea if it is the right thing to do or not (or if I even did it correctly). As discussed in the issue the idea is to encourage slightly better atom selection by penalizing the atom when all the bytes are identical.

This does make the performance of your rule with your test file significantly better as it ends up using `r...` as an atom. However, it is still possible to write a terribly performing regex in your example, but that will always be the case if you use `\.{128,}` in your regex. For example, this performs terribly: `$a = /\x2e\x14\x00\x00\.{128,}/` on the example given but I'm not sure we can do much about it?

PR #1647, opened 2022-02-07
Attempt to improve atom selection.

wxsBSD:
If you have an atom that is all the same bytes we should penalize it a little
bit to try to discourage the selection of that atom. We should prefer atoms with
unique values as much as possible.

### DIFF
--- a/libyara/atoms.c
+++ b/libyara/atoms.c
@@ -87,6 +87,7 @@ will end up using the "Look" atom alone, but in /a(bcd|efg)h/ atoms "bcd" and
 #include <yara/stack.h>
 #include <yara/types.h>
 #include <yara/utils.h>
+#include <yara/re.h>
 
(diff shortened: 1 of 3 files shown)


wxsBSD:
The idea is to penalize atoms that are more than 1 byte in length and contain the same bytes. As discussed in #1646 it is possible to write a regex that ends up with an atom of `....` which performs terribly on the given sample (there are a lot of `.` appended towards the end of the file). Using the example rule in the issue we penalizing this particular atom and end up with an atom of `r...` instead, which is significantly faster in performance on that particular file.

I'm not completely happy with this as I don't think I fully understand what is going on and it feels like a very important thing to get correct so I'd be extra careful when considering this.

plusvic:
This has been addresses in #1656, with some additional fixes.

PR #1648, opened 2022-02-11
Fix cuckoosandbox url

jimmyhmiller:
`https://www.cuckoosandbox.org` has a cert error. But `http://cuckoosandbox.org` works just fine.


jimmyhmiller:
Signed the CLA

plusvic (approved):

PR #1670, opened 2022-03-31
Extend dotnet module

tarterp:
# Overview
This large PR is to extend the dotnet parsing capabilities to look into the .NET directory and MetaData tables further. Not all MetaData tables are being handled, just a few that have been found useful. Further table parsing could be added in the future. This will greatly extend yara capabilities combatting .NET malware that has been previously not available or requiring very complex yara rules.

- Parse more of the .NET Directory, of most notice being the flags and entry point
- Parse the following MetaData Tables
  - Module
  - Typeref
  - Method
  - ImplMap

# Examples
Two examples are provided below, there are more in the tests and documentation. These examples take advantage of the new feature `is_dotnet`.

## Mixed Mode
```
import "pe"
import "dotnet"

rule mixed_mode {
    condition:
      pe.is_pe and
      dotnet.is_dotnet and
      dotnet.Flags & dotnet.COMIMAGE_FLAGS_ILONLY == 0
}
```

## MemberRefs with order preference (if only malware was always this easy)

```
import "pe"
import "dotnet"

rule memberref_order {
  condition:
    pe.is_pe and
    dotnet.is_dotnet and
    for any i in (0..dotnet.number_of_memberrefs - 1): (
        dotnet.memberrefs[i].name == "GetCurrentProcess" and
        dotnet.memberrefs[i+1].name == "AntiDebug" and
        dotnet.memberrefs[i+2].name == "SystemEnumeration" and
        dotnet.memberrefs[i+3].name == "SendHostInfo" and
        dotnet.memberrefs[i+4].name == "HandlerLoop"
    )
}
```

### DIFF
--- a/docs/modules/dotnet.rst
+++ b/docs/modules/dotnet.rst
@@ -30,6 +30,31 @@ using attributes and features of the .NET file format. Let's see some examples:
 Reference
 ---------
 
+.. c:type:: major_runtime_version
+
+    The major version contained in the CLI header
+
+.. c:type:: minor_runtime_version
+
+    The major version contained in the CLI header
+
+.. c:type:: flags
+
+    CLI header runtime flags contains the following values
+    
+    .. c:type:: COMIMAGE_FLAGS_ILONLY 
+    .. c:type:: COMIMAGE_FLAGS_32BITREQUIRED    
+    .. c:type:: COMIMAGE_FLAGS_IL_LIBRARY       
+    .. c:type:: COMIMAGE_FLAGS_STRONGNAMESIGNED 
+    .. c:type:: COMIMAGE_FLAGS_NATIVE_ENTRYPOINT
+    .. c:type:: COMIMAGE_FLAGS_TRACKDEBUGDATA
+
+.. c:type:: entry_point
+
+    If CORHEADER_NATIVE_ENTRYPOINT is set, entry_point represents an RVA 
+    to a native entrypoint. If CORHEADER_NATIVE_ENTRYPOINT is not set, 
+    entry_point represents a managed entrypoint.
+
 .. c:type:: version
 
     The version string contained in the metadata root.
@@ -53,12 +78,12 @@ Reference
     stream object has the following attributes:
 
     .. c:member:: name
-
-        Stream name.
+    
+        Stream name
 
     .. c:member:: offset
 
-        Stream offset.
+        Stream offset
 
     .. c:member:: size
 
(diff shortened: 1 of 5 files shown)


docs/modules/dotnet.rst
```diff
@@ -30,6 +30,31 @@ using attributes and features of the .NET file format. Let's see some examples:
 Reference
 ---------
 
+.. c:type:: major_runtime_version
+
+    The major version contained in the CLI header
+
+.. c:type:: minor_runtime_version
+
+    The major version contained in the CLI header
```
wxsBSD, line 39: s/major/minor/

docs/modules/dotnet.rst
```diff
@@ -30,6 +30,31 @@ using attributes and features of the .NET file format. Let's see some examples:
...
+    .. c:type:: COMIMAGE_FLAGS_IL_LIBRARY       
+    .. c:type:: COMIMAGE_FLAGS_STRONGNAMESIGNED 
+    .. c:type:: COMIMAGE_FLAGS_NATIVE_ENTRYPOINT
+    .. c:type:: COMIMAGE_FLAGS_TRACKDEBUGDATA
+
+.. c:type:: entry_point
+
+    If CORHEADER_NATIVE_ENTRYPOINT is set, entry_point represents an RVA 
+    to a native entrypoint. If CORHEADER_NATIVE_ENTRYPOINT is not set, 
+    entry_point represents a managed entrypoint.
```
wxsBSD, line 56: `represents an RVA to a managed entrypoint.`

This way it is clear that in both cases it is an RVA.

docs/modules/dotnet.rst
```diff
@@ -168,6 +193,153 @@ Reference
...
+
+    The number of PInvoke implmaps in the file
+
+.. c:type:: impl_maps
+
+    A zero based array of impl_map table row. Each entry holds information 
+    about unmanaged methods that can be reached from managed code, using PInvoke 
+    dispatch. A row is entered in the impl_map table for each parent method that
+    is defined with a .pinvokeimpl interoperation attribute. Individual 
+    impl_maps can be accessed by using the [] operator.Each impl_map has the
```
wxsBSD, line 317:
```suggestion
    impl_maps can be accessed by using the [] operator. Each impl_map has the 
```

libyara/include/yara/dotnet.h
```diff
@@ -139,6 +234,29 @@ typedef struct _TILDE_HEADER
...
+  {
+    WORD Name_Short;
+    DWORD Name_Long;
+  } Name;
+  union
+  {
+    WORD Signature_Short;
+    DWORD Signature_Long;
+  } Signature;
+}MEMBERREF_TABLE, *PMEMBERREF_TABLE;
```
wxsBSD, line 258:
```suggestion
} MEMBERREF_TABLE, *PMEMBERREF_TABLE;
```

libyara/include/yara/dotnet.h
```diff
@@ -215,6 +333,30 @@ typedef struct _ASSEMBLYREF_TABLE
   } Name;
 } ASSEMBLYREF_TABLE, *PASSEMBLYREF_TABLE;
 
+//
+// Manifest Resource Table
+// ECMA-335 Section II.22.22
+//
```
wxsBSD, line 339: This seems to be a copy/paste mistake. :)

libyara/modules/dotnet/dotnet.c
```diff
@@ -1057,16 +1177,59 @@ void dotnet_parse_tilde_2(
...
+
+        set_integer(
+          yr_le16toh(implmap_table->MappingFlags), 
+          pe->object, 
+          "impl_maps[%i].mapping_flags",
+          i);
+
+        // Names can be NULL, but still add it to list. This will support
+        // tracing references to MemberForward and Import Scope. 
+        // PInvoke doesn't have to fill in the name if the MemberForwad
```
wxsBSD, line 1208:
```suggestion
        // PInvoke doesn't have to fill in the name if the MemberForward
```

libyara/modules/dotnet/dotnet.c
```diff
@@ -493,6 +498,53 @@ void dotnet_parse_tilde_2(
       row_size = (index_size + (index_sizes.string * 2));
       typeref_row_size = row_size;
       typeref_ptr = table_offset;
+      for (i = 0; i < num_rows; i++)
```
wxsBSD, line 501: I've seen situations in other places where we have a construct like this and it ends up being incorrect. For example, it is possible that we end up setting `number_of_typerefs` to be some non-zero value but never actually populate the array because the string lookups are always returning NULL for some reason. In those cases we have kept a counter to keep track of the number of elements in the array that is independent from the loop counter and use that when setting `number_of_typerefs`.

It has been a while since I wrote this code so I'm not sure if you're just copying my implementation when doing this or not, but it may be worth making a pass through this module to clean this up if it is indeed widespread. If that is the case, then I would suggest doing the cleanup in a follow-up PR.

tarterp: I looked over the four tables this PR is adding: memberref, implmap, typeref, and methoddef. The logic I used for adding a counter was if there is only a name being tracked for the table entry then `counter` was used, if other signatures could be made even with a name not present `i` was used. As mentioned a full sweep of the module could be good in a separate PR with discussion on what warrants the use of a counter with each table.

memberref: used `counter`, only `name` is being tracked for the table
implmap: used `i`, flags alone could be a signature even paired with a null entry name
typeref: used `counter`, only require that name or namespace has an entry to increment counter
methoddef: used `i`, methods also have `rva`, `impl_flags`, and `flags`.

tarterp (commented):
updated per comments given.

wxsBSD (approved):
Apologies for the delay, this completely slipped off my radar. I just took a quick look through this and it looks good to me!

plusvic:
Looks good to me. The only thing I miss is adding `.. versionadded:: 4.3.0` to the new fields in the documentation.

Issue #1667, opened 2022-03-26
make check hangs after all tests are complete


vt-alt:
`make check` hangs and it seems the cause is lost processess:
```
vt_b1     380755  0.0  0.0   3332  1248 ?        SN   14:08   0:00 /bin/sh -c VAR='Hello, world!'; sleep 600; true
vt_b1     380756  0.0  0.0   2652   956 ?        SN   14:08   0:00  \_ sleep 600
```
We run `make check` w/o `/proc` mounted. (There is different `test-rules` error if `/proc` is mounted.)

vt-alt:
Ah, after some investigation, it seems that RPM spec `%check` section hangs because of that lost processes, while `make check` itself does not hang,

vt-alt:
Well, second comment was too much guess and incorrect.

`make check` sometimes hangs, sometimes it isn't, perhaps it's race condition, and yes there are _several_ lost `sleep 600` processes (which affect our `rpmbuild` system too).

It seems we are forced to not run `make check` when building `yara` package for ALT Linux.

vt-alt:
> It seems we are forced to not run `make check` when building `yara` package for ALT Linux.

Yet, there is still a way:

`make -j32 check` — hangs
`make check` — does not hang.

So `make check` became not usable with parallel build (using `-j`) (in contrast to v4.1.3).

thesamesam:
Please re-test now that #1683 is merged.

plusvic:
#1683 doesn't seem to fix the issue with `make -j32 check` hanging, quite the contrary, it introduced the followed warning:

```
tests/test-rules.c:3304:44: warning: unknown escape sequence '\$' [-Wunknown-escape-sequence]
      "VAR='Hello, world!'; sleep 600& PID=\$!; trap \"kill \$PID\" EXIT; "
```

PR #1683, opened 2022-04-08
finish all processes spawned in the test suite

hamarituc:
This PR fixes a test case to ensure all spawned processes are finished. In `tests/test-rules.c` a shell process is created which itself spawns the `sleep`-executable. The latter isn't killed after the test run which confuses some automated test suites an leads to timeout issues (see https://bugs.gentoo.org/836790).

### DIFF
--- a/tests/test-rules.c
+++ b/tests/test-rules.c
@@ -3061,7 +3061,7 @@ void test_process_scan()
     exit(EXIT_FAILURE);
   }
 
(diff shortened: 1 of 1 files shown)


thesamesam:
It looks like this will fix https://github.com/VirusTotal/yara/issues/1667 too.

plusvic (approved):

PR #1685, opened 2022-04-12
Add function to parse rules from raw buffer

1ndahous3:
There are some cases where we need to load rules not only from files, but directly from memory. One of these is updating over the network and (re)parsing without dropping to disk.
`yr_compiler_add_string()` is not suitable because buffers in-memory are not null-terminated, and `strlen()` is time consuming.

Here we add the ability to work with the most portable form of raw data: pointer + size.

### DIFF
--- a/libyara/compiler.c
+++ b/libyara/compiler.c
@@ -629,6 +629,31 @@ YR_API int yr_compiler_add_fd(
   return result;
 }
 
+YR_API int yr_compiler_add_bytes(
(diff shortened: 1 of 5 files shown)


1ndahous3:
@plusvic can you merge this PR? This is not even a "feature", but just one "wrapper" function that uses an already existing API (one in the lexer and one in the compiler).

plusvic (approved):

PR #1691, opened 2022-04-23
Fixed lexer pattern for bad character set

### LINKED ISSUE #1690: RE Lexer improperly parsing dashes after character classes
When using a regular expression containing a character class, followed by a dash `-`, followed by anything else, Yara fails with the following error:
```
$ /usr/bin/yara -s ~/testrule.yara ~/yara_test_file.txt 
~/testrule.yara(3): error: invalid regular expression "$": bad character range
```

It appears this error message is related to the following section of the RE lexer, where it doesn't properly look for the ending square bracket of the character range and therefore over-matches on the dash outside.
https://github.com/VirusTotal/yara/blob/7e00915f1594b60fa19b39f29f77c046c99b14fe/libyara/re_lexer.l#L311

This is the test rule I'm using, and the test file being scanned.
```
$ cat ~/testrule.yara 
rule test {
        strings:
                $ = /[0-9]-2/
        condition:
                all of them
}
```
```
$ cat ~/yara_test_file.txt 
1-2-3-4-5
```

Using grep, regex101, or similar, we can confirm the regex works as expected and is valid.
```
$ cat ~/yara_test_file.txt | grep -o "[0-9]-2"
1-2
```

Sevaarcen:
Fixes an issue with the lexer pattern for character ranges to now no-longer allow an unescaped closing square bracket inside a character set.

I'm using the following test rule, and test data.
```
$ cat ~/testrule.yara 
rule test {
        strings:
                $ = /[0-9]-2/
        condition:
                all of them
}
```
```
$ cat ~/yara_test_file.txt 
1-2-3-4-5
```

Using a version of Yara without a fix, Yara improperly gives an error for the pattern in the rule above.
```
$ /usr/bin/yara --version
3.9.0

/usr/bin/yara -s ~/testrule.yara ~/yara_test_file.txt 
~/testrule.yara(3): error: invalid regular expression "$": bad character range
```

With my changes, Yara appears to be working as expected
```
./yara -s ~/testrule.yara ~/yara_test_file.txt
test /home/cyberdraco/yara_test_file.txt
0x0:$: 1-2
```

Closes #1690

### DIFF
--- a/libyara/re_lexer.l
+++ b/libyara/re_lexer.l
@@ -308,7 +308,7 @@ hex_digit     [0-9a-fA-F]
 
 
 
-<char_class>(\\x{hex_digit}{2}|\\.|[^\\])\-[^]] {
+<char_class>(\\x{hex_digit}{2}|\\.|[^\\\]])\-[^]] {
 
   // A range inside a character class.
   //  [abc0-9]


plusvic:
This issue was already fixed in a91b8a1eece43f8e32af8988d6d392c335df019f and ee96a35dad4b0b4937e44a3eb8e9fd53c662a777. I didn't notice this PR with until know, but your solution was 100% correct.

PR #1698, opened 2022-04-28
Fixed evaluation of none keyword when the evaluated context is true

metthal:
I think I have accidentally found serious bug with `none` specifier which seems to be broken.

`none` keyword is pushing 0 onto stack during the runtime as a constant that is compared against the number of iterations evaluated to true. Since it's 0, it'll always evaluate to true during the first iteration and therefore when using `none` keyword, your loop cannot ever evaluate to false.

This fix treats `none` as `any` with the negation of the result at the end. This way, `none` keyword benefits from short-circuit evaluation of `any`. If at least 1 expression evaluates to `true`, the loop is immediately terminated and result is inverted to `false`. If all the expressions are getting evaluated to `false`, we need to iterate over the whole iterator and at the end, invert the final `false` to `true`.


plusvic (approved):

plusvic:
Closing in favor of #1697 which solves the same issue.

PR #1701, opened 2022-05-01
Unification of for-loops with the same codegen, fixing of their short-circuit evaluation and usage of none with them

metthal:
`for .. in` and `for .. of` loops now use the same codegen for their bytecode using 2 instructions `ITER_CONDITION` and `ITER_END`. `ITER_CONDITION` evaluates each iteration whether the condition of the loop was satisfied prematurely or whether the loop is still satisfiable, so essentially making sure that loops are properly short-circuited. `ITER_END` is used for final evaluation of the loop, whether it is satisfied or not, after the iterator got exhausted or the loop was short-circuited.

In order to be able to use single codegen for both iterators and string sets, new iterator type was instroduced - `YR_ITERATOR_STRING_SET`. When `ITER_START_STRING_SET` is executed, it is capable of consuming all strings pushed on the stack and construct `YR_ITERATOR_STRING_SET` object which can then be used as a regular iterator.

Currently, I'm not completely satisfied with the fact that the logic for short-circuiting and special handling is separate for `OF`, `OF_FOUND_IN`, `OF_PERCENT` and these for-loops. It would be the best if each of them used these new `ITER_CONDITION` and `ITER_END` instructions but I'm not sure whether someone else shares this sentiment with me.

Before I proceed with any further improvements of existing instructions, I wanted to know what would you tell me about this and if there's anything I should improve before this idea is brought even further.


plusvic:
Great PR!. My only comment is that it requires incrementing `YR_ARENA_FILE_VERSION` in `arena.h`, so that compiled rules generated by this version won't be loaded by older versions of YARA that won't recognize the new instructions.

I also like the idea of unifying `OF`, `OF_FOUND_IN` and `OF_PERCENT` with this `for` loops as much as possible.

plusvic (approved):

Issue #1706, opened 2022-05-10
YR_MICRO_VERSION is still 0 (should be 1)


tlansec:
Looks like the newest version of yara has incorrect YR_MICRO_VERSION value specified meaning --version and --help information is inaccurate on terminal.

`https://github.com/VirusTotal/yara/blob/master/libyara/include/yara/libyara.h`

xambroz:
This is also in configure.ac 
AC_INIT([yara], [4.2.0], [vmalvarez@virustotal.com])

plusvic:
Fixed in #1741

PR #1707, opened 2022-05-11
incrementing the minor version of yara to reflect current ver #

tlansec:
YR_MINOR_VERSION is still 0 which doesnt reflect current versioning.

Linked to and/or the sames as: https://github.com/VirusTotal/yara/issues/1706


r0ny123:
fixed with #1741

PR #1709, opened 2022-05-16
Fix some compiler edge cases.

wxsBSD:
Add a warning when "0 of them" is used. Point to some new documentation that
explains why "none of them" is preferred.

Handle some edge cases, which are all now errors:

* -1 of them
* "foo" of them
* /foo/ of them

And this one is also an error if the external variable "x" is a negative integer
or a string:

* x of them

I've also made sure this works with a construct like "0 + x of them" where x is
defined as a negative integer.

Add test cases for as many of these as I can. I don't have test cases for the
external variable case, but it does work as expected. Here is the output of a
rule using external variables where the variable is defined to be 1, 0, -1 and
foo:

```
wxs@mbp yara % cat rules/test.yara
rule a {
  strings:
    $a = "FreeBSD"
  condition:
    x of them
}
wxs@mbp yara % ./yara -d x=1 rules/test.yara /bin/ls
a /bin/ls
wxs@mbp yara % ./yara -d x=0 rules/test.yara /bin/ls
warning: rule "a" in rules/test.yara(5): Consider using "none" keyword, it is less ambiguous. Please see https://yara.readthedocs.io/en/stable/writingrules.html#sets-of-strings-1 for an explanation.
wxs@mbp yara % ./yara -d x=-1 rules/test.yara /bin/ls
error: rule "a" in rules/test.yara(5): invalid value in condition: "-1"
wxs@mbp yara % ./yara -d x=foo rules/test.yara /bin/ls
error: rule "a" in rules/test.yara(5): invalid value in condition: "string in for_expression is invalid"
wxs@mbp yara %
```

### DIFF
--- a/libyara/grammar.y
+++ b/libyara/grammar.y
@@ -2279,6 +2279,40 @@ rule_enumeration_item
 for_expression
     : primary_expression
       {
+        if ($1.type == EXPRESSION_TYPE_INTEGER) {
+          if ($1.value.integer == 0) {
+            yywarning(yyscanner,
+                "Consider using \"none\" keyword, it is less ambiguous. Please "
+                "see https://yara.readthedocs.io/en/stable/writingrules.html#sets-of-strings-1 for an explanation.");
+          }
+
+          if ($1.value.integer < 0) {
+            yr_compiler_set_error_extra_info_fmt(compiler,
+                "%lld", $1.value.integer);
+            fail_with_error(ERROR_INVALID_VALUE);
+          }
+        }
+
+        if ($1.type == EXPRESSION_TYPE_STRING) {
+          SIZED_STRING* ss = yr_arena_ref_to_ptr(compiler->arena,
+              &$1.value.sized_string_ref);
+          // If the expression is an external string variable we need to get
+          // it some other way.
+          if (ss != NULL) {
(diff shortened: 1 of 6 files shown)


libyara/grammar.y
```diff
@@ -2279,6 +2279,40 @@ rule_enumeration_item
 for_expression
     : primary_expression
       {
+        if ($1.type == EXPRESSION_TYPE_INTEGER) {
+          if ($1.value.integer == 0) {
+            yywarning(yyscanner,
+                "Consider using \"none\" keyword, it is less ambiguous. Please "
```
plusvic, line 2285: Start errors and warning with lowercase characters. 

Also, even though the link may be useful, I'm not sure about including it. This warning will be shown in contexts in which a long plain-text link may be cause issues. I'm thinking about VT's web interface for example. So, I would keep the link out of the message.

wxsBSD:
I fixed the message and also added a bunch more warnings and one more error. Specifically these are now warnings:

`2 of ($a)` is a warning because it will always evaluate to false and is likely not what the user meant to happen. The same is true for rule sets and strings in a range like `2 of ($a) in (0..10)`.

In ranges, if we can statically determine the bounds it is now an error if the lower bound is greater than the upper bound.

PR #1716, opened 2022-05-26
Update atom logic to properly score [A-Z]

BitsOfBinary:
The [documentation says](https://github.com/VirusTotal/yara/blob/ee6b2363f7d3bcb6b6bbe655ad8a71a451cc3af8/libyara/atoms.c#L118) `Bytes [a-zA-Z] contribute 18 points each`, but it looks like [in the code](https://github.com/VirusTotal/yara/blob/ee6b2363f7d3bcb6b6bbe655ad8a71a451cc3af8/libyara/atoms.c#L164) that only `[a-z]` is given 18 points, whereas `[A-Z]` is given 20 points. This commit will make sure these ranges have the proper atom scoring applied.

### DIFF
--- a/libyara/atoms.c
+++ b/libyara/atoms.c
@@ -161,8 +161,8 @@ int yr_atoms_heuristic_quality(YR_ATOMS_CONFIG* config, YR_ATOM* atom)
         // than the rest. We want to favor atoms that contain bytes outside
         // those ranges because they generate less additional atoms during
         // calls to _yr_atoms_case_combinations.
-        if (yr_lowercase[atom->bytes[i]] >= 'a' &&
-            yr_lowercase[atom->bytes[i]] <= 'z')
+        if ((yr_lowercase[atom->bytes[i]] >= 'a' && yr_lowercase[atom->bytes[i]] <= 'z') ||
+        (yr_lowercase[atom->bytes[i]] >= 'A' && yr_lowercase[atom->bytes[i]] <= 'Z'))
           quality += 18;
         else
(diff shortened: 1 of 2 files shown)


libyara/atoms.c
```diff
@@ -161,8 +161,8 @@ int yr_atoms_heuristic_quality(YR_ATOMS_CONFIG* config, YR_ATOM* atom)
         // than the rest. We want to favor atoms that contain bytes outside
         // those ranges because they generate less additional atoms during
         // calls to _yr_atoms_case_combinations.
-        if (yr_lowercase[atom->bytes[i]] >= 'a' &&
-            yr_lowercase[atom->bytes[i]] <= 'z')
+        if ((yr_lowercase[atom->bytes[i]] >= 'a' && yr_lowercase[atom->bytes[i]] <= 'z') ||
+        (yr_lowercase[atom->bytes[i]] >= 'A' && yr_lowercase[atom->bytes[i]] <= 'Z'))
```
plusvic, line 165: This change has no effect because `yr_lowercase[atom->bytes[i]]` will never be in the range `A..Z`. The lookup table `yr_lowercase` converts any `atom->bytes[i]` in the range `A..Z` to its lowercase counterpart. In 25f576adf3094061fe5c6ba87c103b460ecb6171 I've added your test case to the original code, and it passes without any other changes.

BitsOfBinary: Thanks @plusvic, sorry I hadn't noticed that `yr_lowercase` routine!

PR #1723, opened 2022-06-01
Extra docs for compiling Windows binaries

BitsOfBinary:
Added some extra documentation to detail how to use `configure` to compile Windows binaries.

### DIFF
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -106,10 +106,13 @@ To install YARA using `Scoop <https://scoop.sh>`_ or `Chocolatey <https://chocol
 ``scoop install yara`` or ``choco install yara``. The integration with both `Scoop` and `Chocolatey` are
 not maintained their respective teams, not by the YARA authors.
 
-To compile Windows binaries from source, you can supply extra commands to the ``configure`` binary. 
-For example, the following command will compile 64-bit versions of the YARA executables::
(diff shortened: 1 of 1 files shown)


plusvic:
Saying that you can run `./configure --host=x86_64-w64-mingw32` in Windows, without explaining which are the pre-requisites are going to be confusing for many people that don't know about mingw32, so I would explain that a little bit about how to setup mingw32 in your machine.

BitsOfBinary:
> Saying that you can run `./configure --host=x86_64-w64-mingw32` in Windows, without explaining which are the pre-requisites are going to be confusing for many people that don't know about mingw32, so I would explain that a little bit about how to setup mingw32 in your machine.

That's a fair comment! My setup was so that I could compile Windows binaries on Linux (using WSL). 

Main reason I want to add these to the docs is that if you're developing YARA on Windows, it wasn't immediately clear on how to compile it from source (I had to check some of the GitHub actions to get some context). Hopefully this addition will make it clear on a method of doing this, but happy to keep tweaking it until it is as clear as possible?

plusvic:
Sure, a more detailed explanation that covers how to build YARA on Windows using Mingw32 is welcome.

BitsOfBinary:
Closing this off as I haven't figured out how to quite make this work yet...

PR #1732, opened 2022-06-19
LNK module

BitsOfBinary:
This module will parse the Windows Shell Link (LNK) file format, and make a lot of it's data accessible via YARA. The motivation for creating this module is that the LNK file format is non-standard to parse, and would be difficult to do so within a YARA rule itself (see the documentation here: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-shllink/16cb4ca1-9339-4d0c-a68d-bf1d6cc0f943). I hope by making this module that it makes life easier to write YARA rules for LNK files.

This module allows for rules like this to be possible:
```
import "lnk"

rule is_lnk
{
    condition:
        lnk.is_lnk
}

rule machine_id_tracking
{
    condition:
        lnk.tracker_data.machine_id == "chris-xps"
}

rule local_base_path
{
    condition:
        lnk.link_info.local_base_path == "C:\\test\\a.txt"
}
```

I'd be very interested to hear feedback on the code itself, or the structure of the data surfaced by the module (e.g. is it easy to access/use).


libyara/modules/lnk/lnk.c
```diff
@@ -0,0 +1,2339 @@
...
+    end_struct_array("item_id_list");
+  end_struct("vista_and_above_id_list_data");
+
+  declare_integer("has_overlay");
+  declare_integer("overlay_offset");
+
+  declare_integer("is_malformed");
+end_declarations
+
+unsigned int parse_id_list(
```
plusvic, line 344: Better use `uint32_t` or `uint64_t`. The size of `unsigned int` depends on the platform and this may cause issues while compiling on 32-bits systems. The same applies to the rest of the functions.

libyara/include/yara/lnk.h
```diff
@@ -0,0 +1,293 @@
...
+
+#pragma pack(pop)
+
+#define HEADER_SIZE  0x0000004C
+#define LINK_CLSID_0 0x00021401
+#define LINK_CLSID_1 0x00000000
+#define LINK_CLSID_2 0x000000C0
+#define LINK_CLSID_3 0x46000000
+
+#define HasLinkTargetIDList         0x00000001
```
plusvic, line 139: Some constants use snake case all in uppercases, but these ones use camel-case. Should we use the same naming style for all constants? Are these constants named this way because of something?

libyara/modules/lnk/lnk.c
BitsOfBinary: Thanks, I've updated this throughout!

libyara/include/yara/lnk.h
BitsOfBinary: Good question, I named some of them based off the casing Microsoft used in their docs (e.g. `HasLinkTargetIDList` has that casing in the LNK documentation - https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-SHLLINK/%5bMS-SHLLINK%5d.pdf).

I'm happy to make this consistent throughout however!

BitsOfBinary: Updated in https://github.com/VirusTotal/yara/pull/1732/commits/5de9c2c1dd26795398d4db3309cc89e497bcce87

BitsOfBinary:
[fuzz_lnks.zip](https://github.com/VirusTotal/yara/files/9905114/fuzz_lnks.zip)
^ I've added the LNKs I used when doing some fuzzing against this module, in case it is helpful!

BitsOfBinary:
Updated this branch to be compatible with YARA 4.3.0. At time of this comment, all tests pass as expected.

BitsOfBinary:
Updated for compatibility with YARA v4.3.2; tests passing as expected.

plusvic (approved):

plusvic:
I tried to merge this PR but it turns out that the tests are failing in big endian platforms: https://github.com/VirusTotal/yara/actions/runs/5960808590/job/16168792006

BitsOfBinary:
> I tried to merge this PR but it turns out that the tests are failing in big endian platforms: https://github.com/VirusTotal/yara/actions/runs/5960808590/job/16168792006

That's weird @plusvic , they seemed to be working before (https://github.com/VirusTotal/yara/actions/runs/5772402091/job/15647348631). I'll debug what is failing in the tests and get back to you.

PR #1767, opened 2022-08-15
Add stringtoint() and inttostring().

wxsBSD:
Add math.stringtoint() and math.inttostring() which let you convert between an
integer and a string representation of it, and back.

### DIFF
--- a/libyara/modules/math/math.c
+++ b/libyara/modules/math/math.c
@@ -28,7 +28,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 #include <stdlib.h>
-#include <sys/errno.h>
+#include <errno.h>
 #include <math.h>
 #include <yara/mem.h>
 #include <yara/modules.h>
--- a/libyara/modules/math/math.c
+++ b/libyara/modules/math/math.c
@@ -37,6 +37,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define MODULE_NAME math
 
 #define PI 3.141592653589793
+// This is more than enough space to hold the maximum signed 64bit integer as a
+// string in decimal, hex or octal, including the sign and NULL terminator.
+#define INT64_MAX_STRING 30
 
 // log2 is not defined by math.h in VC++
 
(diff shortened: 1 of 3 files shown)


wxsBSD:
If this looks good I'll add docs before merging.

libyara/modules/math/math.c
```diff
@@ -738,6 +787,10 @@ begin_declarations
   declare_function("percentage", "i", "f", percentage_global);
   declare_function("mode", "ii", "i", mode_range);
   declare_function("mode", "", "i", mode_global);
+  declare_function("inttostring", "i", "s", inttostring);
```
plusvic, line 790: What about naming the functions `math.to_int` and `math.to_str`? If not, I would follow the snake case convention `int_to_string` and `string_to_int`, or (`int_to_str` and `str_to_int` in abbreviated form)

libyara/modules/math/math.c
```diff
@@ -715,6 +717,53 @@ define_function(mode_global)
...
+  case 8:
+    fmt = "%llo";
+    break;
+  case 16:
+    fmt = "%llx";
+    break;
+  default:
+    return_string(YR_UNDEFINED);
+  }
+  asprintf(&str, fmt, i);
```
plusvic, line 748: address-sanitizer has found a memory leak here, `asprintf` allocates a memory buffer that is never released. `return_string` makes its own copy of the string, that's why in other places where `return_string` is used, it receives a pointer to a local variable in the stack. We would use a large enough local buffer that holds the number in string form and usee `sprintf` instead.

wxsBSD: Good catch! I misread the `return_string` macro. I am now allocating enough space on the stack and using `snprintf()`.

libyara/modules/math/math.c
```diff
@@ -27,6 +27,8 @@ ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
+#include <stdlib.h>
+#include <sys/errno.h>
```
plusvic, line 31: `#include <errno.h>` should be enough, `#include <sys/errno.h>` is not being found and CI builds are failing.

PR #1768, opened 2022-08-17
Added RVA to the import details Function Details Structure

xorhex:
Added the import function's RVA to the import details functions structure.

Can use it like this:

```
import "pe" 
      rule test_rule 
      {
          condition:
            for any import_detail in pe.delayed_import_details: (
                import_detail.library_name == "USER32.dll" and
                for any function in import_detail.functions : (
                    function.name == "MessageBeep" and
                    function.rva == 0x13880 
                )
            )
      }
```

The goal of this is to work towards: https://twitter.com/xorhex/status/1528604721870540800?s=20&t=hruupMGNXyxgvoweJMv0Tg

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -1333,7 +1333,7 @@ Reference
 
         .. c:member:: rva
 
-            .. versionadded:: 4.2.?
+            .. versionadded:: 4.3.0
 
             Relative virtual address (RVA) of imported function. If rva not found then this value is YR_UNDEFINED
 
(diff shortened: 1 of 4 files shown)


plusvic:
The code is ok to me, would you mind to add this new field to the module's documentation in https://github.com/VirusTotal/yara/blob/master/docs/modules/pe.rst?

xorhex:
Field added to module's documentation.  Was unsure which YARA version number should go here: `.. versionadded:: 4.2.?`

PR #1770, opened 2022-08-20
Reject non ascii bytes in regex

vthib:
As far as I know, regexes are not supposed to accept non ascii bytes. This is already rejected inside a char class (so `[<non_ascii_char>]` is rejected), but this is not properly rejected outside of a char class.

This means it is possible to write something like `$foo = /fé+e/` which definitely does not work as expected, even without considering the dependency on the rule encoding.

This PR fixes this by moving the ascii check in the right lex pattern, and adds two quick tests on this.

### DIFF
--- a/libyara/re_lexer.c
+++ b/libyara/re_lexer.c
@@ -1267,70 +1267,77 @@ YY_RULE_SETUP
 {
 
   // Any non-special character is passed as a CHAR token to the scanner.
-
-  yylval->integer = yytext[0];
-  return _CHAR_;
+  if (yytext[0] >= 32 && yytext[0] < 127)
+  {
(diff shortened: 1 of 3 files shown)


plusvic (approved):

plusvic:
I've reverted this change because there are many rules out there that rely on regexp that contains non-ascii characters. In most cases the non-ascii character is not accompanied of quantifiers like `+` or `*` and they work as long as the source code is in the same encoding as the file being scanned. For example, a rule in UTF-8 with the regexp `fóo` will match a file containing `fóo` if it appears as UTF-8.

For the time being I prefer being cautious about this changes that may break existing rules that actually work, even if the behaviour is completely undefined.

PR #1771, opened 2022-08-20
fix computation of serial correlation factor scct1

vthib:
for a sequence `a0, a1, ..., aN``, scct1 is:

`a0 * a1 + ... + a(n-1) * aN + aN * a0`

This was badly computed, as instead of computing aN * a0, aN * aN was
added instead.

This was I guess missed in the tests, because in the only test on
serial_correlation, the string used has a0 = aN.

### DIFF
--- a/libyara/modules/math/math.c
+++ b/libyara/modules/math/math.c
@@ -319,6 +319,7 @@ define_function(data_serial_correlation)
   YR_MEMORY_BLOCK_ITERATOR* iterator = context->iterator;
 
   double sccun = 0;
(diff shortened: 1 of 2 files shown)


wxsBSD (approved):
Wow, what a great find and thank you for the submission! I think this is correct.

I was able to test it against https://fourmilab.ch/random/ - which is where the idea for a lot of this originally came from - and it was indeed incorrect before. The tests now match what ent is saying for the same data.

plusvic (approved):

PR #1779, opened 2022-08-30
Add math.length()

### LINKED ISSUE #1778: [feature request] - Length of object
I have recently come across a couple of scenarios where I want to calculate the length of a feature exposed by a module, but no length field is calculated ahead of time by YARA.

It's nice to have these defined up front in cases like number_of_exports, but more generically it would be nice if there were a math.length() or similar function which calculates the length of objects. 

In my particular case, I want to use the fact that a particular .NET packer has a constant in position 0 containing an encrypted copy of the payload which is extremely large, so what i want to do is:

```yara
condition: 
    length(dotnet.constants[0]) > 16000 and 
    ...[some other conditions]
```

...but there isnt really a reasonable way to do this at present (doing it with a regex is very expensive).

wxsBSD:
Add a math.length() which will return the length of the sequence of bytes,
including any NULL bytes.

Fixes #1778.


plusvic:
I find a bit weird that this is included in the `math` module. Maybe it's time for a `string` module?

wxsBSD:
Sure, I'll do it today.

wxsBSD:
I decided to move the `to_int()` functions from the math module over to the new string module as it made more sense to me to convert to an integer via `string.to_int()` instead of `math.to_int()`. If you want to convert the other way around you still use `math.to_string()`.

wxsBSD:
Merged in the `next` branch.

PR #1782, opened 2022-09-01
Fixed compilation warning in snprintf() with 64-bit int

metthal:
It's not guaranteed that `int64_t` is always `long long`. It's better to use `PRI` defines.


plusvic (approved):

wxsBSD:
I think this should also be fixed in `to_string_base()`.

Thanks for catching this though.

metthal:
Thanks for letting me know. My compiler was kind of lazy and didn't give me warning for that one :) Fixed it.

Issue #1765, opened 2022-08-10
Iterating over constant strings in yara conditions block


nikhilh-20:
Hello,

I recently had a use-case where I wanted to have a list of strings in the `conditions` block in a yara rule. A minimal example is: comparing the SHA1 of a file with a known set of hashes. This is a working yara rule:

`test.yar`:

```
import "hash"

rule test {
    condition:
        hash.sha1(0,filesize) == "0a9fbc6dacd8887bd9d9065bc7d9a4905d7ea687" or
        hash.sha1(0,filesize) == "90cf88f9f3326d2616232d73e5adc1e85d28097f"
}

$ yara test.yar test.txt 
test test.txt
```

For readability, this is alright when there are, say, 5 hashes. But if there are 100 hashes, then the rule file becomes messy. Is there a way in yara to have non-search strings in one variable? And I'm not referring to the `strings` block because those strings are searched in the file. I'm referring to a string variable (containing a list of strings) which is not searched in the file and can purely be used in the `conditions` block.

Something like:

```
import "hash"

rule test {
    condition:
        for any i in ("90cf88f9f3326d2616232d73e5adc1e85d28097f","0a9fbc6dacd8887bd9d9065bc7d9a4905d7ea687"):
            ( i == hash.sha1(0,filesize) )
}
```

The above doesn't work:

```
$ yara test.yar test.txt 
test.yar(5): error in rule "test": wrong type for enumeration item
```

Thanks for the help!

plusvic:
I think it makes sense that iterators work on lists of strings too. It's more intuitive than a long sequence of `or` statements.

wxsBSD:
I've taken a shot at this and am having a hard time getting it to work correctly in the grammar. It seems there is a conflict with the way integer sets are built.

Also there is https://github.com/VirusTotal/yara/pull/1338, which might make this issue obsolete. As such, I'm going to put this on hold until I can come up with a way to make it work or someone with more knowledge than me can do it.

wxsBSD:
After being shown the path by @plusvic I've got this implemented and up for review now.

Thanks, again, Victor!

PR #1787, opened 2022-09-09
Implement text string sets.

wxsBSD:
Add support for text string sets into the grammar. They look like this:

for any s in ("a", "b"): (pe.imphash() == s)

This requires changing integer_set and integer_enumeration to just be set and enumeration, and adding a new type (YR_ENUMERATION) that tracks the type of enumeration (integer or otherwise) and the number of items in the enumeration.

The enumeration now checks that each type of the item in the enumeration is consistent and will cause a compiler error if they are not. For example, this is an error:

for any s in ("a", 0): (s)

Also, fix the build when using --with-debug-verbose option as it was missing the assert.h include.


wxsBSD:
Closes #1765.

wxsBSD:
Forgot to mention that I'll update docs before merging if this looks good.

libyara/include/yara/exec.h
```diff
@@ -123,6 +123,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define OP_OF_FOUND_IN           75
 #define OP_COUNT_IN              76
 #define OP_DEFINED               77
+#define OP_ITER_START_TEXT_STRING_SET 78
```
plusvic, line 126: nit: Re-align constants in the lines above with the `78` in this line.

plusvic:
Looks good to me.

plusvic (approved):

plusvic:
Waiting for doc update before merging.

PR #1790, opened 2022-09-17
Add OP_OF_FOUND_AT.

wxsBSD:
Add support for "any of them at 0" constructs to the language. This allows users to avoid using long or chains like "$a at 0 or $b at 0" and also is a nicer way
to write "for any of them: ($ at 0)".


wxsBSD:
This is mostly a syntactic sugar PR because someone at work asked if there was a nicer way than long `or` chains like `$a at 0 or $b at 0...`.

If this is a good idea I'll update the docs before merging.

plusvic:
I think this is good idea, also for completeness and coherency, because we have `$a in (0..100)`, `$a at 0` and `any of them in (0..100)`, the missing piece is `any of them at 0`.

plusvic (approved):

djlukic:
I was looking for this.
So, if I understood correctly we'll be able to do this

```
$a1 = "MZ"
$a2 = "PE"
```

condition would be

`any of ($a*) at 0x80`

Thanks!

djlukic:
I wanted to ask one more question about this.

`$a1 = "MZ"
$a2 = "PE"`

this condition doesn't work

`($a1 or $a2) at 0x80`

this does work
`$a1 at 0x80
or $a2 at 0x80`

Apologies if this was covered in some other topic.

wxsBSD:
You want to use `any of ($a1, $a2) at 0x80` or the shorter form `any of ($a*) at 0x80` which will be true if at least one of the `$a*` strings are found at that offset.

The idea of string sets is covered here: https://yara.readthedocs.io/en/stable/writingrules.html#sets-of-strings-1 (though the `at` portion of it is not in a released version yet).

djlukic:
Thank you @wxsBSD. Yes, I went through the documentation many times.
If I understand correctly this is a correct way but that is not released yet?

> You want to use `any of ($a1, $a2) at 0x80` or the shorter form `any of ($a*) at 0x80` which will be true if at least one of the `$a*` strings are found at that offset.

wxsBSD:
> Thank you @wxsBSD. Yes, I went through the documentation many times. If I understand correctly this is a correct way but that is not released yet?

Correct. `any of ($a*) at 0x80` is coming in 4.3.0. For now, you can use `$a1 at 0x80 or $a2 at 0x80`.

If you have a lot of strings you want to check at a specific offset you can do it with a loop: `for any of ($a*): ($ at 0x80)`

4.3.0 adds the shorter syntax version than the loop.

PR #1795, opened 2022-09-21
Export Hash (exphash)

LloydLabs:
Much like an imphash, an exphash is simply a MD5 hash of the exports defined in the Export Address Table. This is helpful for comparing PE files which export functions, which can then be compared to others. Found this useful when hunting for DLLs used in DLL-hijacking etc.

If no exports are found, `YR_UNDEFINED` simply returned.

```
import "pe"

rule test_exphash
{
	condition:
		pe.is_dll() and pe.exphash() == "a52adfc0598657d621ede8248dd0ea80"
}
```

### DIFF
--- a/docs/modules/pe.rst
+++ b/docs/modules/pe.rst
@@ -1408,6 +1408,15 @@ Reference
 
     *Example: pe.imphash() == "b8bb385806b89680e13fc0cf24f4431e"*
 
+.. c:function:: exphash()
+
+    .. versionadded:: 4.2.3
+
+    Function returning the export hash or exphash for the PE. The export hash
+    is a MD5 hash of the PE's export table.
+
+    *Example: pe.exphash() == "69e300c67958698c5c147b2b563317bc"*
+
 .. c:function:: section_index(name)
 
     Function returning the index into the sections array for the section that has
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -2699,6 +2699,98 @@ define_function(imphash)
   return_string(digest_ascii);
 }
 
+///
+/// Generate a hash of the imports
+/// Similar to imphash, makes use of export_details already in cache
+///
+define_function(exphash)
+{
+  YR_OBJECT* module = yr_module();
+
+  yr_md5_ctx ctx;
+
+  unsigned char digest[YR_MD5_LEN];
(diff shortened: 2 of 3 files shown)


LloydLabs:
Just a quick note on AppVeyor; for some reason, the unit tests aren't passing for Windows. In my tests on *nix, they passed fine. Not sure what's going on, cc @wxsBSD or @plusvic might know?

ostracon (approved):

ostracon:
Minor update and all tests are passing.

hillu:
The function description should read "Generate a hash of the imports".

`yr_get_integer()` and `yr_get_string()` may return `YR_UNDEFINED` or
`NULL`, respectively. It would make sense to add checks for those
unlikely values, bailing out of the function.

LGTM, otherwise.

LloydLabs:
> The function description should read "Generate a hash of the imports".
> 
> `yr_get_integer()` and `yr_get_string()` may return `YR_UNDEFINED` or `NULL`, respectively. It would make sense to add checks for those unlikely values, bailing out of the function.
> 
> LGTM, otherwise.

Thanks! This is how the `imphash` implementation works too, I'll keep it like this for now unless a reviewer wants to change it.

N0fix:
> Much like an imphash, an exphash is simply a MD5 hash of the exports defined in the Export Address Table. 

As described in this [blog post](https://blog.syscall.party/2023/03/03/introducing-exphash-for-dll-clustering.html) (from which someone links this PR) and in your implementation in [pefile](https://github.com/erocarrera/pefile/blob/593d094e35198dad92aaf040bef17eb800c8a373/pefile.py#L5903), exphash is calculated using SHA256. Is there a reason why md5 would be used in Yara over sh256? 
Having two different ways of calculating exphash looks like a bad idea to me.

N0fix:
As described in this [blog post](https://blog.syscall.party/2023/03/03/introducing-exphash-for-dll-clustering.html) and in your i[mplementation in pefile](https://github.com/erocarrera/pefile/blob/593d094e35198dad92aaf040bef17eb800c8a373/pefile.py#L5903), exphash is calculated using SHA256. Is there a reason why md5 would be used in Yara over sh256? 
Having two different ways of calculating exphash looks like a bad idea to me.

Issue #1780, opened 2022-08-30
(Built-in) function to read bytes at a specific offset


metthal:
Recently I was approach by someone who'd like to compute a hash on a specific part of the file and then compare that hash against bytes at a specific offset. The first part is easy `hash.md5(...)` but then you don't really have any option what to compare it against. Generally, you can say that comparing functions that return strings is hard against content of the file.

That's why I have been thinking about having `bytes(offset, length)` function that would return you a sequence of bytes from the file. I was almost about to implement it but then @wxsBSD raised a good question. Should it be built-in function like `uintXX()` functions are or should it be a module function?

I can't really see a benefit in either of the options so I'm asking here because I don't really want to rewrite the code after PR review and would like to get it to upstream ASAP.

wxsBSD:
My initial thoughts were to make it a module function because it means it is much easier to deal with in gyp and any other parsers out there. However, making it live in a module might be more difficult because I don't believe there is a way to return a `SIZED_STRING` from a module function, which means if your sequence of bytes has a `\x00` then you will have problems. I could see it being a builtin because it does fit nicely with the `uintXX()` functions and resolves the problem of embedded null bytes. On balance, I like it being a builtin more because it does fit nicely there.

If you make it a module you also have the problem of how to not leak memory. If the module function does:

```
// You could do this with a SIZED_STRING* too, but just using this for simplicity sake.
void *p = malloc(length);
memcpy(p, offset, length);
return_bytes(p);
```

Then we need to be careful to eventually `free(p)`. It isn't clear to me how we could do that easily with the existing design. Not saying it is impossible but is something to consider.

If it is a builtin you could add a way to keep track of any allocations made during rule evaluation by the VM and free them in the execution of `OP_MATCH_RULE`, which might be a cleaner solution?

Regardless of how you implement it, an obvious followup is having equivalents of the python functions `binascii.hexlify()` and `binascii.unhexlify()` as a `math` module function. This would allow you to do `math.hexlify(0, 16) == hash.md5(bytes(16, 500))` to check the MD5 of some sequence of bytes against the hexlified contents at the start of the file. The inverse could be `bytes(0, 16) == math.unhexlify(hash.md5(bytes(16, 500)))`.

Obviously it doesn't need to be done with this proposal but it would certainly make the feature all that more powerful.

[...]

PR #1797, opened 2022-09-30
Implement bytes keyword.

wxsBSD:
This commit adds the "bytes" keyword to the language. It can be used to read arbitrary sequences of bytes similar to the int8 family of keywords. The major difference here is you can use it to read arbitrary sequences into strings and compare them.

The implementation is done by dynamically allocating memory for each bytes operation and tracking all the allocations during bytecode execution in a notebook and destroying that notebook when execution is done.

An obvious optimization to make here is to track allocation offset and lengths so that we don't have lots of unnecessary allocations done in loops, like this:

for any s in ("foo", "bar"): (bytes(0, 3) == s)

While the above is an obviously contrived example it does make sense to avoid unnecessary allocations if we already have them.

Discussed in https://github.com/VirusTotal/yara/issues/1780 by @metthal.


wxsBSD:
Some obvious follow-ups to this that would improve it's usefulness is having the ability to convert a SIZED_STRING to it's hexlified equivalent and vice-versa. I think this belongs in the `string` module proposed in https://github.com/VirusTotal/yara/pull/1779, and I'm happy to add it if this is deemed a good idea. Also, I'll add docs before merging, just let me know what you think!

One way this is useful is for the times a hash of a blob is embedded in the file and you want to be able to check it. Here is a silly example of how you could check that:

```
wxs@mbp yara % cat ~/x
I AM A STRING|fd09d734380f7d6fc88655e19d6a5097ee0d1f2beefeaa59358c050bc4d586eewxs@mbp yara % xxd ~/x
00000000: 4920 414d 2041 2053 5452 494e 477c 6664  I AM A STRING|fd
00000010: 3039 6437 3334 3338 3066 3764 3666 6338  09d734380f7d6fc8
00000020: 3836 3535 6531 3964 3661 3530 3937 6565  8655e19d6a5097ee
00000030: 3064 3166 3262 6565 6665 6161 3539 3335  0d1f2beefeaa5935
00000040: 3863 3035 3062 6334 6435 3836 6565       8c050bc4d586ee
wxs@mbp yara % cat rules/test.yara
import "hash"

rule a {
  condition:
    hash.sha256(0, 13) == bytes(filesize - 64, 64)
}
wxs@mbp yara % ./yara rules/test.yara ~/x
a /Users/wxs/x
wxs@mbp yara %
```

chemberger:
Hello !

I believe that if the ensure_defined macro fires in OP_READ_BYTES, the code will break out of the switch and loop again, and the yara VM stack will be misaligned.

The operation looks like it consumes the top two values and puts back the result (undefined if error), however if ensure_defined breaks, no value is put back onto the stack.

Also, it would be nifty if it would continue reading from the next block if the range happens to coincide with a block boundary.

wxsBSD:
Good point! I forgot about that. I've got some planned updates for this and I will include your suggestion too. Thank you for pointing it out!

wxsBSD:
OK, so other than not working across block boundaries I think I'm mostly happy with this implementation now. I think the limits are very conservative but they can be easily tweaked. In fact, I think it probably makes sense to move them to runtime configuration options available via the API.

I'll dig into the cross block boundaries soon but I think the implementation (and even the idea) is ready for a wider discussion. If it is starting to look like it might be merged I'll do the work to update docs and expose it as some configuration options.

wxsBSD:
I'm going to close this out as it isn't completely ready (it doesn't work cross blocks) and I want to focus all new efforts on yara-x in the upcoming year.

PR #1799, opened 2022-10-03
Add signatures details in PE module when using Wincrypt

vthib:
The PE signatures fields depends on OpenSSL, which means they are completely missing when using any other crypto API. This PR aims to introduce some support of those fields when using the Windows API (wincrypt).

This support is only done for the 4.2 fields, before the addition of the authenticode dependency. This allows basic support which covers all the existing rules, and support for the newly added fields can be added in the future.

The code is fairly straightforward and not too complex. The only major issues are that three fields are dependent on OpenSSL specifics, which are:

- `algorithm`, which is a human-readable string for the algorithm_oid.
- `issuer` and `subject`, which depends on how OpenSSL converts a X509 name into a human readable string. `X509_NAME_oneline` is used, which is documented as being non standard. The wincrypt API generates another format, which can be seen in the documentation update included in this PR.

For `algorithm`, I have handled a special case for "sha1WithRSAEncryption" and "sha256WithRSAEncryption" so that those names are the same. For the rest, i'm using the wincrypt name. I don't think this is a big issue, as I'm not sure anything other than the previous two algorithms can really happen in authenticode, and the field is already documented as being subject to change, and `algorithm_oid` should be used instead.

For `issuer` and `subject`, in all of the rules that i have found, the matching is done on the content of one component and are not subject to changes on the separation of those components, so afaict the differences should be OK as well.

I have tried to match the existing coding style, If I made any mistakes please tell me and I will correct them.

Let me know if such a feature addition sounds OK for you, it would be a great help for us as we try to avoid depending on OpenSSL.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,6 +30,12 @@ if UB_SANITIZER
 AM_CFLAGS+=-fsanitize=undefined -fno-sanitize-recover=undefined
 endif
 
+if LINK_CRYPT32
+  USE_CRYPT32=-lcrypt32
+else
+  USE_CRYPT32=
+endif
+
 # Build the library in the hand subdirectory first.
 SUBDIRS = libyara
 DIST_SUBDIRS = libyara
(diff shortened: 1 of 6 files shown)


plusvic:
Sorry for not looking into this earlier. I just reviewed this PR and the difference in outputs between OpenSSL and Wincrypt bothers me. Specifically this...

```
        A string containing information about the issuer. The formatting of
        those strings depends on the library used, OpenSSL or Wincrypt.
        These are some examples with OpenSSL::

            "/C=US/ST=Washington/L=Redmond/O=Microsoft Corporation/CN=Microsoft Code Signing PCA"

            "/C=US/O=VeriSign, Inc./OU=VeriSign Trust Network/OU=Terms of use at https://www.verisign.com/rpa (c)10/CN=VeriSign Class 3 Code Signing 2010 CA"

            "/C=GB/ST=Greater Manchester/L=Salford/O=COMODO CA Limited/CN=COMODO Code Signing CA 2"

        And the same examples with Wincrypt::

            "C=US, S=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Code Signing PCA"

            "C=US, O=\"VeriSign, Inc.\", OU=VeriSign Trust Network, OU=Terms of use at https://www.verisign.com/rpa (c)10, CN=VeriSign Class 3 Code Signing 2010 CA"

            "C=GB, S=Greater Manchester, L=Salford, O=COMODO CA Limited, CN=COMODO Code Signing CA 2"
```

This may cause issues with certain rules that won't match properly depending on whether YARA was built with OpenSSL or Wincrypt support. This harms cross-platform compatibility of YARA rules, and that's something that we should avoid. 

We already have issues like this, like for example with some modules like `magic` which is not available on Windows. But in this case the incompatibility  is even more subtle and harder to be aware of.

vthib:
Thanks, yes I agree that this is an issue. I'll try to see how feasible it is to get the same formatting as OpenSSL.

vthib:
I have added a commit that generates the same format as the OpenSSL one. The gives the same string in almost all cases. I could only find differences in a single file using more exotic RDNs, where I have this:

With OpenSSL:

```
signatures
		[0]
			issuer = "/C=BE/O=GlobalSign nv-sa/CN=GlobalSign CodeSigning CA - G3"
			subject = "/C=US/ST=Kansas/L=Olathe/O=Power Admin LLC/OU=Development/CN=Power Admin LLC/emailAddress=support@poweradmin.com"
		[1]
			issuer = "/C=US/O=Symantec Corporation/OU=Symantec Trust Network/CN=Symantec Class 3 Extended Validation Code Signing CA - G2"
			subject = "/jurisdictionC=US/jurisdictionST=Kansas/jurisdictionL=Olathe/businessCategory=Private Organization/serialNumber=3594967/C=US/ST=Kansas/L=Olathe/O=Power Admin LLC/CN=Power Admin LLC"
```

With wincrypt:

```
signatures
		[0]
			issuer = "/C=BE/O=GlobalSign nv-sa/CN=GlobalSign CodeSigning CA - G3"
			subject = "/C=US/ST=Kansas/L=Olathe/O=Power Admin LLC/OU=Development/CN=Power Admin LLC/E=support@poweradmin.com"
		[1]
			issuer = "/C=US/O=Symantec Corporation/OU=Symantec Trust Network/CN=Symantec Class 3 Extended Validation Code Signing CA - G2"
			subject = "/OID.1.3.6.1.4.1.311.60.2.1.3=US/OID.1.3.6.1.4.1.311.60.2.1.2=Kansas/OID.1.3.6.1.4.1.311.60.2.1.1=Olathe/OID.2.5.4.15=Private Organization/SERIALNUMBER=3594967/C=US/ST=Kansas/L=Olathe/O=Power Admin LLC/CN=Power Admin LLC"
```

There is a `E=` instead of a `emailAddress=` in the first signature, and the OIDs in the second one.

Those are for RDNs that are basically not included in any spec (neither RFC 1779 or RFC 2253), and impls are allowed to do whatever they want. This is quite hard to handle, except for making a huge hashmap from OIDs to the name used in OpenSSL, which is feasible, but i'm not sure it is worth it.

In addition, the behavior of the X509_NAME_oneline function in OpenSSL changes between the 1.1 version and the 3.0 one, as the 1.1 one was quite buggy. I went with using the 3.0 version, but afaict the authenticode lib does the opposite, so I can change that as well.

vthib:
I have done a few changes to improve compatibility:

- I now follow the OpenSSL 1.1 format instead of the 3.0 one. This is similar to what the authenticode parser does (ensure the 1.1 format is used).
- I added code to easily allow translating a RDN name from wincrypt to openssl, and I used it to convert the names I mention earlier (the E/emailAddress, the jurisdiction*, etc). I think this code can be easily modified if more translations are needed.

Please let me know how this looks to you. I tried to do all I could to make the differences as limited as possible, and I completely understand being wary of adding code that can provide differences in rule evaluation. However, that feature would be very useful for users of the wincrypt backend (much more useful that not having pe.signatures at all). Maybe the documentation can be updated to warn users of the wincrypt backend of possible minor differences in the RDN names in the "issuer" and "subject" field?

vthib:
@plusvic Any opinion on those changes, does it looks good to you?

plusvic:
Sorry for the delayed response, I've been very busy lately. I've reviewed the latest changes and certainly the outputs are more similar now. However I prefer putting this change on hold for the time being. I'm currently exploring the idea of porting YARA to Rust, and I'm being very conservative with respect to the new code merged into the repo, specially if those changes make things harder in the case that the Rust implementation actually takes off.

All the OpenSSL stuff is going to be particularly tricky to reimplement, so until I have a clearer view of the path head I prefer being cautious.

vthib:
I see, thanks for the answer, I understand.

As a sidenote, I am as well working on a Rust implementation which is quite far down the line 😄 I intend to publish it soon

PR #1801, opened 2022-10-12
Print module names.

### LINKED ISSUE #737: Command to list the modules installed
In order to see the modules compiled with yara, we can include a command like:
```
$yara -xl
pe
elf
cuckoo
math
[...]
```

wxsBSD:
Add a -M option to the cli which will print the module names to stdout. As the number of modules has grown it can become confusing to know what modules are available, so this option will display them on stdout.

NOTE: I'm not sure I like how I implemented this. It seems hackish to expose an API that just prints things on stdout but it felt even more hackish to walk the yr_modules_table array from outside of the module code. ;)

Fixes #737.

### DIFF
--- a/libyara/modules.c
+++ b/libyara/modules.c
@@ -56,17 +56,16 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 YR_MODULE yr_modules_table[] = {
 #include <modules/module_list>
-};
+    {NULL, NULL, NULL, NULL, NULL, NULL}};
 
 #undef MODULE
 
 int yr_modules_initialize()
 {
-  int i;
-
-  for (i = 0; i < sizeof(yr_modules_table) / sizeof(YR_MODULE); i++)
+  for (YR_MODULE* module = yr_modules_table; module->initialize != NULL;
+       module++)
   {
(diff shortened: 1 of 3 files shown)


libyara/modules.c
```diff
@@ -191,3 +191,9 @@ int yr_modules_unload_all(YR_SCAN_CONTEXT* context)
 
   return ERROR_SUCCESS;
 }
+
+void yr_modules_print_names(void)
```
plusvic, line 195: I think we should make this a proper API that returns the list of module names without printing it. You may want to get the list of modules for some other reason besides printing it to the console.

wxsBSD: Done, but I'm not sure you will like the name I went with: `yr_modules_get_table()`. I'm happy to rename it to something better if you have suggestions, maybe `yr_modules_get_modules()`?

Also, I'll update docs for this before merging.

Lastly, any thoughts on adding something like this to yara-python? I think it would be useful to be able to get that information from there too. I'll add it there if you agree.

plusvic (approved):

libyara/modules.c
plusvic: Yes, I think that adding it to yara-python may be useful too.

PR #1806, opened 2022-10-19
Add warnings for edge cases.

wxsBSD:
When using "all of them at 0" it is now a warning if you have more than 1 string defined in "them".

"N of them" (where N > 1) is also a warning.

### DIFF
--- a/libyara/grammar.y
+++ b/libyara/grammar.y
@@ -1788,7 +1788,8 @@ expression
         //
         // This means you can do "all of them at 0" if you only have one string
         // defined in the set.
-        if (($1.type == EXPRESSION_TYPE_INTEGER && $1.value.integer > 1) ||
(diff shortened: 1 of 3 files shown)


libyara/grammar.y
```diff
@@ -1780,6 +1780,22 @@ expression
...
 
+        // Both of these are warnings:
+        //
+        // "N of them at 0" where N > 1
+        //
+        //"all of them at 0" where there is more than 1 in "them".
+        //
+        // This means you can do "all of them at 0" if you only have one string
+        // defined in the set.
+        if (($1.type == EXPRESSION_TYPE_INTEGER && $1.value.integer > 1) ||
```
plusvic, line 1791: We should check for `!IS_UNDEFINED($1.value.integer)`

plusvic (approved):

PR #1811, opened 2022-10-22
Merge Makefile.am into a single file

hillu:
Benefits
- Changes that result in a rebuild of libyara will also cause test programs to be rebuilt
- Avoid repeating flag definitions


plusvic:
CIFuzz is failing, it may need to change this path: https://github.com/Oss-fuzz/oss-fuzz/blob/master/projects/yara/build.sh#L30

plusvic:
I tried merging this PR but I'm getting errors like this when running tests:

```
dyld[36973]: symbol not found in flat namespace (_string__finalize)
```

It looks like it's related to linking the tests against the dynamic library instead of the static library.

hillu:
> I tried merging this PR but I'm getting errors like this when running tests:

I suppose we would have seen that right away if we'd had MacOSX CI tests. :-)

hillu:
@plusvic I have been able to reproduce the problem using CI. Investigating …

hillu:
Fixed. The root cause of the problem was the new `string` module.

I had wondered where those `string__something` symbols might come from but otherwise ignored them because of the unfamiliar (to me) messages from the MacOSX dynamic linker. And perhaps because of your assumption about static linking, @plusvic. :-)

Feeling like a total newbie at this "compiling stuff from source" thing right now. ;-)

PR #1831, opened 2022-12-08
Add pe.import_rva() functions.

wxsBSD:
Add pe.import_rva("foo.dll", "func1") which returns the RVA of the imported function. Also add pe.import_rva("foo.dll", 1) which does the same but the import is done by ordinal.


wxsBSD:
If this makes sense I can do the same for `delayed_import_details` and update docs before merging.

wxsBSD:
Merged in the `next` branch.

PR #1842, opened 2022-12-21
fix error handling in string.to_int()

vthib:
Error handling in the to_int method of the new string module was not properly done.

- errno was not reset prior to the call to strtoll, so YR_UNDEFINED could be returned if a prior call failed, but not the current one.
- overflow/underflow was not properly checked, as it returns LLONG_MIN or LLONG_MAX and not 0.

Instead, do what is recommended in the man page for strtol:
- set errno to 0 prior to the call
- check errno after the call

### DIFF
--- a/docs/modules/string.rst
+++ b/docs/modules/string.rst
@@ -28,7 +28,7 @@ strings section of your rule.
     .. versionadded:: 4.3.0
 
     Convert the given string, interpreted with the given base, to a signed
-    integer. Base must be 0 or between 2 and 32 inclusive. If it is zero then
+    integer. Base must be 0 or between 2 and 36 inclusive. If it is zero then
     the string will be intrepreted as base 16 if it starts with "0x" or as base
     8 if it starts with "0". Leading '+' or '-' is also supported.
 
--- a/libyara/modules/string/string.c
+++ b/libyara/modules/string/string.c
@@ -77,6 +77,9 @@ define_function(to_int_base)
   int64_t base = integer_argument(2);
   int64_t result = 0;
 
+  if (!(base == 0 || (base >= 2 && base <= 36))) {
+      return_integer(YR_UNDEFINED);
+  }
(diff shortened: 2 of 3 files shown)


vthib:
Test fail on macos on the `string.to_int("") == 0` test I also added. I guess this is because this is unspecified behavior in strtol, see man page of strtol:

> The implementation may also set to EINVAL in case no conversion was performed (no digits seen, and 0 returned).

I can remove this test to make this MR go through, but there is definitely an issue here, i'm not sure what the desired behavior is. IMHO not returning an error when not parsing a single digit is unexpected, but I don't see how it could be detected if strtol does not set errno in those cases in some libc implementations. Discussion about this could probably be moved in another issue however

vthib:
I have opened https://github.com/VirusTotal/yara/issues/1843 to discuss the `to_int("")` case

vthib:
I added a fix for #1843 as well. This way, this PR properly fixes and specifies all error/corner cases for string.to_int

wxsBSD:
Thanks for taking care of this! FWIW, I think these changes are good. :)

plusvic (approved):

plusvic:
This test case is failing when compiling with Cygwin on Windows:

```
  assert_true_rule(
      "import \"string\" \
      rule test { \
        condition: \
          not defined string.to_int(\"1\", -1) and \
          not defined string.to_int(\"1\", 1) and \
          not defined string.to_int(\"1\", 37) \
      }",
      NULL);
```

The implementation of `strtoll` is returning with `errno = 0` with incorrect base values like -1, 1 and 37. So, it looks like `strtoll` is not behaving in the same ways in all its implementation. We better check for valid base values in `string_to_int`

vthib:
Thanks you for debugging it, I didn't see that there was an issue with Cygwin. I have added a commit to validate the base before calling strtol, hopefully this should fix it

plusvic (approved):

PR #1847, opened 2022-12-31
fix telfhash computation on x64 specific functions

vthib:
The telfhash algorithm excludes x64 functions, see here: https://github.com/trendmicro/telfhash/blob/d2a82f85a5b/telfhash/telfhash.py#L59

Looks like there was a copy paste error when porting this algorithm in YARA, as the test for "is_bad_prefix" is copied, but this is not the right test of "is_x86_64".

This PR properly fixes it.

CC @HoundThe


libyara/modules/elf/elf.c
```diff
@@ -102,7 +102,7 @@ define_function(telfhash)
         string functions (str.* and mem.*), gcc changes them depending on arch
         symbols starting with . or _ */
     bool is_bad_prefix = name[0] == '.' || name[0] == '_';
-    bool is_x86_64 = name[0] == '.' || name[0] == '_';
+    bool is_x86_64 = strncmp(name + strlen(name) - 2, "64", 2) == 0;
```
HoundThe, line 105: I don't think a symbol name must be non-empty, so `strlen(name)` can return 0 and read into `name - 2` - pre-running the buffer.

HoundThe (changes requested):
Thanks for noticing the copy-paste error and fixing it!

libyara/modules/elf/elf.c
vthib: I think a non-empty symbol is impossible given we filtered symbols that are global and functions previously, however a one letter symbol is definitely possible, that's a mistake on my part, thanks.

I added a test using custom made elf file with a bit of everything to check the telfhash is correct. I added a one letter symbol and valgrind does indicate out of bound read, as expected.

HoundThe (approved):
Thanks! LGTM

plusvic (approved):

Issue #1838, opened 2022-12-16
Yara doesn't print out anything when we try to scan `/proc/{pid}/maps`


leomao10:
**Describe the bug**
We try to use yara to scan `/proc/{pid}/maps` and found that nothing get print out even the rules maps strings in the files.

The rules we use is like this:
```
rule node_package {
    meta:
        name = "babelcli"

    strings:
        $name = "babelcli",
        $text = "node_modules"
    condition:
        all of them
}
```

yara command is like `yara ./rule.yara /proc/{id}/maps`, and nothing get print out. However, if I copy `maps` file to other location like `cp /proc/{id}/maps /{id}_maps`, and use similar command to scan the new file: `yara ./rule.yara /{id}_maps`, then there is output from yara with rule matched.

**To Reproduce**
- `docker container run -it --entrypoint /bin/bash ubuntu`
- run command `apt-get update; apt-get install yara procps` in container
- Create sleep_rule.yara in container and add following rule:
```
rule LOG_sleep {
    meta:
        name = "sleep process map"

    strings:
        $text1 = "ld-linux-aarch64.so.1" nocase
        $text2 = "aarch64-linux-gnu" nocase
    condition:
        all of them
}
```
- run `sleep 1000 &` in container and get its pid
- run `yara /sleep_rule.yara /proc/{id}/maps` in container and there is nothing prints out
- `cp /proc/{id}/maps /{id}_maps` and then run `yara /sleep_rule.yara /{id}_maps` and we see `LOG_sleep /502_maps` message print out.

**Expected behavior**
When using `yara /sleep_rule.yara /proc/{id}/maps`, it should print out the rule if yara rule match.

**Please complete the following information:**
 - OS: ubuntu
 - YARA version: 3.9.0, 4.2.1

hillu:
YARA wants to memory-map the files it scans.
I was going to write that this is not possible at all with files in `/proc`, but the root cause for the silent failure is that YARA uses the file size as reported by _fstat(3)_ to determine the size it uses for the memory map – and the file size for any file in `/proc/<pid>/` is 0.

In my opinion,YARA should at least recognize files where using the usual fstat+mmap approach does not make sense and refuse to scan them.

plusvic:
Closed in #1848

PR #1848, opened 2022-12-31
Refuse to map procfs files on Linux

hillu:
It makes no sense to try to mmap files for which the filesystem lies about the size (0).

Clsoe #1838 

(It may make sense to consider other pseudo filesystems on Linux and other Unices as well.)

### DIFF
--- a/libyara/filemap.c
+++ b/libyara/filemap.c
@@ -157,6 +157,11 @@ YR_API int yr_filemap_map_fd(
 
 #else  // POSIX
 
+#ifdef __linux__
+#include <sys/vfs.h>
+#include <linux/magic.h>
+#endif
+
 #define MAP_EXTRA_FLAGS 0
 
 #if defined(__APPLE__)
@@ -184,6 +189,9 @@ YR_API int yr_filemap_map_fd(
     YR_MAPPED_FILE* pmapped_file)
 {
   struct stat st;
+#ifdef __linux__
+  struct statfs stfs;
+#endif
 
   pmapped_file->file = file;
   pmapped_file->data = NULL;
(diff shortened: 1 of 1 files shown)


plusvic (approved):

gwillem:
This adds a dependency to libmagick, linux/magic.h which complicates cross compilation. @hillu how did you fix this for your own projects?

hillu:
> A dependency to libmagick, linux/magic.h

No, `linux/magic.h` has nothing to do with libmagick; on Debian and derivative distributions, it is shipped as part of `linux-libc-dev`.

"magic" in this case refers to filesystem magic numbers reported by `fstatfs`, we only need `PROC_SUPER_MAGIC` which is part of the Kernel API. So we could get away with this patch:
``` diff
--- a/libyara/filemap.c
+++ b/libyara/filemap.c
@@ -159,7 +159,7 @@ YR_API int yr_filemap_map_fd(
 
 #ifdef __linux__
 #include <sys/vfs.h>
-#include <linux/magic.h>
+#define PROC_SUPER_MAGIC       0x9fa0
 #endif
 
 #define MAP_EXTRA_FLAGS 0
```

gwillem:
Thanks a lot Hillu! Somehow the Go linker keeps adding -lmagic to my build which I assumed was related to the above but I should continue my search. Anyhow, thanks for the patch!

gwillem:
Thanks @plusvic. To clarify, there were two issues:

1. Dep on linux.h, solved by Hillu's patch above. 

2. Unknown origin of -lmagic injection by the Go linker. This was introduced by our upgrade of yara 4.0.2 to 4.3.2. However, we just upgraded to 4.5.1 and our build succeeds. Apologies for reporting an issue with an older version 😅 

(We use https://github.com/hillu/go-yara v4.3, we incorrectly assumed we had to use a corresponding yara version)

PR #1907, opened 2023-04-07
Index of Coincidence

BitsOfBinary:
The index of coincidence "provides a measure of how likely it is to draw two matching letters by randomly selecting two letters from a given text" (reference: https://en.wikipedia.org/wiki/Index_of_coincidence).

As we know this index will be around `0.067` for English text, this may be a useful indicator for plaintext strings. An example usage can be seen here: https://gchq.github.io/CyberChef/#recipe=Index_of_Coincidence()&input=VGhpcyBhIHRlc3QgZm9yIHRoZSBpbmRpY2F0b3Igb2YgY29pbmNpZGVuY2U

I've added this function for both data offsets/string values to the `math` module. I've also added documentation and tests for it.

### DIFF
--- a/docs/modules/math.rst
+++ b/docs/modules/math.rst
@@ -184,3 +184,19 @@ file and create signatures based on those results.
 
     *Example: math.to_string(32, 16) == "20"*
     *Example: math.to_string(-1, 16) == "ffffffffffffffff"*
+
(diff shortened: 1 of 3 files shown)


BitsOfBinary:
Updated for compatibility with YARA v4.3.2; tests passing as expected.

PR #1919, opened 2023-05-07
Warn on always false integer reads and comparisons.

### LINKED ISSUE #1918: Check for impossible conditions when using uint() operators.
**Is your feature request related to a problem? Please describe.**
Sometimes when writing rules I'll use the wrong value of x for uintx type rules meaning the condition will never match. For example

```yara
condition:
    uint32(0) == 0xABCD
````

**Describe the solution you'd like**
Troubleshooting a condition like this is relatively easy once you figure out what is going on but it can be easy to gloss over. It would be useful if when compiling rules the compiler checked that the right handside contained a number of bytes that could plausibly match the left hand side,

**Describe alternatives you've considered**
I could just write better rules I guess :(

wxsBSD:
When reading an integer of a specific size and comparing it to an integer of a larger size where any of the upper bits are set we will now emit a warning because the comparison is always false.

These will always evaluate to false because the "extra" bytes are non-zero:

uint8(0) == 0x1100
uint16(0) == 0x110000
uint32(0) == 0x1100000000

While I'm here, move a test into a better place for it. I added it in the wrong place in ccbc405.

Fixes #1918.

### DIFF
--- a/libyara/grammar.c
+++ b/libyara/grammar.c
@@ -952,13 +952,13 @@ static const yytype_int16 yyrline[] =
     1163,  1167,  1202,  1255,  1297,  1320,  1326,  1332,  1344,  1354,
     1364,  1374,  1384,  1394,  1404,  1414,  1428,  1443,  1454,  1529,
     1567,  1471,  1695,  1706,  1717,  1736,  1755,  1767,  1804,  1810,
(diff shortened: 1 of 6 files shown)


plusvic:
YARA is currently in maintenance mode, but this is already implemented in YARA-X.

PR #1927, opened 2023-06-13
feat: skip bytecode evaluation for some rules without string matches

secDre4mer:
Optimize a common case where YARA conditions are formed like e.g. "... and 1 of them and ...", in other words, requiring a string match to ever be true.
By noting these cases and recording in a bitmap if a string match occurred, the condition evaluation for these rules can be skipped entirely in most cases.

### DIFF
--- a/libyara/scanner.c
+++ b/libyara/scanner.c
@@ -508,16 +508,10 @@ YR_API int yr_scanner_scan_mem_blocks(
     if (result != ERROR_SUCCESS)
       goto _exit;
 
-    for (i = 0; i < scanner->rules->num_rules; i++)
-    {
-      // If a rule has no required_strings, this means that the condition might
-      // evaluate to true without any matching strings, and we therefore have to
-      // mark it as "to be evaluated" from the beginning.
-      if (scanner->rules->rules_table[i].required_strings == 0)
-      {
-        yr_bitmask_set(scanner->rule_evaluate_condition_flags, i);
-      }
-    }
+    memcpy(
+      scanner->rule_evaluate_condition_flags,
+      scanner->rules->rule_evaluate_condition_flags,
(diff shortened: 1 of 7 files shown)


secDre4mer:
In case it's relevant, some background information that led to this PR:
We use ~20000 YARA rules in a ruleset and noticed that in such a large ruleset, condition evaluation takes up a serious amount of CPU time (~50-70% of total YARA scan time). 
This PR tries to reduce the overhead created by conditions (in our case, ~90% of all rules fall into the "only evaluate if a string matches" category introduced by the PR). Initial timings look good, with condition evaluation time dropping by ~65% in our case.

Some additional optimizations (like: tracking which strings actually have to match for the condition to possibly be true, tracking the number of string matches, ...) could be implemented if considered worthwhile.

libyara/scanner.c
```diff
@@ -498,6 +508,17 @@ YR_API int yr_scanner_scan_mem_blocks(
     if (result != ERROR_SUCCESS)
       goto _exit;
 
+    for (i = 0; i < scanner->rules->num_rules; i++)
```
plusvic, line 511: My only concern is this loop. The loop gets executed with every scanned file, and it may take a while when you have hundreds of thousands of rules as we have in VirusTotal. I think this could be donde more efficiently if we store the original bitmask in `scanner->rules`, and with every scanned file `rule_evaluate_condition_flags` is initialized by making a copy of the original bitmask using `memcpy`. This should be much more efficient as memcpy will copy multiple bits per CPU instruction, instead of having multiple instructions per bit, like in this case.

secDre4mer: Done

plusvic (approved):

gabriellandau:
Thanks for adding this.  Was this benchmarked?  Are there any numbers?

PR #1936, opened 2023-07-24
fix: If any signature is valid, the file is properly signed

aliyunav:
If any signature will be valid, pe file is correctly signed


plusvic:
Do you have some test case for this? Some file that mixes invalid and valid signatures?

aliyunav:
No, there is no such situation.The last signature taken by the old code is not as expected. It also does not match the current comment of the variable ”signature_valid“

plusvic:
It would be great if we can add a test case for this, that's why I was asking for file.

plusvic (approved):

PR #1944, opened 2023-08-10
Fix imphash issue with empty import names.

### LINKED ISSUE #1943: yara imphash and pefile imphash are different
**Describe the bug**
yara imphash and pefile imphash are different
So you can't use imphash in yara rule.
Any reason?
sha256: 803d740df04cc995fd7010473be6b06e190c7ca5114a525d951ae11e7394ab12
yara imphash: ecb1f45f310c2bf07612a019ce157b0f
pefile imphash: b22fdcce4de8993adbff1ff244d1b4a5
https://www.virustotal.com/gui/file/803d740df04cc995fd7010473be6b06e190c7ca5114a525d951ae11e7394ab12/details

**To Reproduce**

**Expected behavior**

**Screenshots**

**Please complete the following information:**
 - OS: Windows
 - YARA version: 3.8.0

**Additional context**

wxsBSD:
If an import has an empty name skip processing it. This is consistent with the behavior of pefile (https://github.com/erocarrera/pefile/blob/593d094e35198dad92aaf040bef17eb800c8a373/pefile.py#L5871-L5872).

Add a test case which is just the tiny test file with the first import name set to all NULL bytes. I tested that pefile calculates the imphash of it, which matches what YARA now calculates too:

>>> import pefile
>>> pe = pefile.PE('/Users/wxs/src/yara/tests/data/tiny_empty_import_name')
>>> pe.get_imphash()
'0eff3a0eb037af8c1ef0bada984d6af5'
>>>

Fixes #1943


plusvic:
The file `tests/data/tiny_empty_import_name` wasn't added in the PR.  Interesting enough, the tests are failing with a memory leak that doesn't seem related to the change done here.

wxsBSD:
Oops, I forgot to add it! Will do so as soon as I'm able.

wxsBSD:
Please don't merge this yet. There are other problems raised in the linked issue that are being looked into.

wxsBSD:
> The file `tests/data/tiny_empty_import_name` wasn't added in the PR. Interesting enough, the tests are failing with a memory leak that doesn't seem related to the change done here.

Actually, the memory leak was because of my changes. I was not calling `yr_free()` if the import had an invalid name.

I also discovered that if we failed to allocate a new `IMPORT_FUNCTION` we would free the name (correctly) but then continue without incrementing the thunk pointer or function index, which would cause an infinite loop. I've fixed that too.

plusvic (approved):

PR #1958, opened 2023-08-24
Add docs for unreferenced strings.

wxsBSD:
Add docs for the unreferenced string feature in 4.4.0

### DIFF
--- a/docs/writingrules.rst
+++ b/docs/writingrules.rst
@@ -747,6 +747,17 @@ condition. Any other string must be referenced in the condition. This is useful
 if you want to search for particular strings and handle them in a custom
 callback but don't really need them for your condition logic.
 
+.. code-block:: yara
+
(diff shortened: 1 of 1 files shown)


docs/writingrules.rst
```diff
@@ -738,6 +738,15 @@ the C API.
...
 
+Unreferenced strings
+--------------------
+
+YARA 4.4.0 allows for unreferenced strings in the condition. If a string
+identifier starts with an ``_`` then it does not have to be referenced in the
+condition. Any other string must be referenced in the condition. This is useful
+if you want to search for particular strings and handle them in a custom
+callback but don't really need them for your condition logic.
+
```
plusvic, line 749: An example would be nice, is not clear if the underscore must be before or after the `$`.

plusvic (approved):

PR #1988, opened 2023-10-12
fix verified field in pe module with multiple signatures

vthib:
Commit 1a85738 fixed one bug but introduced a new one. The `pe.is_signed` field is now properly set (true if any signature is valid), but the `signatures[i].verified` is now incorrectly set: instead of being set to whether this signature is verified, it is set to true if any previous signature is verified.

Unfortunately, there is no test file that is triggering this behavior. This would require having a signature that is invalid appearing after a signature that is valid.

I targetted master, but we probably want this fix in 4.4.0. I'm not sure how you handle this, do you want me to target 4.4.x?

### DIFF
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -1750,19 +1750,18 @@ void _process_authenticode(
   if (!auth_array || !auth_array->count)
     return;
 
-  /* If any signature will be valid -> file is correctly signed */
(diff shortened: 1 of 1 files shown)


vthib:
@plusvic Do you have time to take a look at this as well as my other PRs? I'd love to be able to get those fixes in the next 4.4 release

plusvic (approved):

PR #1991, opened 2023-10-17
feat: memory module

secDre4mer:
A small YARA module (inspired by https://twitter.com/NinjaParanoid/status/1712743509961380325) that allows rules to query memory protection for live process memory. This allows writing conditions like `for any i in (1..#a) : ( memory.Protection(@a[i]) & memory.EXECUTE == memory.EXECUTE)` for strings that should only match on executable memory.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -80,6 +80,7 @@ yara_library(
         "time",
         "console",
         "string",
+        "memory",
     ],
     modules_srcs = [
         "libyara/modules/cuckoo/cuckoo.c",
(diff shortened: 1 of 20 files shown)


klabksec:
in addition to mem RWX permissions and region size and all, it would be useful to describe/access characteristics of the current process - dll(s) loaded and their associated file path, any particular port listening or named pipes, number of threads, etc

plusvic:
I see the utility of this new module, but I'm worried about its maintenance and testability. One of the problem with process memory scanning in general is that it's poorly tested, and this adds up to the already existing problem. Testing process scanning is hard, I never put too much effort on that, and I regret it. But at this point I don't want to make this issue bigger than it is.

Also, I prefer not adding new modules to YARA in general, given that YARA-X is probably the future.

PR #1992, opened 2023-10-23
Fix compiler crash.

wxsBSD:
The compiler hits some asserts when compiling this rule:

```
rule r{condition:for 5f in(""):(f
```

It looks like the value of the expression here is undefined, so when trying to use the buffer_id and offset fields of the sized_string_ref we end up using the upper and lower halves of YR_UNDEFINED as if they are real values.

I'm not sure I like checking for YR_UNDEFINED using the integer field of the union, so I'm open to better ideas here.

Found by: Google clusterfuzz

### DIFF
--- a/libyara/grammar.y
+++ b/libyara/grammar.y
@@ -1318,6 +1318,10 @@ regexp
 boolean_expression
     : expression
       {
+        // If the value of an expression is ever undefined we have a problem.
+        if (IS_UNDEFINED($1.value.integer))
+          fail_with_error(ERROR_INTERNAL_FATAL_ERROR);
+
         if ($1.type == EXPRESSION_TYPE_STRING)
         {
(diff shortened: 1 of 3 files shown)


libyara/grammar.y
```diff
@@ -1318,6 +1318,10 @@ regexp
 boolean_expression
     : expression
       {
+        // If the value of an expression is ever undefined we have a problem.
```
plusvic, line 1321: I would say that an expression can be undefined, in fact, most expressions will be undefined because their value can't be known at compile time. I'm not understanding the reasoning behind this fix.

wxsBSD: It fixes a compiler crash. I don't think the fix is technically correct but I did want to bring it up as a fuzzer found it and it got auto assigned to me. ;)

PR #1996, opened 2023-10-29
stop fetching memory blocks when an elf is found

vthib:
When scanning multiple memory blocks, we should stop as soon as a valid elf is found. This was broken as the "break" applies to the switch, and not the outer loop. Fix this by reworking the code to remove the switch.

This bug had two repercutions:

- instead of parsing the first memory block that contains an elf, it returned the last one, or actually even worse, a mix of all elfs depending on which values are set again or not. This however should not manifest in normal situations, notably during process scanning, as I suppose only one memory region would contain an elf executable.

- the module caused the fetch of all memory blocks instead of stopping as soon as one was found, which is worse performance wise.

### DIFF
--- a/libyara/modules/elf/elf.c
+++ b/libyara/modules/elf/elf.c
@@ -1123,91 +1123,79 @@ int module_load(
       return ERROR_INSUFFICIENT_MEMORY;
 
     module_object->data = elf;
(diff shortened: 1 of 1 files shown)


vthib:
@plusvic anything blocking this MR that I need to modify?

plusvic (approved):

Issue #2007, opened 2023-11-23
ECHO macro causes CLI tool to crash at runtime (Windows)


1ndahous3:
The bug was discovered when the CLI tool crashed while loading a simple invalid rule:
```
rule test{
	strings:
		$s = /\
}
```

Actually MSVC CRT [parameter validation](https://learn.microsoft.com/en-us/cpp/c-runtime-library/parameter-validation) failed at runtime:
```
>	yara64.exe!_write_nolock(int fh, const void * buffer, unsigned int buffer_size, __crt_cached_ptd_host & ptd)
 	yara64.exe!_write_internal(int fh, const void * buffer, unsigned int size, __crt_cached_ptd_host & ptd)
 	yara64.exe!__acrt_stdio_flush_nolock(_iobuf * public_stream, __crt_cached_ptd_host & ptd)
 	yara64.exe!__acrt_stdio_end_temporary_buffering_nolock(bool flag, _iobuf * public_stream, __crt_cached_ptd_host & ptd)
 	yara64.exe!__acrt_stdio_temporary_buffering_guard::~__acrt_stdio_temporary_buffering_guard()
 	yara64.exe!_fwrite_internal::__l2::<lambda>()
 	yara64.exe!__crt_seh_guarded_call<unsigned __int64>::operator()<void <lambda>(void),unsigned __int64 <lambda>(void) &,void <lambda>(void)>(__acrt_lock_stream_and_call::__l2::void <lambda>(void) && setup, _fwrite_internal::__l2::unsigned __int64 <lambda>(void) & action, __acrt_lock_stream_and_call::__l2::void <lambda>(void) && cleanup)
 	yara64.exe!__acrt_lock_stream_and_call<unsigned __int64 <lambda>(void)>(_iobuf * const stream, _fwrite_internal::__l2::unsigned __int64 <lambda>(void) && action)
 	yara64.exe!_fwrite_internal(const void * const buffer, const unsigned __int64 size, const unsigned __int64 count, _iobuf * const stream, __crt_cached_ptd_host & ptd)
 	yara64.exe!fwrite(const void * buffer, unsigned __int64 size, unsigned __int64 count, _iobuf * stream)
 	yara64.exe!yara_yylex(YYSTYPE * yylval_param, void * yyscanner, _YR_COMPILER * compiler)
 	yara64.exe!yara_yyparse(void * yyscanner, _YR_COMPILER * compiler)
[...]

PR #2010, opened 2023-11-25
Redefine ECHO macro to not print unmatched text to the console

1ndahous3:
Fix for #2007.


plusvic:
330bc748e4bab110bea6a8ff1e7e81ae675681df is a more appropriate fix to #2007. I like your comment explaining what `YY_FATAL_ERROR` does, I'm adding it in another commit.

PR #2016, opened 2023-11-29
Fix segfault in math module when scanning a process that dies during scan

vthib:
The math module retrieves the first block of memory to do some validation on parameters. However, contrary to for example the hash module, the block value is not checked. This can lead, if for example scanning a process and having the process die during the scan, to a SIGSEGV/Null access violation, as the first block call will return NULL.

This bug manifests in surprising ways. Since yara runs by default evaluation of condition in a try catch block, this bug is actually caught by the try catch block, and a surprising "ERROR_COULD_NOT_MAP_FILE" error is then returned.

In addition, since the try catch bypasses the whole stack, this leads to a memory leak that can add up as more instances of this bug is triggered.

In addition, the magic module is also fixed (although it should not really trigger since the process memory flag is checked beforehand), and the doc is updated.

### DIFF
--- a/docs/writingmodules.rst
+++ b/docs/writingmodules.rst
@@ -657,11 +657,14 @@ checks in your code nevertheless). In those cases you can use the
         const uint8_t* block_data;
 
         block = first_memory_block(context);
-        block_data = block->fetch_data(block)
-
-        if (block_data != NULL)
+        if (block != NULL)
         {
-          ..do something with the memory block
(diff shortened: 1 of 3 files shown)


plusvic (approved):

vthib:
@plusvic Would it be possible to do a new 4.4 release soon with all the new fixes? Thanks a lot

PR #2031, opened 2024-01-27
modules/pe: make resources sanity check stricter

PeterMatula:
Make number-based PE resources sanity checks a bit stricter to avoid more potentially corrupt files.

We came across [this](https://www.virustotal.com/gui/file/ef866e5eeacd096c4dab73c6d2b098253ba46f1ecf45467e6d65a8e1a75b4ca9) sample, whose `NumberOfNamedEntries` and `NumberOfIdEntries` resource entries would pass the resource-numbers-based PE parsing sanity checks, but its resources are corrupt. Resource parsing for the sample in question is still skipped thanks to `yr_le32toh(resource_dir->Characteristics) != 0` part of the condition, but if it were not there, it would pass and run into trouble.

The "solution" is not ideal, it only makes the existing conditions a bit stricter, but it should not really have any downside.

cc @metthal @ladislav-zezula

### DIFF
--- a/libyara/modules/pe/pe.c
+++ b/libyara/modules/pe/pe.c
@@ -450,15 +450,18 @@ static int _pe_iterate_resources(
 
   // A few sanity checks to avoid corrupt files
 
+  WORD NumberOfNamedEntries = yr_le16toh(resource_dir->NumberOfNamedEntries);
+  WORD NumberOfIdEntries = yr_le16toh(resource_dir->NumberOfIdEntries);
+
   if (yr_le32toh(resource_dir->Characteristics) != 0 ||
-      yr_le16toh(resource_dir->NumberOfNamedEntries) > 32768 ||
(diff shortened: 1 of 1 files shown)


plusvic:
My concern with this change is that it doesn't tackle any real issue with files that are corrupt but have a low enough  `NumberOfNamedEntries` and `NumberOfIdEntries` that passes the check. If the problem is that YARA can crash while scanning certain corrupted files, we should aim for avoiding the crash. By lowering the threshold we are only reducing the odds for the crash to occur, but we are not fixing the real issue.

PeterMatula:
I agree with you, this is by no means a comprehensive fix. It only makes the current checks slightly stricter, and maybe, the current state slightly better. Currently, it is not really in our powers to look into it deeper and implement a better solution. It is fine if its not merged.

PR #2044, opened 2024-02-25
countersignature: remove TS_VERIFY_CTX_init() call

orbea:
This is redundant and already handled by `TS_VERIFY_CTX_new()`. Additionally it causes build failures with LibreSSL `>= 3.8.0` which removed this function since it serves no purpose with opaque `TS_VERIFY_CTX` where it will zero out an already zero'd out `ctx` or cause a leak.

See: https://github.com/libressl/portable/issues/1006


orbea:
No Google account, I can't sign the CLA...

PR #2064, opened 2024-04-07
fix several bugs in array type handling in dotnet module

vthib:
Several bugs were present in the parsing and generation of a string to represent an array type, notably when the "lo_bound" value is set.

- The "read_blob_signed" was buggy for values outside the `[-2**6, 2**6-1]`, the sign edition did not use the right bitmask, and the type used was unsigned.

- The display when lo_bound != 0 was buggy:

  - size=5, lobound=0 => should be `5`, this was ok
  - size=5, lobound=1 => should be `1...5`, this was buggy and displayed `1...6`. The "range format" is inclusive.

- 0 should still be displayed if size is 0. Only when size is unset should it be left out. So an array declared as `[5,0,3]` should be displayed the same way, and not as `[5,,3]`.

### DIFF
--- a/libyara/modules/dotnet/dotnet.c
+++ b/libyara/modules/dotnet/dotnet.c
@@ -581,10 +581,10 @@ static int32_t read_blob_signed(const uint8_t** data, uint32_t* len)
   {
     uint16_t tmp1 = yr_be16toh(yr_unaligned_u16(*data));
     // shift and leave top 2 bits clear
-    uint16_t tmp2 = (tmp1 >> 1) & 0x3FFF;
+    int16_t tmp2 = (tmp1 >> 1) & 0x3FFF;
     // sign extension in case of negative number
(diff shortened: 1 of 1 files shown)


vthib:
Yes, I crafted some files for testing in boreal, one of which test array sizes: [types2.dll](https://raw.githubusercontent.com/vthib/boreal/2780b338/boreal/tests/assets/dotnet/types2.dll)

lo bound values on arrays cannot be used in C# files afaik, but they can easily be tested with MSIL files. The aforementioned file was generated from this source code: [types2.cil](https://github.com/vthib/boreal/blob/2780b33/boreal/tests/assets/dotnet/types2.cil)

This makes it easy to compare the syntax used in the MSIL file with the type strings generated in the yara module.

vthib:
I fixed an unnecessary change, should be ok now

plusvic (approved):

PR #2069, opened 2024-04-24
Fix dex module inconsistencies

vthib:
Fix several inconsistencies in the dex module:

- some bytestring values were not properly set: the magic constants, `header.magic` and `header.signature`. Those were cut if a nul byte was present (which is the case in the magic values

- `dex.field[*].static` and `dex.field[*].instance` were defined, but not declared, making their use impossible. They are now properly declared.

- several fields in `dex.method[*].code_item` were declared but never defined: `padding`, `tries` and `handlers`. Those are removed since they couldn't have been used.


vthib:
I'm not sure what the macosx build issue is. It complains about missing libjansson, which is not related to the dex module but to the cuckoo one, which i do not touch. So afaict it is unrelated.

vthib:
I fixed an unnecessary change, should be ok now

plusvic (approved):

PR #2074, opened 2024-05-02
Multiline strings in meta section.

wxsBSD:
This commit implements multiline strings in the meta section. It is now possible to do the following:

rule a {
  meta:
    b = """
This is a multiline string literal.

	\x41

Hooray!
"""
  condition:
    true
}

Escape sequences are handled so the sequence "\x41" in the above example comes out to a single 'A'. Everything else is copied into the rule verbatim.

This is different behavior from regular strings in that things like "\t" are converted to a single byte (0x09). If you want a tab character in your multiline string, you must put it in yourself. Same goes for newlines.


wxsBSD:
Going to close this out and implement it in yara-x instead.

PR #2079, opened 2024-05-08
Modifying yara-python docs to reflect existence of strict_escape bool

tlansec:
Adds documentation to the `yarapython.rst` file which tells users how to use the `strict_escape` .


TommYDeeee (approved):
LGTM, totally forgot about documentation for python. Thanks for this addition

plusvic (approved):

PR #2087, opened 2024-06-13
feat: selectively zero match lists

secDre4mer:
Instead of completely zeroing match lists at the end of each scan (which is O(num_strings) and thus scales badly with large rule sets), keep track of which list entries have been used. Only those entries need to be zeroed at the end. This essentially changes the O(num_strings) to O(num_matched_strings), which is usually far lesser.

### DIFF
--- a/libyara/exec.c
+++ b/libyara/exec.c
@@ -1438,7 +1438,7 @@ int yr_execute_code(YR_SCAN_CONTEXT* context)
 
     case OP_FOUND:
       pop(r1);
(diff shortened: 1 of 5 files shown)


secDre4mer:
Since this hasn't seen any attention so far:  @plusvic do performance improvements like this still have a chance of being merged (since YARA doesn't gain any more features and effort instead goes into yara-x)? If not, I'd like to close this.

plusvic:
Sorry for the delayed response, this fell trough the cracks. As this is a relatively important change, and all my efforts are put now in YARA-X I prefer not merging this PR. So I'm closing it.

PR #2103, opened 2024-09-05
filemap: define PROC_SUPER_MAGIC, avoid linux/magic.h

hillu:
PR #1848 caused build issues with some "unusual" build configurations – apparently we can't rely on linux/magic.h being present when cross-building for musl libc.

Defining PROC_SUPER_MAGIC should not cause a problems since it should be considered part of the Linux kernel/user API and it is unlikely to change.

### DIFF
--- a/libyara/filemap.c
+++ b/libyara/filemap.c
@@ -159,7 +159,12 @@ YR_API int yr_filemap_map_fd(
 
 #ifdef __linux__
 #include <sys/vfs.h>
-#include <linux/magic.h>
+// This constant can be found in linux/magic.h and the statfs(2) manpage.
+// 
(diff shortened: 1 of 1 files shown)


libyara/filemap.c
```diff
@@ -159,7 +159,7 @@ YR_API int yr_filemap_map_fd(
 
 #ifdef __linux__
 #include <sys/vfs.h>
-#include <linux/magic.h>
+#define PROC_SUPER_MAGIC	0x9fa0
```
plusvic, line 162: Can you add a comment here pointing to the github issue? I wasn't to make sure that in the future we known where this constant came from.

plusvic (approved):

PR #2128, opened 2024-12-27
Remove the libyara from path libyara/Makefile.am

n0tduck1e:
The file is no longer present on that path and moved to root of project. Fixing the docs so it does'nt cause confusion to others like it did to me.


plusvic:
Duplicate of #2129

PR #2133, opened 2025-01-09
fix: correct indentation issues

secDre4mer:
There were two minor issues in the YR_DEBUG_FPRINTF statements: 

- yr_process_get_next_memory_block had a debug statement starting with `+`, which YR_DEBUG_FPRINTF interpreted as a function starting, but since there was no corresponding `}`, this caused indentation to permanently increase by 1
- test_pass had a closing debug statement at its end, which decreased indentation (and possibly caused the test to fail if indentation was already 0), but no opening debug statement, causing indentation to effectively decrease by 1

This can be reproduced with `./configure --with-debug-verbose=2`.

### DIFF
--- a/libyara/proc/linux.c
+++ b/libyara/proc/linux.c
@@ -412,7 +412,7 @@ YR_API YR_MEMORY_BLOCK* yr_process_get_next_memory_block(
     }
     else
     {
-      YR_DEBUG_FPRINTF(2, stderr, "+ %s() = NULL\n", __FUNCTION__);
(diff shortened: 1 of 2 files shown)


secDre4mer:
On a different topic: Indentation can also permanently increase by 1 if a SIGBUS or SIGSEGV is caught (since in that case the "closing statement" which decreases indentation is never executed). That's a minor issue, though, and not directly related to the fixes addressed in this PR.

PR #2142, opened 2025-03-17
Integrated Similarity_Matching as a Module

chapl1n03:
This integration enhances the existing YARA engine by integrating similarity matching functionality calculated based on Levenstein distance alongside its traditional exact matching capabilities. By leveraging fuzzy logic, the tool classifies detection results more effectively, allowing for a more comprehensive analysis of potential threats.

By adding similarity matching, the tool significantly improves malware detection rates, especially in scenarios where exact matches fail. Notably, this tool is highly effective when specific YARA rules for a malware are unavailable. It can detect malware by leveraging the similarity to existing YARA rules designed for related or similar malware. This capability makes the tool a powerful solution for identifying unknown or evolving threats, expanding the scope of malware detection and strengthening cybersecurity defenses.

Key Benefits:
1. Improved Detection Rates
2. If specific YARA rules for a malware variant are unavailable, the tool can detect threats based on their similarity to existing YARA rules for related malware.
3. This approach enhances the detection of evolving and obfuscated malware, making it a valuable addition to malware analysis workflows.

Implementation Details
	•	Introduces a partial matching module. (-p)
	•	Adding -p argument to YARA engine, allows users to specify a similarity threshold to perform similarity matching


plusvic:
This is a major change, both in design and implementation, but this project is currently frozen while we work on a new implementation (https://github.com/VirusTotal/yara-x). Every new feature will be added to the new implementation, and this project will receive only bug fixes.

PR #2148, opened 2025-04-24
Calculate a fully matching regex for base64 strings

fwaggle:
Hi folks!

This is a proof-of-concept and needs some serious clean-up, but for now 
I've left it needlessly verbose to hopefully demonstrate better what I'm 
doing and that hopefully I'm not barking up the wrong tree and my 
assumptions (largely gleaned from the Wikipedia page on base64) are 
correct.

It's long been annoying to me that Yara's implementation of base64 is 
fairly naive, particularly not matching terribly well on shorter strings, 
due to it throwing away up to the first couple of characters and the last 
characters if they're a partial match, when it's possible to write a 
regex that'll match it correctly. This is because when there's a partial 
match (the character is covered by two encoded characters), I believe 
there's only a limited subset of characters it can potentially be - 
either 4 or 16 to be precise.

Take for example the string `<?php`. This can have the following 
representations:

```
$ = /PD9waH[ABCD]/
$ = /[DTjz]w\/cGhw/
$ = /[AEIMQUYcgkosw048]8P3Boc[ABCDEFGHIJKLMNOP]/
```

Yara's `base64` keyword however, by throwing away the partial matched 
characters, renders the three possible combinations as:

```
$ = /PD9waH/
$ = /w\/cGhw/
$ = /8P3Boc/
```

This results in the unfortunate behaviour of, for example, `<?phq` 
matching as a false positive.

The changes in this PR add two additional fields to a base64 node: `pre` 
and `post` strings, and adds the combinations of characters that each 
could potentially be, which then `_yr_base64_create_regexp` renders the 
regex at the start/end of the given pattern - we allocate enough space 
for the extra characters and up to two sets of square brackets to 
surround them, leave out the bit that was formerly dropped off and 
replace it with the start and end patterns.

I'm not sure if this is the best way to do it, but it seemed reasonable?

As mentioned at the beginning, this needs cleaning up a lot - I'm very 
rusty at C, probably wasn't great at my prime, my comments are garbage, 
the code is needlessly verbose in places and not readable in others, and 
I'm making various assumptions.

I'm making the following assumptions that I'm aware of:

* That my bitwise math is correct, in all my testing it seems to be 
correct.
* That my math on figuring out the sizes to allocate is correct.
* That yr_free() doesn't free null pointers.
* that _yr_base64_destroy_nodes() cleans up memory I'm allocating 
correctly.
* That yr_free() frees children inside nodes (as I write this I'm now 
fairly sure this isn't correct and I'll need to fix that?)
* That the wide-text mode still works correctly - I'm not 100% sure I 
didn't mess this up.
* That allowing what would have formerly been a zero-length string to go 
through is okay. I'm aware it's fairly disastrous for rule performance, 
but searching for a single character string encoded to base64 now works - 
though one of the three representations renders as a regex with two sets 
of four possible characters with no static characters whatsoever.

Please let me know if any of these look incorrect.

If it's otherwise not objectionable I should be able to put the work in 
to cleaning this up so it's mergeable, writing some tests, and updating 
the documentation, but I am happy if someone else hammers it into shape 
as well.

## Test cases for your perusal

Strings that should match the pattern `$ = "<?php" base64`, currently do, 
and do after the patch:

```
PD9waHAK
IDw/cGhwCg==
QUE8P3BocAo=
```

Some strings that currently match that shouldn't, and don't after the 
patch:

```
PD9waHEK
ICA8P3BocQo=
```

### DIFF
--- a/libyara/base64.c
+++ b/libyara/base64.c
@@ -207,15 +207,15 @@ static int _yr_base64_create_nodes(
 {
   SIZED_STRING* encoded_str;
   SIZED_STRING* final_str;
+  SIZED_STRING* pre_str;
+  SIZED_STRING* post_str;
   BASE64_NODE* node;
+  char *alphabet_str = alphabet->c_string;
 
   int pad;
 
   for (int i = 0; i <= 2; i++)
   {
-    if (i == 1 && str->length == 1)
-      continue;
-
     node = (BASE64_NODE*) yr_malloc(sizeof(BASE64_NODE));
     if (node == NULL)
       return ERROR_INSUFFICIENT_MEMORY;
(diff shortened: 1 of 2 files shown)


fwaggle:
I've changed this to draft while I agree the test situation, and I can probably make the main text a lot more terse because of it.

fwaggle:
I have fixed most of the broken tests, however I've very clearly not implemented wide strings in the pre/post correctly, and I've also managed to mess up the custom alphabet too.

fwaggle:
@plusvic Yeah, I snoozed on this too long I guess - I meant to do something about it several years back, and the itch got too great to ignore. I wasn't actually aware YARA was in maintenance mode at all, guess I should look at updating our application to use YARA-X instead 😅 

Oh well, I had an entertaining time doing it.

PR #2153, opened 2025-05-22
Add module for IEEE 754

adepasquale:
The IEEE 754 module allows you to parse float32 and double64 formats from any offset and to create signatures based on the interpreted floating point value.

```
import "ieee754"

rule IEEE754_Example {
  condition:
    ieee754.float32le(0) > 3.14159 and ieee754.double64be(8) < 2.718281828
}
```


plusvic:
Given that similar built-in function exists for integers, I would prefer that these functions are also built-in and not part of some additional module. In the other hand, I'm not adding new features to YARA at this point, only bug fixes. Any new feature will be added to YARA-X instead (https://github.com/VirusTotal/yara-x).

adepasquale:
Understood. Feel free to close this PR.

PR #2191, opened 2026-02-08
Make yr_string_set_destroy non-public

hillu:
The function is only called from within the compiler; there's no need to expose it to users of the library.


hillu:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project.

Oops, initially committed with an email address that the bot hadn't known about, fixed.

plusvic (approved):

PR #2200, opened 2026-05-30
guard against undersized rich header in pe_parse_rich_signature

jmestwa-coder:
When the backward scan in pe_parse_rich_signature finds a DanS marker only one to three dwords ahead of the Rich marker, rich_len ends up smaller than RICH_SIGNATURE. The unsigned subtraction rich_len - sizeof(RICH_SIGNATURE) then wraps and rich_count becomes a huge bogus value that drives the version_data allocation and copy loop. _rich_version already rejects rich_length < sizeof(RICH_SIGNATURE) for the same computation, so apply the same check before computing rich_count here.


plusvic:
@jmestwa-coder do you have some file that we can use as a test case?

plusvic (approved):

jmestwa-coder:
Yeah, just pushed one. `tests/data/undersized_rich` is a minimal PE where the DanS marker lands one dword ahead of Rich, so rich_length comes out as 4 and the old code underflowed on `rich_len - sizeof(RICH_SIGNATURE)`. Added an assert in test-pe.c that checks it parses as a PE with rich_signature.length == 4.

PR #2211, opened 2026-06-24
fix: Fix INT64_MIN UB in constant folding (*, \, unary -)

Nambers:
Constant folding in `grammar.y` hits UB at the `INT64_MIN` boundary, reachable from rule text:

- `*`: `llabs(INT64_MIN)` is UB and returns a negative value, so the overflow guard is bypassed and folds to a garbage constant instead of `ERROR_INTEGER_OVERFLOW`. Now compares magnitudes in unsigned space.
- `\`: `INT64_MIN / -1` is overflow UB and raises SIGFPE on x86. Now rejected with `ERROR_INTEGER_OVERFLOW`.
- unary `-`: `-INT64_MIN` is negation UB. Now negated in unsigned space (wraps to `INT64_MIN`, unchanged behavior).

Reproduce  `condition` expr (each reported an UBSan error):
    `~9223372036854775807 * 9223372036854775807`
    `~9223372036854775807 \ -1`
    `-(~9223372036854775807)`

Example Logs:
```
libyara/grammar.y:2813:30: runtime error: negation of -9223372036854775808 cannot be represented in type 'int64_t' (aka 'long'); cast to an unsigned type to negate this value to itself
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior libyara/grammar.y:2813:30 
```

Note: The VM ops in `exec.c` and the `%` rule likely share this UB.

Disclose: assisted with AI

### DIFF
--- a/libyara/grammar.c
+++ b/libyara/grammar.c
@@ -959,7 +959,7 @@ static const yytype_int16 yyrline[] =
     2419,  2429,  2444,  2443,  2456,  2457,  2462,  2495,  2520,  2576,
     2583,  2589,  2595,  2605,  2609,  2617,  2629,  2643,  2650,  2657,
     2682,  2694,  2706,  2718,  2733,  2745,  2760,  2803,  2824,  2859,
-    2894,  2928,  2953,  2970,  2980,  2990,  3000,  3010,  3030,  3050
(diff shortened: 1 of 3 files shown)


plusvic:
Thanks for your contribution @Nambers, can you add some test case for this fix?

Nambers:
@plusvic done

PR #2224, opened 2026-07-11
bound tilde stream row-count read in dotnet_parse_tilde_2

### LINKED ISSUE #2196: Out-of-Bounds Read in dotnet_parse_tilde
https://github.com/VirusTotal/yara/blob/323356eede76abb87d7b14526cb2970883b87a27/libyara/modules/dotnet/dotnet.c#L3138

The issue sits in the if condition of the ROW_CHECK macro definition:

  #define ROW_CHECK(name) \
    if (fits_in_pe(pe, row_offset, (matched_bits + 1) * sizeof(uint32_t))) \
      [rows.name](http://rows.name/) = *(row_offset + matched_bits);

The loop iterates up to 64 times driven by the attacker-controlled `Valid` bitmask in the TILDE header, and each iteration reads a row count entry at `row_offset + matched_bits`. The guard is checking bounds against the PE file's outer boundary, not the tilde stream's declared size. The stream's own extent — derivable from `streams->tilde->Offset + streams->tilde->Size` — is never consulted. This means the condition can pass while the read on the next line is already outside the stream.

A crafted PE with `Valid = 0xFFFFFFFFFFFFFFFF` and an undersized tilde stream will pass the check as long as adjacent PE data occupies memory beyond the stream, and subsequent iterations will read out of stream bounds silently.

Downstream impact is meaningful: `dotnet_parse_tilde_2` inherits the corrupted row counts from the ROWS struct and uses them to calculate table offsets and sizes throughout its own processing loop. Garbage row counts there can cause further out-of-bounds reads, misaligned table parsing, and potential information disclosure depending on what lies beyond the stream in the file.

Recommended fix has two parts.

First, replace the per-iteration `fits_in_pe` check with a stream-boundary-aware guard:

  const uint8_t* tilde_stream_end = pe->data + metadata_root
      + yr_le32toh(streams->tilde->Offset)
      + yr_le32toh(streams->tilde->Size);

  #define ROW_CHECK(name) \
    if ((uint8_t*)(row_offset + matched_bits + 1) <= tilde_stream_end) \
      [rows.name](http://rows.name/) = *(row_offset + matched_bits);

Second, add a pre-loop eager validation using popcount on the Valid field to confirm the entire row count array fits within the stream before any iteration begins — this eliminates the per-iteration ambiguity entirely and is the cleaner long-term fix:

  uint32_t valid_count = __builtin_popcountll(yr_le64toh(tilde_header->Valid));
  if (!fits_in_pe(pe, row_offset, valid_count * sizeof(uint32_t)))
      return;

munzzyy:
Since #2196, `dotnet_parse_tilde` computes `tilde_stream_end` and checks each row-count read against it instead of `fits_in_pe`, because `fits_in_pe` was only bounding the read against the whole PE buffer rather than the `#~` stream's own declared `Size`.

`dotnet_parse_tilde` calls `dotnet_parse_tilde_2` right after, passing the same `tilde_header` and `streams`. But `dotnet_parse_tilde_2` doesn't use the already-validated row counts to drive its own table walk - it independently re-reads the identical row-count array with the old `fits_in_pe` check, which #2196 never touched:

```c
if (!fits_in_pe(pe, row_offset + matched_bits, sizeof(uint32_t)))
  return;

num_rows = yr_le32toh(*(row_offset + matched_bits));
```

So a `#~` stream with an undersized `Size` still lets `dotnet_parse_tilde_2` read row counts (and therefore drive its table_offset walk) from whatever bytes physically follow the stream in the file - padding, another stream, the cert table. That read is still bounded by `fits_in_pe` against the real file allocation, so it's not an overrun of the buffer, but it can surface unrelated file content through `dotnet.*` fields as if it were legitimate metadata table data.

This gives `dotnet_parse_tilde_2` its own `tilde_stream_end` and swaps in the same `<= tilde_stream_end` guard `dotnet_parse_tilde` already uses, so both functions agree on where the stream actually ends.

Built locally and ran the dotnet test binary (`make test-dotnet`), passes clean. Also confirmed by diffing against the commit that closed #2196 - it only touched `dotnet_parse_tilde`, this just brings the sibling in line with it.

### DIFF
--- a/libyara/modules/dotnet/dotnet.c
+++ b/libyara/modules/dotnet/dotnet.c
@@ -1917,6 +1917,10 @@ void dotnet_parse_tilde_2(
   const uint8_t* string_offset;
   const uint8_t* blob_offset;
 
+  const uint8_t* tilde_stream_end = pe->data + metadata_root +
+                                    yr_le32toh(streams->tilde->Offset) +
+                                    yr_le32toh(streams->tilde->Size);
+
   uint32_t num_rows = 0;
   uint32_t valid_rows = 0;
   uint32_t* row_offset = NULL;
@@ -1992,7 +1996,8 @@ void dotnet_parse_tilde_2(
     if (!((yr_le64toh(tilde_header->Valid) >> bit_check) & 0x01))
       continue;
 
-    if (!fits_in_pe(pe, row_offset + matched_bits, sizeof(uint32_t)))
+    if (!fits_in_pe(pe, row_offset + matched_bits, sizeof(uint32_t)) ||
+        (uint8_t*) (row_offset + matched_bits + 1) > tilde_stream_end)
       return;
 
     num_rows = yr_le32toh(*(row_offset + matched_bits));


plusvic:
I'm wondering if checking against `tilde_stream_end` is actually correct, as `tilde_stream_end` is computed from data that is read directly from the file its value could be manipulated.

munzzyy:
You're right, good catch.

`tilde_stream_end` is built from `streams->tilde->Offset` and `Size`, both read raw off the stream header in `dotnet_parse_stream_headers` (dotnet.c:1833-1840). Nothing clamps those against `pe->data_size` there - `struct_fits_in_pe` at line 1814 only checks the header struct itself fits, not the Offset/Size values it stores. So `tilde_stream_end` can end up pointing past the real buffer.

`dotnet_parse_tilde` gets away with this (line 3119) because it still runs `fits_in_pe(pe, row_offset, valid_count * sizeof(uint32_t))` over the whole row array first (line 3147), and only uses `tilde_stream_end` afterward as an extra restriction on top of that already-verified range (line 3159). The bound that actually protects the read is the `fits_in_pe` call, not `tilde_stream_end`.

My change to `dotnet_parse_tilde_2` dropped the per-row `fits_in_pe` and made `tilde_stream_end` the only check, so it lost that protection. Pushed a fix that keeps `fits_in_pe` and adds the `tilde_stream_end` check on top instead of replacing it (dotnet.c:1999), same two-layer pattern `dotnet_parse_tilde` uses. Ran `make test-dotnet` locally again, still passes.

plusvic (approved):

PR #2247, opened 2026-09-18
Add yr_get_version to report the version of the library loaded at runtime

### LINKED ISSUE #2136: Missing yr_get_version API function
**Is your feature request related to a problem? Please describe.**

Many Linux distros provide and distribute `libyara.so` as a shared library in a separate package from the applications which use Yara. This is advantageous to applications which use libyara since they can benefit from independent updates. 

However there is no API in `libyara.so` which would allow such an application to determine and show the library version of libyara loaded via dlopen() in runtime.

**Describe the solution you'd like**

Add an API function:

```
void yr_get_version( int * major, int * minor, int * micro)
{
  *major = YR_MAJOR_VERSION;
  *minor = YR_MINOR_VERSION;
  *micro = YR_MICRO_VERSION;
}
```
or something similar.

**Describe alternatives you've considered**
An alternative would be to rebuild all packages using Yara statically, and stop providing libyara.so but this isn't what Linux distros are doing.

MRX-72:
YR_MAJOR/MINOR/MICRO_VERSION are compile-time constants, so a program linked against a shared libyara can't tell which version it actually loaded at runtime. That matters on distros that ship libyara.so as its own package so it can be updated independently of the programs using it.

This adds:

    YR_API void yr_get_version(int* major, int* minor, int* micro);

Any of the pointers can be NULL if you only want some of the fields, and it touches no global state so it works before yr_initialize (which is the point, for the dlopen case).

Tested with a shared build on macOS: dlopen the lib, dlsym yr_get_version, and it reports the version without initializing. I also extended tests/test-version.c to cover the three fields and the NULL combinations, and the full suite passes.

I set the docs to versionadded 4.5.9 as a guess since the tree is at 4.5.8 - change it to whatever the next release ends up being.

Closes #2136


MRX-72:
@plusvic friendly ping — CLA is signed and this small addition is ready for review whenever you have a moment.

docs/capi.rst
```diff
@@ -510,6 +510,19 @@ Functions
...
+  Get the version of the library loaded at runtime. The
+  :c:macro:`YR_MAJOR_VERSION`, :c:macro:`YR_MINOR_VERSION` and
+  :c:macro:`YR_MICRO_VERSION` macros are fixed when your program is compiled,
+  so a program linked against a shared ``libyara`` cannot use them to tell
+  which version it actually got. This function reports that.
+
+  Any argument may be ``NULL`` if you do not need that component. Unlike the
+  rest of the API, this may be called before :c:func:`yr_initialize`.
+
+  .. versionadded:: 4.5.9
```
plusvic:
```suggestion
  .. versionadded:: 4.6.0
```

plusvic: A new API is worth a minor version bump.

plusvic (approved):

